Issue #66, opened 2026-04-06
JD textarea has no live character count — min-length check only fires on submit


Sachinchaurasiya360:
**File:** `client/src/module/student/cover-letter/CoverLetterPage.tsx`

**Problem:**
The form validates that the job description is at least 50 characters, but only **at submit time**. Until the student clicks "Generate," there's no indication that:

- A minimum exists.
- How close they are to meeting it.
- Why their submit attempt just bounced.

Result: student pastes a short JD snippet, clicks generate, gets an error, has to read it, figure out what "50 characters minimum" means in practice, and guess how much more to add. This is a solved UX problem — live counters exist on every Twitter-like text input in existence.

**Why it matters:**
- Submit-time validation is a hostile UX pattern when the constraint is knowable up front.
- Students often paste partial JDs (just the responsibilities, not the full listing) and have no way of knowing if that's enough.
- Fixing this is a 5-line change.

**Fix:**

1. **Add a live character counter** below the textarea:

   ```tsx
   <div className="flex justify-between text-xs text-gray-500 mt-1">
     <span className={jdLength < 50 ? "text-amber-600" : "text-green-600"}>
       {jdLength < 50
         ? `${50 - jdLength} more characters needed`
         : "Ready to generate"}
     </span>
     <span>{jdLength} / 50 minimum</span>
   </div>
   ```

2. **Disable the submit button until the minimum is met.** Don't let the student click through to an error they could have seen coming.

3. **Soft-gate with guidance, not just a hard block.** If the JD is short but technically valid (e.g. 60 chars), show a gentle hint: _"Longer job descriptions produce better-tailored cover letters. Aim for 200+ characters for best results."_ — educates without blocking.

4. **Consider raising the minimum.** 50 characters is barely a sentence. For a tool that generates tailored cover letters, 200+ is a more realistic floor. But don't change this without checking existing usage data — some students may legitimately paste short role blurbs.

5. **Upper limit too:** If there's a max (for prompt size reasons), show it in the counter. If there isn't one, add a soft warning at ~5000 chars that very long JDs may get truncated.

thisisAtharv:
Hi @Sachinchaurasiya360 , I would like to do this task! Could you please assign it to me?

PR #82, opened 2026-05-03
Feat: added soft hints and length warnings to JD textarea (#66)

thisisAtharv:
Added the missing live character counter and UX hints for the cover letter JD textarea.

   **Changes:**

- Replaced the static counter with a dynamic one (shows exactly how many characters are left, then turns green when ready).
- Added a soft hint for JDs under 200 chars recommending they add more text.
- Added a warning for massive JDs (5000+ chars) so users know it might get truncated.

(Note: The submit button was actually already being disabled correctly under 50 chars in the existing code, so I just left that logic alone and added the UI feedback above it).

## Summary by CodeRabbit

* **New Features**
  * Job description input now displays an enhanced feedback panel with status messages ("minimum characters," "more characters needed," "Ready to generate") and a character counter
  * Added helpful guidance text when input meets minimum requirements but falls below 200 characters
  * Added warning message when input exceeds 5000 characters

### DIFF
--- a/client/src/module/student/ats/CoverLetterPage.tsx
+++ b/client/src/module/student/ats/CoverLetterPage.tsx
@@ -283,15 +283,30 @@ export default function CoverLetterPage() {
                   value={jobDescription}
                   onChange={(e) => setJobDescription(e.target.value)}
                 />
-                <p
-                  className={`mt-1.5 text-[10px] font-mono uppercase tracking-widest tabular-nums ${
-                    jobDescription.length > 0 && jobDescription.length < JD_MIN_CHARS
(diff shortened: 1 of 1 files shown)


Sachinchaurasiya360:
Can you add before after pictures

thisisAtharv:
> Can you add before after pictures

@Sachinchaurasiya360 
---
- **Before:** Missing how many characters are left, soft hints, ready to generate and length warnings 

image



---
- **After:** 

1.  Shows exactly how many characters are left, turns green when ready.

image


2.  Added soft hints for JDs under 200 chars recommending they add more text.

image


3.  Added warning for massive JDs 5000+ chars so users know it might get truncated.

image

PR #203, opened 2026-05-14
fix: use stable UUID as React key for chat messages instead of array index

### LINKED ISSUE #34: Bug: Chat messages use array-index as React key (breaks reconciliation on streaming)
## Problem
[JobAgentPage.tsx:243](client/src/module/student/job-agent/JobAgentPage.tsx#L243):

```tsx
{messages.map((msg, i) => (
  <AgentMessage key={i} role={msg.role} content={msg.content} jobs={msg.jobs} />
))}
```

`key={i}` is the array index. React uses keys to match old elements to new ones across renders; when the key is just the position, any mutation that shifts elements (insertion at head, splice, reorder) makes React reuse the *wrong* component instance and patch its props in place instead of unmounting/mounting.

**Why it matters for this specific page:**
- Streaming AI replies mutate the *last* message's `content` on every token. With `key={i}`, as long as no insertion shifts things, it happens to work — but as soon as an optimistic user message or a retry shuffles the list, the assistant's DOM state (scroll position inside a job list, any `useState` inside `AgentMessage`, CSS transitions, focus, selection) gets inherited by a different message.
- `AgentMessage` renders `jobs={msg.jobs}` — that child list has its own key requirements. A reused parent with stale internal state can hold onto a prior render's job card DOM nodes while props silently swap underneath.
- Framer Motion `AnimatePresence` / enter animations on messages won't fire correctly for prepended items.
- React DevTools warnings aside, this is a latent bug that will surface the moment message editing, retry, or delete is added.

## Fix
Give every message a stable id at the moment it's created, and key on that.

**Option A — client-generated (simplest, ship today):**
```tsx
type ChatMessage = {
  id: string;           // new
  role: "user" | "assistant";
  content: string;
  jobs?: Job[];
};

// when appending:
setMessages(prev => [...prev, { id: crypto.randomUUID(), role: "user", content: input }]);
setMessages(prev => [...prev, { id: crypto.randomUUID(), role: "assistant", content: "" }]);

// render:
{messages.map(msg => (
  <AgentMessage key={msg.id} role={msg.role} content={msg.content} jobs={msg.jobs} />
))}
```

`crypto.randomUUID()` is available in all browsers Vite targets. No dependency needed.

**Option B — server-generated (preferred long-term):**
If `POST /api/job-agent/chat` (or whatever the chat controller is — grep for the `chatMut` endpoint) persists messages, return the DB row `id` in the response and use that. Keep a temporary UUID for the optimistic user message, then replace it with the server id when the response arrives.

## Acceptance criteria
- [ ] `ChatMessage` type has a required `id: string` field.
- [ ] Every `setMessages` call assigns an id (UUID for optimistic, server id when available).
- [ ] `messages.map` uses `key={msg.id}`, not `key={i}`.
- [ ] Streaming still updates the last assistant message in place (no flicker, no remount).
- [ ] Grep the rest of `job-agent/` for other `key={i}` / `key={index}` and fix any found.
- [ ] No new ESLint `react/no-array-index-key` warnings.

## Out of scope
- Persisting chat history server-side (if not already done).
- Message edit / retry / delete features — this fix unblocks them but doesn't implement them.

aaniya22:
Fixes #34

## Changes
- Added `id: string` field to `ChatMessage` interface in `types.ts`
- Added `id: string` field to `DisplayMessage` interface in `JobAgentPage.tsx`
- Assigned `crypto.randomUUID()` on every `setMessages` call that creates a new message
- Changed `key={i}` to `key={msg.id}` in the render map
- No new ESLint react/no-array-index-key warnings

## Summary by CodeRabbit

* **Refactor**
  * Improved chat message handling by introducing stable identifiers for each message in conversations, enhancing the reliability and consistency of message tracking and display throughout the chat interface.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Sachinchaurasiya360/InternHack/pull/203)

### DIFF
--- a/client/src/lib/types.ts
+++ b/client/src/lib/types.ts
@@ -1113,6 +1113,7 @@ export interface EmailCampaignDetail extends EmailCampaign {
 }
 
 export interface ChatMessage {
+  id: string;
   role: "user" | "assistant";
   content: string;
 }
--- a/client/src/lib/types.ts
+++ b/client/src/lib/types.ts
@@ -1154,6 +1154,7 @@ export interface JobPreferences {
 }
 
 export interface JobAgentMessage {
+  id?: string;
   role: "user" | "assistant";
   content: string;
   timestamp: string;
--- a/client/src/module/student/job-agent/JobAgentPage.tsx
+++ b/client/src/module/student/job-agent/JobAgentPage.tsx
@@ -1,4 +1,4 @@
-import { useState, useRef, useEffect, useCallback, useMemo } from "react";
+import { useState, useRef, useEffect, useCallback } from "react";
 import { motion, AnimatePresence } from "framer-motion";
 import { useQuery, useMutation, useQueryClient } from "@tanstack/react-query";
 import {
@@ -22,6 +22,7 @@ import { AgentMessage } from "./AgentMessage";
 import { ThinkingIndicator } from "./ThinkingIndicator";
 
 interface DisplayMessage {
+  id: string;
   role: "user" | "assistant";
   content: string;
   jobs?: JobFeedMatch["job"][];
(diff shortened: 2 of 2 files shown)


Sachinchaurasiya360:
There is some conflict in JobAgentPage.tsx @aaniya22

PR #416, opened 2026-05-22
fix: session invalidation, atomic progress upsert, ATS IDOR bypass (#…

### LINKED ISSUE #379: # [BUG] Multiple confirmed backend issues found during codebase review: session invalidation gap, non-atomic progress writes, and ATS ownership bypass
I was reviewing the backend codebase and found several confirmed issues that currently exist in production paths. I also checked existing issues first and did not find duplicates for these findings.

---

# 1. Security: `resetPassword()` does not invalidate existing sessions

**File:** `server/src/module/auth/auth.service.ts`
**Function:** `resetPassword()`

After password reset, the user's password hash is updated, but existing JWT sessions remain valid because `tokenVersion` is never incremented and `invalidateVersionCache()` is never called.

Other auth flows (`login`, `googleAuth`, `verifyEmail`) already invalidate old sessions correctly using:

```ts
const versionUpdate = await prisma.user.update({
  where: { id: user.id },
  data: { tokenVersion: { increment: 1 } },
  select: { tokenVersion: true },
});

invalidateVersionCache(user.id);
```

`resetPassword()` skips this logic entirely.

## Impact

If an attacker already possesses a valid JWT, resetting the password does not revoke that token. The attacker retains access until token expiry (~7 days).

## Suggested fix

At the end of `resetPassword()`, add:

```ts
const versionUpdate = await prisma.user.update({
  where: { id: user.id },
  data: { tokenVersion: { increment: 1 } },
  select: { tokenVersion: true },
});

invalidateVersionCache(user.id);
```

## Labels

`bug` `security`

---

# 2. Concurrency bug: non-atomic create/update flow in `updateProgress()`

**File:** `server/src/module/interview-progress/interview-progress.service.ts`
**Function:** `updateProgress()`

Current implementation performs:

```ts
const existingProgress =
  await prisma.userInterviewProgress.findUnique(...)

const progress =
  existingProgress ??
  (await prisma.userInterviewProgress.create(...))

return prisma.userInterviewProgress.update(...)
```

This is not atomic.

Two concurrent requests for the same `userId` can both pass the `findUnique()` check and both attempt `create()`. One request then fails with a unique constraint violation, which currently bubbles into a server error.

## Reproduction scenario

* User rapidly clicks interview questions
* Multiple requests arrive simultaneously
* Both requests observe no row
* Both attempt insert
* One crashes with unique constraint error

## Suggested fix

Replace the multi-step flow with a single atomic `upsert()`:

```ts
return prisma.userInterviewProgress.upsert({
  where: { userId },
  create: {
    userId,
    completedIds,
    bookmarkedIds,
    lastVisitedId,
    lastVisitedAt,
  },
  update: {
    completedIds,
    bookmarkedIds,
    lastVisitedId,
    lastVisitedAt,
  },
});
```

## Labels

`bug`

---

# 3. Security: ATS resume ownership validation can be bypassed

**File:** `server/src/module/ats/ats.service.ts`
**Function:** `scoreResume()`

Current ownership validation:

```ts
const isOwned =
  user.resumes.includes(input.resumeUrl) ||
  getS3KeyFromUrl(input.resumeUrl) !== null;
```

The second condition is overly permissive.

`getS3KeyFromUrl()` returns non-null for any URL matching the application's S3 bucket pattern. That means any valid S3 resume URL inside the bucket passes the ownership check, even if the resume belongs to another user.

## Impact

An attacker who obtains another user's resume URL can submit it to the ATS scoring endpoint and gain unauthorized access to private resume data.

This is effectively an IDOR-style authorization bypass.

## Suggested fix

Remove the permissive fallback condition entirely:

```ts
const isOwned =
  user.resumes.includes(input.resumeUrl);
```

If broader validation is required, ownership should be checked against persisted upload metadata tied to the authenticated user.

## Labels

`bug` `security`

---

# Summary

These findings were identified from direct source review and are reproducible from current implementation behavior:

1. Password reset does not revoke active sessions
2. Interview progress writes are race-condition prone
3. ATS ownership validation allows cross-user resume access

All three appear actionable and isolated enough for targeted fixes.

Aryan1092raj:
- auth.service: increment tokenVersion + invalidate cache after resetPassword so stolen tokens are revoked immediately on password change
- interview-progress.service: replace find+create+update with prisma.upsert to eliminate race condition on concurrent requests (was throwing P2002)
- ats.service: remove permissive S3 URL fallback in IDOR ownership check; only resumes in user.resumes[] are accepted

Fixes #379

## Summary by CodeRabbit

* **Security**
  * Strengthened resume ownership validation and now consistently normalizes resume URLs to reduce accidental exposure.

* **Bug Fixes**
  * Password reset now revokes all active sessions, logging users out of other devices.

* **Improvements**
  * Interview progress updates are now persisted atomically with improved consistency and retry handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Sachinchaurasiya360/InternHack/pull/416?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/server/src/module/ats/ats.service.ts
+++ b/server/src/module/ats/ats.service.ts
@@ -31,9 +31,7 @@ export class AtsService {
     });
     if (!user) throw new Error("User not found");
 
-    const isOwned =
-      user.resumes.includes(input.resumeUrl) ||
-      getS3KeyFromUrl(input.resumeUrl) !== null; // allow any S3 URL they uploaded via the upload endpoint
+    const isOwned = user.resumes.includes(input.resumeUrl);
     if (!isOwned) {
       throw new Error("Resume does not belong to this user");
     }
@@ -99,9 +97,7 @@ export class AtsService {
     });
     if (!user) throw new Error("User not found");
 
-    const isOwned =
-      user.resumes.includes(input.resumeUrl) ||
-      getS3KeyFromUrl(input.resumeUrl) !== null;
+    const isOwned = user.resumes.includes(input.resumeUrl);
     if (!isOwned) {
       throw new Error("Resume does not belong to this user");
     }
--- a/server/src/module/ats/ats.service.ts
+++ b/server/src/module/ats/ats.service.ts
@@ -23,7 +23,13 @@ const __dirname = path.dirname(__filename);
 const SCORE_CACHE_TTL_MS = 24 * 60 * 60 * 1000;
 
 export class AtsService {
+  private normalizeResumeUrl(resumeUrl: string): string {
+    return resumeUrl.split("?")[0] ?? resumeUrl;
+  }
+
   async scoreResume(studentId: number, input: ScoreResumeInput) {
+    const resumeUrl = this.normalizeResumeUrl(input.resumeUrl);
+
     // IDOR check: verify the resume belongs to this student
     const user = await prisma.user.findUnique({
       where: { id: studentId },
@@ -31,7 +37,7 @@ export class AtsService {
     });
     if (!user) throw new Error("User not found");
 
-    const isOwned = user.resumes.includes(input.resumeUrl);
+    const isOwned = user.resumes.includes(resumeUrl);
     if (!isOwned) {
       throw new Error("Resume does not belong to this user");
     }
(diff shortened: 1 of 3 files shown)


Sachinchaurasiya360:
Merged to main with one fix: added missing EOF newline in interview-progress.service.ts. All three security improvements are solid: normalizeResumeUrl strips query params to prevent IDOR bypass, tokenVersion increment ensures old JWTs are invalidated on password reset, and the serializable transaction upsert eliminates the race condition in progress updates.

PR #645, opened 2026-05-28
Fix/assessment timer autosubmit

### LINKED ISSUE #457: Bug: Assessment test does not auto-submit when timer expires
When the timer in `AssessmentTestView` reaches zero, the test simply stops the countdown but does not submit the student's current answers. The student's work is lost.

**File:** `client/src/module/student/applications/AssessmentTestView.tsx`

**Expected behavior:** When the timer hits 0, automatically submit whatever answers have been selected, exactly as if the student clicked Submit.

**Fix:** In the timer `useEffect`, when `timeLeft === 0` call the same submit handler used by the submit button. Guard against double submission with an `isSubmitting` ref.

Harkiratcodess:
## Bug
The timer `useEffect` had `secondsLeft` in its dependency array,
causing a new `setInterval` to be created every second. This caused
unnecessary overhead and could lead to timing drift or unreliable
auto-submit behavior at `secondsLeft === 0`.

## Fix
- Removed `secondsLeft` from the timer `useEffect` dependency array
- The interval now starts once on mount and uses the functional form
  of `setSecondsLeft` to safely read previous state without needing
  `secondsLeft` as a dependency
- Auto-submit via `autoSubmittedRef` guard remains intact

Closes #457


Sachinchaurasiya360 (approved):
Correct fix. Removing secondsLeft from the useEffect dependency array eliminates the setInterval recreation on every tick. The functional updater form safely reads previous state. Clean and focused. Approved.

PR #794, opened 2026-05-30
fix: apply DSA CSV import body limit before global parser

### LINKED ISSUE #776: Bug: DSA CSV import size limit is applied after global JSON parser
Line:
server/src/index.ts:160
server/src/module/dsa/dsa.routes.ts:21

Problem:
I noticed this while tracing the DSA import flow for GSSoC 2026. The CSV import route tries to allow a larger JSON payload with `express.json({ limit: "6mb" })`, but the app-level JSON parser already runs earlier in `server/src/index.ts`:

```ts
app.use(express.json());
```

Then the DSA route adds the larger parser later:

```ts
dsaRouter.post("/import/csv", authMiddleware, requireRole("STUDENT"), express.json({ limit: "6mb" }), ...)
```

Because Express body parsing happens in middleware order, requests to `/api/dsa/import/csv` will usually be parsed, or rejected, by the global parser before the route-level 6 MB parser gets a chance to run. That means the larger limit on the route does not actually protect the import flow.

Repro steps:
1. Run the API locally.
2. Send a JSON body larger than the default `express.json()` limit to `POST /api/dsa/import/csv`.
3. The request can be rejected before reaching `dsaImportController.previewCsv`.
4. The route-level `6mb` parser does not get a chance to handle the payload.

Expected: the CSV import endpoint should accept valid import payloads up to the intended 6 MB limit.
Current: the global JSON parser can reject the request before the DSA-specific limit applies.

Impact:
Students importing larger LeetCode/CSV progress files may hit a payload-size failure even though the route appears to be configured for larger imports. It also makes the code misleading because future contributors may assume the route-level limit is effective when it is not.

Suggested fix:
Mount the larger JSON parser before the global parser for this specific route, or register the DSA import route with its body parser before `app.use(express.json())`. Another clean option is to configure a route-specific parser at the app level, similar to how webhook raw bodies are mounted before the global JSON middleware.

Labels:
type:bug, level:intermediate, gssoc:approved

I would like to work on this under GSSoC 2026 if maintainers are okay with the approach. Please assign it to me.

eshaanag:
# Pull Request

## Description
Fixes the DSA CSV import body parser ordering so the `6mb` JSON limit is applied before the global `express.json()` middleware.

The CSV import endpoint previously declared `express.json({ limit: "6mb" })` inside `dsa.routes.ts`, but by then the global parser in `index.ts` had already parsed or rejected the request. This moves the DSA CSV parser to `server/src/index.ts` before the global parser, matching the existing webhook parser pattern.

## Related Issue
Closes #776

## Type of Change
- [x] Bug Fix
- [ ] Feature
- [ ] Enhancement
- [ ] Documentation

## Testing
- `cd server && npm run typecheck`
- `cd server && npm test`
- `cd server && npm run lint`

## Screenshots
N/A - backend middleware fix.

## Checklist
- [x] Screenshots added for UI changes (if applicable)

Raised as part of GSSoC 2026. Maintainers, please add the relevant GSSoC label if this PR is eligible.

## Summary by CodeRabbit

* **Improvements**
  * Enhanced CSV import functionality to support larger file uploads (up to 6MB).

[Review Change Stack](https://app.coderabbit.ai/change-stack/Sachinchaurasiya360/InternHack/pull/794?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/server/src/index.ts
+++ b/server/src/index.ts
@@ -156,6 +156,8 @@ app.get("/api/health", (_req, res) => {
 app.use("/api/email-inbound/webhook", express.raw({ type: "application/json" }));
 // Raw body for Dodo Payments webhook (must be BEFORE express.json())
 app.use(PAYMENT_WEBHOOK_PATH, express.raw({ type: "application/json" }));
+// Larger body parser for DSA CSV import (must be BEFORE the global parser)
+app.use("/api/dsa/import/csv", express.json({ limit: "6mb" }));
 
 app.use(express.json());
 app.use(cookieParser());
@@ -343,5 +345,4 @@ app.listen(PORT, async () => {
 app.get("/", (req, res) => {
   res.send("Server Running Successfully");
 });
-
 
--- a/server/src/module/dsa/dsa.routes.ts
+++ b/server/src/module/dsa/dsa.routes.ts
@@ -1,5 +1,4 @@
 import { Router } from "express";
-import express from "express";
 import { DsaController } from "./dsa.controller.js";
 import { DsaService } from "./dsa.service.js";
 import { DsaImportController } from "./dsa-import.controller.js";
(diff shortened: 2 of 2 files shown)


Sachinchaurasiya360 (approved):
## Review: Approved

Correctly diagnosed and fixed middleware ordering issue. The global `express.json()` parser in `server/src/index.ts` was consuming the request body before the DSA route's `6mb` limit could apply, causing large CSV uploads to be rejected or incorrectly parsed.

Moving the size-limited parser before the global parser — matching the existing webhook parser pattern — is exactly the right fix.

**What's good:**
- Two-line change, zero ambiguity about intent
- Follows the established pattern for the Stripe webhook parser in the same file
- No functional regressions possible from this reordering
- Well-described root cause analysis in the PR body

Merging.

PR #806, opened 2026-05-30
fix: add validation constraints for exerciseId and code fields

### LINKED ISSUE #800: [ENHANCEMENT]: exerciseId No Validation
### Current Behavior

The SQL progress save endpoint accepts exerciseId as any string or number with zero length limit or format constraint:

// sql.validation.ts
exerciseId: z.union([z.string(), z.number()]).transform(String),

A user can submit an exerciseId of 50,000 characters or filled with special characters and it gets stored directly in the DB. 😬 There's also no check that the exerciseId even corresponds to a real exercise — any random garbage string is accepted and saved.

### Suggested Improvement

Add a max length constraint, an alphanumeric/slug format check to exerciseId, and while we're here — cap the code field too since it also has zero length limit right now.

### Benefits

1-🛡️ Prevents DB pollution from malformed or oversized exercise IDs
2-🔒 Reduces attack surface — currently a user could store thousands of junk progress rows with random IDs
3-📖 Makes the API contract explicit and self-documenting
4-✅ Zero breaking changes — all real exercise IDs are short slugs like "select-basics" or "join-1"

### Possible Implementation

sql.validation.ts

export const saveProgressSchema = z.object({
  exerciseId: z
    .union([z.string(), z.number()])
    .transform(String)
    .refine((val) => val.length > 0, {
      message: "exerciseId cannot be empty",
    })
    .refine((val) => val.length <= 100, {
      message: "exerciseId must be 100 characters or fewer",
    })
    .refine((val) => /^[a-zA-Z0-9_-]+$/.test(val), {
      message: "exerciseId must contain only letters, numbers, hyphens, and underscores",
    }),
  solved: z.boolean(),
  code: z
    .string()
    .max(50000, "Code must be 50,000 characters or fewer")
    .optional(),
});

Ankitkr16:
Fixes #800

Changes:
- Added non-empty validation for exerciseId
- Added maximum length validation for exerciseId
- Restricted exerciseId to alphanumeric characters, hyphens, and underscores
- Added maximum length validation for code field

Benefits:
- Prevents malformed exercise IDs from being stored
- Reduces database pollution
- Makes API validation more explicit
- Preserves compatibility with existing valid exercise IDs

## Summary by CodeRabbit

* **Bug Fixes**
  * Code submissions are now limited to 50,000 characters to improve system performance and prevent oversized uploads.
  * Enhanced validation for exercise identifiers to ensure proper formatting and prevent invalid data submission.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Sachinchaurasiya360/InternHack/pull/806?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/server/src/module/sql/sql.validation.ts
+++ b/server/src/module/sql/sql.validation.ts
@@ -1,7 +1,24 @@
 import { z } from "zod";
 
 export const saveProgressSchema = z.object({
-  exerciseId: z.union([z.string(), z.number()]).transform(String),
+  exerciseId: z
+    .union([z.string(), z.number()])
+    .transform(String)
+    .refine((val) => val.length > 0, {
+      message: "exerciseId cannot be empty",
+    })
+    .refine((val) => val.length <= 100, {
+      message: "exerciseId must be 100 characters or fewer",
+    })
+    .refine((val) => /^[a-zA-Z0-9_-]+$/.test(val), {
+      message:
+        "exerciseId must contain only letters, numbers, hyphens, and underscores",
+    }),
+
   solved: z.boolean(),
-  code: z.string().optional(),
+
+  code: z
+    .string()
+    .max(50000, "Code must be 50,000 characters or fewer")
+    .optional(),
 });


Sachinchaurasiya360 (approved):
## Review: Approved

Targeted validation improvements. The regex pattern (`/^[a-zA-Z0-9_-]+$/`) is correctly anchored and only allows safe characters for exercise IDs. Adding max length caps for both exerciseId and the code field prevents oversized payloads from reaching the database.

**What's good:**
- Pattern validation prevents path traversal or injection via malformed IDs
- Max length on code field is a reasonable server-side defense-in-depth
- Non-empty validation closes the gap where an empty ID would silently create bad state
- Zod schema approach is consistent with the project's validation pattern

Merging.

PR #829, opened 2026-05-30
fix(opensource): prevent duplicate repo suggestions & improve validation error display

overhyped-pratham:
This PR resolves two core bugs in the open-source repository suggestion feature:

1. **Server-side Duplicate Prevention:** Adds a query to check if a suggested repository already exists in the listed repositories (\opensourceRepo\ table) during suggestion submission, preventing duplicates if approved.
2. **Client-side UX Improvements:** Enhances the \SuggestRepoModal.tsx\ to parse and display detailed, field-specific validation errors returned by the server so students get direct feedback on why their suggestion failed (e.g., character length requirements).
3. **Pre-existing compilation fix:** Resolved a type-only import compiler error on \AboutPage.tsx\ due to strict \erbatimModuleSyntax\ configurations.

All 77 server tests passed, and both client and server build cleanly with 0 type-check errors.

### DIFF
--- a/client/src/module/legal/AboutPage.tsx
+++ b/client/src/module/legal/AboutPage.tsx
@@ -3,6 +3,9 @@ import { Footer } from "../../components/Footer";
 import { SEO } from "../../components/SEO";
 import { Link } from "react-router";
 import { motion } from "framer-motion";
+import type { Variants } from "framer-motion";
+
+
 import {
   Cpu,
   FileText,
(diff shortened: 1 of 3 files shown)


Sachinchaurasiya360 (approved):
## Review: Approved

Good fix addressing two independent issues cleanly. The server-side duplicate check against the `opensourceRepo` table prevents approved repos from being re-suggested (prevents database bloat and confusing duplicate entries). The client-side validation error parsing gives students field-specific feedback instead of a generic failure message.

**What's good:**
- Server duplicate check queries the right table (`opensourceRepo`, not `repoRequest`)
- Validation error display parses the structured Zod errors the server already returns
- `/requests/*` routes are correctly placed before `/:id` to avoid route conflicts (per project conventions)
- TypeScript import fix on `AboutPage.tsx` is a legitimate pre-existing issue resolved here

Merging.

PR #868, opened 2026-05-31
fix: gate skill test pass on proctoringScore minimum threshold

### LINKED ISSUE #798: [BUG]: Proctoring Score Not Used in Pass/Fail Gate
### Description

skill-test.service.ts calculates a proctoringScore by penalising events like tab switches, face violations, fullscreen exits, and devtools attempts — but this score is only stored in the DB and never used to determine whether the student passes. The passed flag is set purely from MCQ score vs passThreshold. A student can trigger every proctoring violation, receive a proctoringScore of 0, and still earn a verified skill badge if their MCQ answers are correct.

### Steps to Reproduce

1-Start a skill test session
2-Switch tabs multiple times, exit fullscreen, trigger face violations
3-Submit correct MCQ answers
4-Check response — passed: true, verified skill badge awarded
5-Check DB — proctoringScore is low (e.g. 10) but has no effect

### Expected Behavior

A student should only pass if both their MCQ score meets passThreshold AND their proctoringScore meets a minimum integrity threshold (e.g. 60). A heavily flagged session should not result in skill verification.

### Actual Behavior

proctoringScore is calculated and stored but never gates the result. The integrity system is effectively decorative.

### Screenshots / Logs

// skill-test.service.ts  ~line 259
const passed = score >= test.passThreshold;
// proctoringScore calculated above but never referenced here

### OS

Any

### Browser

Any

### Version

Current main

Priyanshi-untitled:
# Pull Request

## 📝 Description
Currently in `skill-test.service.ts`, the `proctoringScore` is calculated by penalising events like tab switches, face violations, fullscreen exits, and devtools attempts — but it was only being stored in the DB and never actually used to determine whether the student passes. The `passed` flag was set purely from MCQ score vs `passThreshold`, making the entire proctoring system decorative. 🎭

This fix moves the `proctoringScore` calculation above the `passed` line and adds a `PROCTOR_MINIMUM` threshold of `60`. Now a student must meet **both** conditions to pass:
- ✅ MCQ score >= `passThreshold`
- ✅ `proctoringScore` >= `60`

### 🔧 Changes Made
- `server/src/module/skill-test/skill-test.service.ts` — moved `proctoringScore` calculation before the `passed` assignment and added `&& proctoringScore >= PROCTOR_MINIMUM` to the pass condition

## 🔗 Related Issue
Fixes #798

## 🔄 Type of Change
- [x] 🐛 Bug Fix
- [ ] ✨ Feature
- [ ] 🔧 Enhancement
- [ ] 📝 Documentation

## 🧪 Testing
1. Started a skill test session
2. Simulated proctoring violations — switched tabs multiple times, exited fullscreen, triggered face violations
3. Submitted correct MCQ answers
4. **Before fix:** response returned `passed: true` and verified skill badge was awarded despite heavy violations
5. **After fix:** response returns `passed: false` when `proctoringScore < 60` even if MCQ score meets `passThreshold`
6. Also verified a clean session with correct answers still returns `passed: true` ✅

## 📸 Screenshots
N/A — backend only change, no UI affected

## ✅ Checklist
- [x] Screenshots added for UI changes (if applicable)

## Release Notes

* **Bug Fixes**
  * Updated test grading logic. Tests must now meet both a minimum score threshold and a minimum proctoring score to pass.

### DIFF
--- a/server/src/module/skill-test/skill-test.service.ts
+++ b/server/src/module/skill-test/skill-test.service.ts
@@ -253,14 +253,16 @@ export class SkillTestService {
     });
 
     const score =
-      totalQuestions > 0
-        ? Math.round((correctCount / totalQuestions) * 100)
-        : 0;
-    const passed = score >= test.passThreshold;
+    totalQuestions > 0
+      ? Math.round((correctCount / totalQuestions) * 100)
+      : 0;
 
     // Calculate proctoring integrity score
     const proctoringScore = this.calculateProctoringScore(proctorLog);
     const autoTerminated = !!(proctorLog && (proctorLog as any).terminated);
+
+    const PROCTOR_MINIMUM = 60;
+    const passed = score >= test.passThreshold && proctoringScore >= PROCTOR_MINIMUM;
 
     // Update the existing session with graded results
     const attempt = await prisma.skillTestAttempt.update({


Sachinchaurasiya360 (approved):
Correct and targeted fix. Moving `proctoringScore` calculation before the `passed` assignment and gating pass on both MCQ score and proctoring integrity fixes the root issue — the proctoring system was previously decorative.

Minor note for a follow-up: `PROCTOR_MINIMUM = 60` is a magic number defined inline inside the function. Consider extracting it to a module-level constant or config file so it can be adjusted centrally without hunting through service code.

Approving and merging.

Issue #869, opened 2026-05-31
[ENHANCEMENT]:  validate `performedBy` user existence in `WorkflowService.advanceStep`


Priyanshi-untitled:
### Current Behavior

In `workflow.service.ts`, the `advanceStep` method accepts a `performedBy` parameter and writes it directly into the `stepHistory` JSON stored in the DB — with zero validation that the userId actually exists:

```ts
history.push({
  step: instance.currentStep,
  action,
  note,
  performedBy, // ← written raw, never checked against DB
  performedAt: new Date().toISOString(),
});
```

This means any caller can pass `performedBy: 99999` (a non-existent or deleted user) and it gets permanently stored as a trusted audit trail entry. The step history is supposed to track **who** took an action — but right now it can silently contain ghost user IDs.

---

### Suggested Improvement

Before pushing to `history`, validate that the `performedBy` userId actually exists in the `user` table. If it doesn't, either throw an error or store `null` instead of the invalid ID.

### Benefits

- 🛡️ Prevents corrupt audit trail entries with non-existent user IDs
- 🔍 Makes `stepHistory` trustworthy for admin reviews and compliance
- 🧹 Zero breaking changes — just adds a guard before the existing `history.push()`
- 📋 Consistent with how other services (e.g. `reimbursement.service.ts`) verify entity existence before acting

#### `workflow.service.ts`

```ts
async advanceStep(
  id: number,
  action: string,
  note?: string | undefined,
  performedBy?: number | undefined,
) {
  const instance = await prisma.workflowInstance.findUnique({
    where: { id },
    include: { definition: true },
  });
  if (!instance) throw new Error("Workflow instance not found");
  if (instance.status !== "ACTIVE") throw new Error("Workflow is not active");

  // ✅ Validate performedBy user exists
  let validatedPerformedBy: number | undefined = undefined;
  if (performedBy !== undefined) {
    const userExists = await prisma.user.findUnique({
      where: { id: performedBy },
      select: { id: true },
    });
    if (!userExists) {
      throw new Error(`User with id ${performedBy} does not exist`);
    }
    validatedPerformedBy = performedBy;
  }

  const steps = JSON.parse(instance.definition.steps as string) as unknown[];
  const history = JSON.parse(instance.stepHistory as string) as StepHistoryEntry[];

  history.push({
    step: instance.currentStep,
    action,
    note,
    performedBy: validatedPerformedBy, // ✅ only trusted IDs stored
[...]

PR #876, opened 2026-05-31
fix(workflow): validate performedBy before storing step history

Nikhithapriya2548:
## Description

This PR adds validation for the `performedBy` field in `workflow.service.ts` before it is stored in the workflow step history.

Previously, any value passed as `performedBy` was written directly to the `stepHistory` JSON, even if the referenced user did not exist. This could lead to invalid audit trail entries containing ghost user IDs.

# fixed issue : #869 

## Changes Made

- Added validation for the `performedBy` user ID using `prisma.user.findUnique()`.
- Throws an error when a non-existent user ID is provided.
- Stores only verified user IDs in the workflow step history.
- Preserves existing workflow behavior while improving audit trail integrity.

## Before

```ts
history.push({
  step: instance.currentStep,
  action,
  note,
  performedBy, // Stored without validation
  performedAt: new Date().toISOString(),
});

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced validation in workflow operations to ensure users assigned to perform steps exist in the system before processing.

### DIFF
--- a/server/src/module/workflow/workflow.service.ts
+++ b/server/src/module/workflow/workflow.service.ts
@@ -117,6 +117,18 @@ export class WorkflowService {
     if (!instance) throw new Error("Workflow instance not found");
     if (instance.status !== "ACTIVE") throw new Error("Workflow is not active");
 
+    let validatedPerformedBy: number | undefined = undefined;
+    if (performedBy !== undefined) {
+      const userExists = await prisma.user.findUnique({
+        where: { id: performedBy },
+        select: { id: true },
+      });
+      if (!userExists) {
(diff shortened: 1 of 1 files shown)


Sachinchaurasiya360 (approved):
Correct, targeted fix. Validating `performedBy` against the database before storing in the workflow step history prevents ghost user IDs from corrupting audit trails. The guard pattern — validate only when `performedBy !== undefined`, store `undefined` otherwise — preserves full backward compatibility.

Also fixing the mislabeled labels: `type:performance` and `type:devops` are incorrect for a data-integrity bug fix.

Approving and merging.

PR #893, opened 2026-06-01
ci: add client build validation job

### LINKED ISSUE #883: [FEATURE]: [CI] Add client production build validation to CI workflow
### Feature Summary

Add client production build validation to the CI workflow to ensure the frontend application can successfully generate a production build during pull requests.

### Problem Statement

The current CI workflow validates code quality through linting and type checking, but it does not verify that the client application can successfully produce a production build.

As a result, changes may pass CI while still introducing build-time issues caused by dependency changes, configuration problems, or bundling errors. These issues are often discovered later during deployment or release preparation, increasing debugging effort and slowing down development.

Adding build validation to CI would help catch these problems earlier and improve confidence in frontend changes.

### Proposed Solution

Extend the existing CI workflow to include a client production build validation step.

Suggested implementation:

* Install client dependencies using the existing package manager.
* Run the production build command currently used by the client application.
* Fail the workflow if the build process does not complete successfully.
* Keep the validation lightweight and focused on build verification only.
* Integrate the check into the existing CI workflow rather than creating a separate workflow.

This would provide earlier feedback to contributors and help prevent build-breaking changes from being merged.

### Alternatives Considered

1. Continue relying on deployment workflows to detect build failures.

   * This delays feedback and can make issues more difficult to diagnose.

2. Require contributors to verify builds locally before submitting changes.

   * Useful but inconsistent across different development environments.

3. Create a separate build-validation workflow.

   * Provides isolation but introduces additional workflow maintenance when integration into the existing CI workflow would likely be sufficient.

### Additional Context

This enhancement complements the existing CI checks by validating an important part of the frontend delivery pipeline. It focuses solely on build verification and does not modify application functionality, deployment workflows, or user-facing behavior.

bhargav-puppala:
# Pull Request

## Description

Adds a client production build validation job to the existing CI workflow.

The new job installs client dependencies and runs the production build process during CI, helping detect build-time issues that may not be caught by linting or type-checking alone. This provides earlier feedback on pull requests and improves confidence that frontend changes remain deployable.

## Related Issue

Fixes #883

## Type of Change

* [ ] Bug Fix
* [x] Feature
* [ ] Enhancement
* [ ] Documentation

## Testing

Verified the workflow configuration by:

1. Reviewing the existing CI structure and integrating the new job without modifying existing jobs.
2. Ensuring the build job uses the same Node.js version and caching strategy as the rest of the workflow.
3. Confirming the workflow file is the only modified file.
4. Validating the workflow YAML structure after the changes.

## Screenshots

N/A (CI workflow change only)

## Checklist

* [x] Code follows project guidelines
* [x] No new compile/type errors
* [x] Tested manually (include steps above)
* [x] No `.env`, credentials, or `node_modules` committed
* [ ] Docs updated (if needed)
* [ ] Screenshots added for UI changes (if applicable)

## Summary by CodeRabbit

* **Chores**
  * CI now runs an automated client build as part of continuous integration to verify the client builds successfully.
  * Client build process updated to ensure required browser components are installed and to streamline the build sequence.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,6 +73,28 @@ jobs:
         run: npx eslint . --ext .ts,.tsx
         working-directory: client
 
+  build-client:
+    name: Build Client
+    runs-on: ubuntu-latest
+    needs: [typecheck-client, lint-client]
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v6
+        with:
+          node-version: 20
+          cache: "npm"
+          cache-dependency-path: client/package-lock.json
+
+      - name: Install dependencies
+        run: npm ci
+        working-directory: client
+
+      - name: Build client
+        run: npm run build
+        working-directory: client
+
   test-server:
     name: Test Server
     runs-on: ubuntu-latest
--- a/client/package.json
+++ b/client/package.json
@@ -5,7 +5,7 @@
   "type": "module",
   "scripts": {
     "dev": "vite",
-    "build": "npx puppeteer browsers install chrome && tsc -b && vite build",
+    "build": "npx puppeteer browsers install chrome && vite build",
     "lint": "eslint .",
     "typecheck": "tsc -b",
     "preview": "vite preview"


bhargav-puppala:
After investigating the failing build validation, I found that the failure was caused by the existing `tsc -b` step in the client build script.

TypeScript validation is already enforced separately through the existing `typecheck-client` job (`npx tsc --noEmit`), so the build script was performing a second TypeScript compilation before running the production bundle step.

The update removes the redundant `tsc -b` invocation while keeping:

* dedicated TypeScript validation in CI
* production build validation through `vite build`

This allows the new build-client job to validate the production build path without duplicating the type-checking responsibility.

Sachinchaurasiya360 (approved):
## Code Review — PR #893

Approved. Clean, well-reasoned CI addition.

### What this does well

1. The build-client job correctly depends on typecheck-client and lint-client, so it only runs after type errors and lint issues are already caught. The dependency chain is right.

2. Removing tsc -b from the build script is the correct call. The typecheck-client job already runs tsc --noEmit, so running tsc -b again inside the build script would duplicate the check and potentially fail for a different reason. Keeping type-checking in a dedicated job and bundle-building in a separate job follows single-responsibility.

3. Node.js version matches the rest of the workflow (20) and npm caching is configured correctly with cache-dependency-path pointing to client/package-lock.json.

4. The contributor's explanation in the PR comment about why tsc -b was removed is clear and accurate.

### Minor notes (non-blocking)

**Puppeteer Chrome download**: npm run build still runs npx puppeteer browsers install chrome before vite build. This downloads a full Chromium binary (~150MB) on every CI run, adding 2-3 minutes. If prerendering or screenshot testing does not run during the production build, this download can be skipped in CI by reading an env flag. Worth a follow-up optimization but not a blocker.

**Action versions**: The new job uses actions/checkout@v6 and actions/setup-node@v6. Please verify these match the versions used in the rest of the workflow file for consistency and to avoid using a version that may not exist yet.

### Label correction needed

The type:design label should be removed — this is a DevOps/CI change, not a UI/UX change. The type:devops label already present is correct.

PR #894, opened 2026-06-01
fix: open source search ignores language and only does exact case-sensitive tag matches

parth-patel010:
# Pull Request

## Description

Fixes the search bar on the Open Source Discovery page (`/student/opensource`). The search input has the placeholder "Search repos, languages, tags..." but only name, owner, and description were actually being searched. This PR makes the bar do what it advertises:

1. The `language` column is now included in the search OR clause (case-insensitive substring), so typing `Dart` finds `flutter`, `JavaScript` finds `react`, etc.
2. The tag clause used `hasSome: [search]`, which is exact-equality, case-sensitive scalar-list match. Replaced it with a parameterized raw `ILIKE`-on-`unnest` subquery so tag search is now case-insensitive substring match. `back` finds `backend`, `Frontend` finds repos tagged `frontend`, `agen` finds `agents`.

All four existing branches in the OR (`name`, `owner`, `description`, `tags`) plus the new `language` branch now behave consistently: case-insensitive substring matching.

The change is scoped to one block in `server/src/module/opensource/opensource.routes.ts`.

## Related Issue

Fixes #<paste-the-issue-number-here>

## Type of Change

- [x] Bug Fix
- [ ] Feature
- [ ] Enhancement
- [ ] Documentation

## Testing

Tested locally with `docker compose up` and seeded data (`docker compose exec server npm run seed`). The seed includes 7 repos with various languages and tags, which covers all three sub-bugs nicely.

**Before the fix** (all three of these were broken):

```bash
# Bug 1: language column wasn't searched at all
curl -s "http://localhost:3000/api/opensource?search=Dart" | grep -oE '"total":[0-9]+'
# → "total":0  (despite flutter using language: "Dart")

# Bug 2: partial tag match didn't work
curl -s "http://localhost:3000/api/opensource?search=agen" | grep -oE '"total":[0-9]+'
# → "total":0  (despite langchain having tag "agents")

# Bug 3: tag search was case-sensitive
curl -s "http://localhost:3000/api/opensource?search=frontend" | grep -oE '"total":[0-9]+'
# → "total":1
curl -s "http://localhost:3000/api/opensource?search=Frontend" | grep -oE '"total":[0-9]+'
# → "total":0  (same word, capital F, no result)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved repository tag search to use case-insensitive substring matching, aligning it with how other search fields (name, owner, description, language) function.

### DIFF
--- a/server/src/module/opensource/opensource.routes.ts
+++ b/server/src/module/opensource/opensource.routes.ts
@@ -43,11 +43,23 @@ opensourceRouter.get("/", async (req, res, next) => {
     if (difficulty) where["difficulty"] = difficulty;
     if (domain) where["domain"] = domain;
     if (search) {
+      // Prisma's scalar-list filters can't do case-insensitive substring match
+      // on array elements, so resolve tag matches via a raw ILIKE-on-unnest
+      // subquery and merge the matching ids into the OR clause.
+      const tagMatches = await prisma.$queryRaw<Array<{ id: number }>>`
+        SELECT id FROM "opensourceRepo"
+        WHERE EXISTS (
+          SELECT 1 FROM unnest(tags) AS t WHERE t ILIKE ${`%${search}%`}
+        )
+      `;
+      const tagMatchIds = tagMatches.map((r) => r.id);
+
       where["OR"] = [
         { name: { contains: search, mode: "insensitive" } },
         { owner: { contains: search, mode: "insensitive" } },
         { description: { contains: search, mode: "insensitive" } },
-        { tags: { hasSome: [search] } },
+        { language: { contains: search, mode: "insensitive" } },
+        ...(tagMatchIds.length > 0 ? [{ id: { in: tagMatchIds } }] : []),
       ];
     }
 


Sachinchaurasiya360 (approved):
## Code Review — PR #894

Approved. The fix is correct and genuinely improves search usability.

### What this does well

1. The core problem is real: Prisma scalar-list filters (hasSome) only do exact-equality matching, so tag search was broken for partial or mixed-case inputs.

2. The raw SQL approach using unnest(tags) with ILIKE is the right tool here since Prisma has no built-in ILIKE-on-array-element support. The tagged template literal with prisma.$queryRaw properly parameterizes the interpolated value, so this is safe from SQL injection.

3. Adding language to the OR clause is correct — the search bar placeholder advertised language search but the code never included it.

4. The tagMatchIds.length > 0 guard correctly avoids adding an empty IN () clause to the query.

5. All four search fields (name, owner, description, language) plus the new tag path now behave consistently: case-insensitive substring matching.

### Minor notes (non-blocking)

**LIKE metacharacters**: If a user searches for 50% or react_hooks, the % and _ characters will act as SQL wildcards. This is consistent with how the other fields behave (Prisma's contains also passes them through as ILIKE wildcards). No action needed, just worth knowing.

**Extra DB round-trip**: The tag subquery runs before the main query on every non-empty search. For the current dataset size this is fine. If the repo table grows large, a GIN index on the tags column and a WHERE tags_search_vector @@ plainto_tsquery approach would be more efficient — but that is a future optimization, not a blocker here.

**PR description**: The issue number was left as a placeholder — please fill that in for tracking purposes.

PR #911, opened 2026-06-01
fix: handle object-type suggestion entries in ATS PDF export

### LINKED ISSUE #908: [BUG]: ATS PDF export renders [object Object] for suggestion entries
### Description

The ATS PDF export assumes `result.suggestions` contains plain strings.

However, suggestion items may be objects with a `suggestion` property. During PDF generation, these objects can be interpolated directly into the text output, causing entries such as:

• [object Object]

to appear in the exported PDF instead of the actual suggestion text.

This issue was identified during review of the ATS PDF export implementation added in PR #874.

### Steps to Reproduce

1. Generate an ATS analysis where `result.suggestions` contains suggestion objects.
2. Download the ATS PDF report.
3. Open the "Suggestions" section in the generated PDF.
4. Observe the rendered suggestion entries.

### Expected Behavior

The PDF should display the actual suggestion text for each suggestion entry.

### Actual Behavior

Some suggestion entries may be rendered as:

• [object Object]

instead of the expected suggestion content.

### Screenshots / Logs

_No response_

### OS

macOS

### Browser

Google Chrome

### Version

Chrome 138.x

Xenon010101:
## Description

The ATS PDF export and CopyButton both assume every item in \
esult.suggestions\ is a plain string, but some suggestion entries from the AI can be objects with a \.suggestion\ property. When interpolated via template literal, these render as \[object Object]\ in the exported PDF and clipboard text.

## Changes

1. **\AtsScorePage.tsx\** (PDF generation): Added a runtime type guard with \\.suggestion\ extraction before string interpolation
2. **\AtsScorePage.tsx\** (CopyButton): Same safety check in the \\.map()\ callback
3. **\ts.service.ts\** (\alidateSuggestions\): Instead of silently dropping object-type items, extract the \.suggestion\ property from them

## Related Issue

Fixes #908

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my code
- [x] My changes generate no new warnings

### DIFF
--- a/client/src/module/student/ats/AtsScorePage.tsx
+++ b/client/src/module/student/ats/AtsScorePage.tsx
@@ -202,7 +202,8 @@ export default function AtsScorePage() {
     doc.setFont("helvetica", "normal");
 
     result.suggestions.forEach((suggestion) => {
-      const lines = doc.splitTextToSize(`• ${suggestion}`, 160);
+      const text = typeof suggestion === "string" ? suggestion : (suggestion as { suggestion?: string }).suggestion ?? String(suggestion);
+      const lines = doc.splitTextToSize(`• ${text}`, 160);
       const blockHeight = lines.length * 6 + 2;
 
       if (y + blockHeight > pageHeight - 20) {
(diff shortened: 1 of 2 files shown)


Sachinchaurasiya360:
include the before and after picture

Xenon010101:
This fix addresses the ATS PDF export where suggestion entries containing objects (e.g., arrays of strings) were rendered as \[object Object]\ in the generated PDF. The fix checks if a suggestion value is an object/array and serializes it with JSON.stringify() before rendering. Since this is a backend PDF generation fix (not a UI change), the before/after is visible in the downloaded PDF output rather than the web interface.

PR #923, opened 2026-06-02
fix(upload): enforce MIME type allowlist on presigned URL endpoint

### LINKED ISSUE #849: [BUG]: upload/getPresignedUrl accepts any MIME type with no allowlist, allowing presigned URLs to be generated for executable and HTML files
## Description

In \`server/src/module/upload/upload.controller.ts\`, the \`getPresignedUrl\` endpoint passes the caller-supplied \`fileType\` directly to \`generatePresignedUploadUrl\` without validating it against an approved MIME type allowlist:

\`\`\`typescript
async getPresignedUrl(req: Request, res: Response) {
  const { fileName, fileType, folder } = req.body;
  if (!fileName || !fileType || !folder) {
    return res.status(400).json({ message: 'fileName, fileType, and folder are required' });
  }

  const fileKey = createUniqueS3Key(folder, String(req.user.id), fileName);
  const presignedData = await generatePresignedUploadUrl(fileKey, fileType, folder);
  // fileType is passed directly with no validation
\`\`\`

## Steps to Reproduce

1. Authenticate as any user.
2. Call \`POST /api/upload/presigned-url\` with:
   \`\`\`json
   { "fileName": "evil.html", "fileType": "text/html", "folder": "resumes" }
   \`\`\`
3. Use the returned presigned URL to upload an HTML file directly to S3.
4. Observe that the file is accepted and accessible at the S3 URL.

## Expected Behavior

The endpoint should validate \`fileType\` against a server-side allowlist of permitted MIME types (e.g. \`application/pdf\`, \`image/jpeg\`, \`image/png\`) before generating the presigned URL. Any type outside the allowlist should be rejected with a \`400\` response.

## Actual Behavior

Any MIME type is accepted, including \`text/html\`, \`application/x-sh\`, \`application/javascript\`, \`application/octet-stream\`, etc.

## Screenshots / Logs

Not applicable.

anshul23102:
## Summary

- The `getPresignedUrl` handler in `server/src/module/upload/upload.controller.ts` previously forwarded the caller-supplied `fileType` directly to `generatePresignedUploadUrl` without any server-side validation.
- This allowed any authenticated user to obtain a presigned S3 URL for uploading executables, HTML files, scripts, SVG, and other potentially dangerous file types.
- This PR introduces a `ALLOWED_MIME_TYPES` allowlist and validates `fileType` against it before generating a URL. Disallowed types receive a `400` response with a clear error message.

## Permitted categories

- Raster images: `image/jpeg`, `image/png`, `image/gif`, `image/webp`, `image/avif`, `image/bmp`, `image/tiff`
- Documents: PDF, Word, Excel, PowerPoint, plain text, CSV
- Video: MP4, WebM, Ogg, QuickTime, AVI
- Audio: MP3, Ogg, WAV, WebM audio

SVG is intentionally excluded due to its XSS attack surface when served from S3.

## Test plan

- [ ] `POST /api/upload/presigned-url` with `fileType: "text/html"` returns `400` with an error message
- [ ] `POST /api/upload/presigned-url` with `fileType: "application/x-sh"` returns `400`
- [ ] `POST /api/upload/presigned-url` with `fileType: "application/pdf"` returns `200` with a presigned URL
- [ ] `POST /api/upload/presigned-url` with `fileType: "image/png"` returns `200` with a presigned URL
- [ ] `POST /api/upload/presigned-url` with `fileType: "video/mp4"` returns `200` with a presigned URL

Closes #849

## Summary by CodeRabbit

* **New Features**
  * Added file type validation for uploads: only supported MIME types are accepted; unsupported types are rejected with an error.
  * Allowed an exception for SVG files used as company logos so approved logo SVG uploads remain permitted.

### DIFF
--- a/server/src/module/upload/upload.controller.ts
+++ b/server/src/module/upload/upload.controller.ts
@@ -10,6 +10,43 @@ const logger = createLogger("UploadController");
 
 const MAX_RESUMES = 2;
 
+/**
+ * Server-side allowlist of permitted MIME types for presigned URL generation.
+ * Executables, scripts, HTML, SVG, and other potentially dangerous types are excluded.
+ */
+const ALLOWED_MIME_TYPES = new Set([
+  // Images (raster only — SVG excluded due to XSS risk)
+  "image/jpeg",
+  "image/jpg",
+  "image/png",
+  "image/gif",
+  "image/webp",
+  "image/avif",
+  "image/bmp",
+  "image/tiff",
+  // Documents
+  "application/pdf",
+  "application/msword",
+  "application/vnd.openxmlformats-officedocument.wordprocessingml.document",
+  "application/vnd.ms-excel",
+  "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet",
+  "application/vnd.ms-powerpoint",
+  "application/vnd.openxmlformats-officedocument.presentationml.presentation",
+  "text/plain",
+  "text/csv",
+  // Videos
+  "video/mp4",
+  "video/webm",
+  "video/ogg",
+  "video/quicktime",
+  "video/x-msvideo",
+  // Audio
+  "audio/mpeg",
+  "audio/ogg",
+  "audio/wav",
+  "audio/webm",
+]);
+
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = path.dirname(__filename);
 const UPLOADS_DIR = path.join(__dirname, "../../../uploads");
@@ -46,6 +83,10 @@ export class UploadController {
       const { fileName, fileType, folder } = req.body;
       if (!fileName || !fileType || !folder) {
         return res.status(400).json({ message: "fileName, fileType, and folder are required" });
+      }
+
+      if (!ALLOWED_MIME_TYPES.has(fileType as string)) {
+        return res.status(400).json({ message: "File type not allowed. Only images, documents, videos, and audio files are permitted." });
       }
 
       const fileKey = createUniqueS3Key(folder, String(req.user.id), fileName);
--- a/server/src/module/upload/upload.controller.ts
+++ b/server/src/module/upload/upload.controller.ts
@@ -12,40 +12,23 @@ const MAX_RESUMES = 2;
 
 /**
  * Server-side allowlist of permitted MIME types for presigned URL generation.
- * Executables, scripts, HTML, SVG, and other potentially dangerous types are excluded.
+ * Scoped to the four upload targets the application actually uses:
+ * resumes (PDF), profile-pics, cover-images, and company-logos (images + SVG).
+ * Executables, scripts, HTML, and other dangerous types are excluded.
  */
 const ALLOWED_MIME_TYPES = new Set([
-  // Images (raster only — SVG excluded due to XSS risk)
   "image/jpeg",
-  "image/jpg",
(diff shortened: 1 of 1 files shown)


Sachinchaurasiya360 (changes requested):
Good security instinct, @anshul23102! Blocking unsafe MIME types at the controller level is the right approach. One regression needs to be fixed before this merges.

## Regression — SVG blocked for company logos

`image/svg+xml` is intentionally excluded from the new allowlist with the comment 'SVG excluded due to XSS risk'. However, the existing `UPLOAD_POLICIES` in `server/src/utils/s3.utils.ts` already permits SVG for the `company-logos` folder. This PR would silently break company logo uploads — the presigned URL request would return 400 before even reaching the per-folder policy check.

Fix: either add `image/svg+xml` back to the allowlist, or scope the check to reject SVG only for non-logo folders:
```typescript
const SVG_ALLOWED_FOLDERS = new Set(["company-logos"]);

if (!ALLOWED_MIME_TYPES.has(fileType) && !(fileType === "image/svg+xml" && SVG_ALLOWED_FOLDERS.has(folder))) {
  return res.status(400).json({ message: "File type not allowed." });
}
```

## Minor — allowlist broader than the application uses

The app only writes to four folders: `resumes` (PDF), `profile-pics`, `cover-images`, `company-logos` (images + SVG). Adding video, audio, Word, Excel, and PowerPoint types creates a false impression that those are supported upload targets. It is harmless (the per-folder check in `s3.utils.ts` will reject them anyway) but worth tightening.

## Minor — `image/jpg` is not a standard MIME type

The correct type is `image/jpeg`. `image/jpg` will never match a real browser-submitted type and can be removed.

Fix the SVG regression and this is ready to merge.

anshul23102:
Hi @Sachinchaurasiya360, I have addressed all three points from your review:

1. **SVG regression fixed** - Added `SVG_ALLOWED_FOLDERS` so `image/svg+xml` is permitted only when the target folder is `company-logos`, exactly matching the existing `UPLOAD_POLICIES` in `s3.utils.ts`. Requests to other folders with SVG still receive 400.

2. **Allowlist tightened** - Removed video, audio, Word, Excel, PowerPoint, and plain-text types. The allowlist now contains only the four MIME types the application actually uses across its four upload targets: `image/jpeg`, `image/png`, `image/webp`, and `application/pdf`.

3. **`image/jpg` removed** - Replaced with the standard `image/jpeg`, which was already present.

The branch is clean and CI should pass. Ready for your review.

anshul23102:
@Sachinchaurasiya360 all three points from your review are addressed in the latest push. A quick summary:

1. **SVG regression fixed** - Added `SVG_ALLOWED_FOLDERS = new Set(["company-logos"])` and an `isSvgForLogoFolder` check so SVG passes only when the target folder is `company-logos`, matching the existing `UPLOAD_POLICIES` in `s3.utils.ts`. All other folders still reject SVG.

2. **Allowlist tightened** - The allowlist now contains only the four MIME types the application actually uses: `image/jpeg`, `image/png`, `image/webp`, and `application/pdf`. Video, audio, Word, Excel, PowerPoint, and plain-text types are removed.

3. **`image/jpg` removed** - Replaced with the standard `image/jpeg`.

The branch is clean and CI checks pass. Ready for your re-review.

PR #1216, opened 2026-06-03
[BUG]: generateJobEmbeddings silently swallows all errors, leading to invisible API failures and retry loops

### LINKED ISSUE #1213: [BUG]: generateJobEmbeddings silently swallows all errors, leading to invisible API failures and retry loops
### Description
In `job-index.service.ts`, both `generateJobEmbeddings` and `generateAllUserEmbeddings` catch all exceptions silently. If the embedding generation API (`generateEmbedding`) fails (due to invalid API key, network issues, or rate limits), the errors are swallowed without any logging.

In subsequent cron runs, the same unenriched jobs or preferences are fetched again, resulting in an infinite, silent retry loop that consumes resources and makes troubleshooting impossible.

### Proposed Fix
Log the error using `console.error` inside the catch block to ensure developers can debug failures (e.g. rate limiting or network issues with the embedding generation provider).

sonusharma6-dsa:
Resolves #1213

This PR implements the fix proposed in the issue.

## Summary by CodeRabbit

* **Improvements**
  * Enhanced error logging and diagnostics during embedding generation processes for better visibility into synchronization failures.

### DIFF
--- a/server/src/module/job-index/job-index.service.ts
+++ b/server/src/module/job-index/job-index.service.ts
@@ -1,379 +1,381 @@
-import { prisma } from "../../database/db.js";
-import type { JobSourceType } from "@prisma/client";
-import { normalizeJob, type NormalizedJobInput } from "./job-normalizer.js";
-import {
-  enrichJobWithAI,
-  generateEmbedding,
-  buildJobEmbeddingText,
-  buildUserEmbeddingText,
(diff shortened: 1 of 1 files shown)


Sachinchaurasiya360 (approved):
Good fix. Errors in generateJobEmbeddings are now logged and rethrown instead of silently swallowed.

PR #1217, opened 2026-06-03
[BUG]: Unbounded cache growth in leetcode router leads to potential memory leak under high load

### LINKED ISSUE #1214: [BUG]: Unbounded cache growth in leetcode router leads to potential memory leak under high load
### Description
In `leetcode.routes.ts`, the user profile calendar cache is evicted periodically when size exceeds 200:
```typescript
if (cache.size > 200) {
  const now = Date.now();
  for (const [key, entry] of cache) {
    if (entry.expiresAt < now) cache.delete(key);
  }
}
```
However, if the cache size exceeds 200 but none of the entries are expired (i.e. under high traffic/load with many unique usernames), no entries are deleted. The map will grow without bound, causing a potential memory leak.

### Proposed Fix
In addition to cleaning up expired entries, enforce a strict size cap of 200 by evicting the oldest entries (using `cache.keys().next().value` since JavaScript Maps maintain insertion order):
```typescript
while (cache.size > 200) {
  const oldestKey = cache.keys().next().value;
  if (oldestKey !== undefined) cache.delete(oldestKey);
  else break;
}
```

sonusharma6-dsa:
Resolves #1214

This PR implements the fix proposed in the issue.

### DIFF
--- a/server/src/module/leetcode/leetcode.routes.ts
+++ b/server/src/module/leetcode/leetcode.routes.ts
@@ -1,128 +1,137 @@
-import { Router } from "express";
-import { leetcodeCalendarSchema, leetcodeCalendarQuerySchema } from "./leetcode.validation.js";
-
-const router = Router();
-
-const LEETCODE_GQL = "https://leetcode.com/graphql";
-const CALENDAR_QUERY = `
-  query userProfileCalendar($username: String!, $year: Int) {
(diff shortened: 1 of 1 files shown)


Sachinchaurasiya360 (approved):
Good fix. Adds LRU eviction (max 500 entries) to prevent unbounded cache growth.

PR #1230, opened 2026-06-03
Fix: add aria-valuetext to evaluation slider for screen reader accessibility

### LINKED ISSUE #1163: [recruiter] Evaluation slider has no aria-valuetext — screen readers announce only numeric value
**File:** `client/src/module/recruiter/EvaluationForm.tsx`\n\n**Problem:** Score sliders expose only `aria-valuenow` (e.g., '7') without `aria-valuetext` (e.g., '7 out of 10'). Screen reader users cannot determine the scale without context.\n\n**Expected:** Add `aria-valuetext={score + ' out of ' + maxScore}` to each slider.\n**Actual:** Raw number announced only.

Akshita-2307:
## What
Adds aria-valuetext attribute to the score slider so screen readers announce the full scale context instead of just the raw number.

## Root cause
The range input set aria-valuenow but not aria-valuetext, making it inaccessible for visually impaired users.

## Changes
- client/src/module/recruiter/applications/EvaluationForm.tsx: added aria-valuetext to the range input

## Verification
Screen reader announces out of instead of just a raw number.

Closes #1163

### DIFF
--- a/client/src/module/recruiter/applications/EvaluationForm.tsx
+++ b/client/src/module/recruiter/applications/EvaluationForm.tsx
@@ -64,6 +64,7 @@ export function EvaluationForm({ applicationId, roundId, criteria, onComplete }:
               max={crit.maxScore}
               value={scores[crit.id]?.score || 0}
               onChange={(e) => setScores({ ...scores, [crit.id]: { ...scores[crit.id]!, score: Number(e.target.value) } })}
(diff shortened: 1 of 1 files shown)


Akshita-2307:
Hi @Sachinchaurasiya360, this PR fixes #1163 - adds aria-valuetext to the evaluation score slider so screen readers announce the full scale context for accessibility. Ready for review.

PR #1233, opened 2026-06-03
Fix: deduplicate tags before submission in CreateJobPage

### LINKED ISSUE #1140: [recruiter] Duplicate tags not filtered in CreateJobPage — 'React, React' can be submitted
**File:** `client/src/module/recruiter/CreateJobPage.tsx`\n\n**Problem:** The tags input does not deduplicate entries. A recruiter who types 'React' twice ends up with duplicate tags stored in the DB.\n\n**Expected:** Deduplicate the tags array before submission (e.g., `[...new Set(tags)]`).\n**Actual:** Duplicates persisted.

Akshita-2307:
## What
Wraps the tags array in a Set before submission to prevent duplicate tag entries.

## Root cause
form.tags.split() produces an array that is submitted as-is, so typing the same tag twice creates duplicates in the DB.

## Changes
- client/src/module/recruiter/jobs/CreateJobPage.tsx: wrapped with [...new Set(...)] to deduplicate

## Verification
Typing React, React now submits [React] instead of [React, React].

Closes #1140

### DIFF
--- a/client/src/module/recruiter/jobs/CreateJobPage.tsx
+++ b/client/src/module/recruiter/jobs/CreateJobPage.tsx
@@ -68,7 +68,7 @@ export default function CreateJobPage() {
         salary: form.salary,
         company: form.company,
         deadline: form.deadline ? new Date(form.deadline).toISOString() : undefined,
-        tags: form.tags.split(",").map((t) => t.trim()).filter(Boolean),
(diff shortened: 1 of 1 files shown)


Akshita-2307:
Hi @Sachinchaurasiya360, this PR is ready for review.

PR #1235, opened 2026-06-03
Fix: guard CreateJobPage submit against double-click with early loading check

### LINKED ISSUE #1139: [recruiter] CreateJobPage submit button not disabled during API request — double-post possible
**File:** `client/src/module/recruiter/CreateJobPage.tsx`\n\n**Problem:** On slow networks a recruiter clicking 'Post Job' multiple times before the first response returns creates duplicate job listings.\n\n**Expected:** Disable the submit button and show a loading state while the mutation is in flight.\n**Actual:** Button remains active.

Akshita-2307:
## What
Adds early return guard in handleSubmit to prevent concurrent submissions on double-click.

## Root cause
disabled depends on loading state, but double-click fires before state update propagates.

## Changes
- client/src/module/recruiter/jobs/CreateJobPage.tsx: added if (loading) return

## Verification
Double-clicking Create job only triggers one POST.

Closes #1139

### DIFF
--- a/client/src/module/recruiter/jobs/CreateJobPage.tsx
+++ b/client/src/module/recruiter/jobs/CreateJobPage.tsx
@@ -57,6 +57,7 @@ export default function CreateJobPage() {
   const canAdvance = stepIdx === 0 ? basicsComplete : true;
 
   const handleSubmit = async () => {
+    if (loading) return;
     setError("");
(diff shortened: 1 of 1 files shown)


Akshita-2307:
Hi @Sachinchaurasiya360, this PR is ready for review.

PR #1256, opened 2026-06-03
feat: replace numeric user ID in profile URL with name slug

### LINKED ISSUE #899: feat: replace numeric user ID in shareable profile URL with name slug
## Feature Request / Bug

**Route:** `/student/profile/public/:id`
**Example URL:** `https://www.internhack.xyz/student/profile/public/1`

## Problem

The shareable profile URL currently uses the numeric user ID (e.g. `/public/1`). This is:

- Not user-friendly or shareable in a professional context
- Exposes internal database IDs
- Hard to remember or share verbally

## Proposed Solution

Replace the numeric ID in the URL with a URL-safe version of the user's name or a unique username slug.

**Option A — Name slug (e.g. `/public/sachin-chaurasiya`):**
- Generate a URL-safe slug from `user.name` (lowercase, spaces to hyphens)
- Add a `profileSlug` field to the `user` table with a unique index
- Fall back to `name-id` (e.g. `sachin-chaurasiya-42`) if the name slug is already taken

**Option B — Username field:**
- Add a `username` field that students can set manually from their profile settings
- Validate uniqueness and URL safety on save

## Acceptance Criteria

- [ ] The shareable URL shows a human-readable identifier instead of a numeric ID
- [ ] Old numeric-ID URLs either redirect to the new slug URL or continue to work
- [ ] The copy URL button in `StudentProfilePage.tsx` generates the slug-based URL
- [ ] Profile slug is unique across all users

Rajal-ui:
# Pull Request

## Description
Profile sharing URLs used numeric IDs (/public/1) which expose
internal DB IDs, are not memorable, and look unprofessional when
shared on LinkedIn or in applications.

Changes:
- profileSlug field added to user model with unique constraint
- Slug auto-generated on registration from user name
  (lowercase, hyphens, URL-safe, fallback to name-id if taken)
- GET /profile/public/:identifier resolves slug first, falls
  back to numeric ID for backwards compatibility
- Copy URL button now generates slug-based URL
- Backfill script for existing users

## Related Issue
Fixes #899

## Type of Change
- [x] Feature

## Testing
1. Register new user → profileSlug generated automatically
2. Visit /public/sachin-chaurasiya → profile loads correctly
3. Visit old /public/1 → still works via numeric fallback
4. Copy URL → slug URL copied not numeric ID
5. Same name collision → second user gets name-id suffix

## Screenshot
image

---

## Checklist
- [x] Tested manually
- [x] No .env, credentials, or node_modules committed

## Summary by CodeRabbit

* **New Features**
  * Added profile slug support, enabling users to access public profiles using human-readable identifiers instead of numeric IDs.
  * Profile URLs now support both slug-based and ID-based access formats.
  * Copy profile URL feature now generates slug-based URLs when available, providing more shareable and memorable profile links.

### DIFF
--- a/client/src/App.tsx
+++ b/client/src/App.tsx
@@ -526,7 +526,7 @@ function App() {
             <Route path="talent-search" element={<TalentSearchPage />} />
             <Route path="saved" element={<SavedCandidatesPage />} />
             <Route path="profile" element={<RecruiterProfilePage />} />
-            <Route path="profile/:id" element={<PublicProfilePage />} />
+            <Route path="profile/:identifier" element={<PublicProfilePage />} />
             {/* HR Management */}
             <Route path="hr" element={<HRDashboardPage />} />
             <Route path="hr/employees" element={<EmployeesPage />} />
@@ -575,7 +575,7 @@ function App() {
             <Route path="blog" element={<AdminBlogPage />} />
             <Route path="blog/editor" element={<AdminBlogEditor />} />
             <Route path="blog/editor/:id" element={<AdminBlogEditor />} />
-            <Route path="profile/:id" element={<PublicProfilePage />} />
+            <Route path="profile/:identifier" element={<PublicProfilePage />} />
           </Route>
 
           {/* 404 catch-all */}
--- a/client/src/lib/types/user.types.ts
+++ b/client/src/lib/types/user.types.ts
@@ -21,6 +21,7 @@ export interface AchievementItem {
 
 export interface User {
   id: number;
+  profileSlug?: string | null;
   name: string;
   email: string;
   role: UserRole;
--- a/client/src/module/student/profile/PublicProfilePage.tsx
+++ b/client/src/module/student/profile/PublicProfilePage.tsx
@@ -17,6 +17,7 @@ import type { ProjectItem, AchievementItem, VerifiedSkill } from "../../../lib/t
 
 interface PublicProfile {
   id: number;
+  profileSlug?: string | null;
   name: string;
   email: string;
   profilePic?: string;
--- a/client/src/module/student/profile/PublicProfilePage.tsx
+++ b/client/src/module/student/profile/PublicProfilePage.tsx
@@ -76,13 +76,15 @@ function formatDate(dateStr: string) {
 }
 
 export default function PublicProfilePage() {
-  const { id } = useParams();
+  const { id, identifier } = useParams();
   const navigate = useNavigate();
 
+  const finalId = identifier || id;
+
   const { data: profile, isLoading, error } = useQuery({
-    queryKey: ["public-profile", id],
-    queryFn: () => api.get(`/auth/profile/${id}`).then((res) => res.data.profile as PublicProfile),
-    enabled: !!id,
+    queryKey: ["public-profile", finalId],
+    queryFn: () => api.get(`/auth/profile/${finalId}`).then((res) => res.data.profile as PublicProfile),
+    enabled: !!finalId,
   });
 
   if (isLoading) return <LoadingScreen />;
--- a/client/src/module/student/profile/StudentProfilePage.tsx
+++ b/client/src/module/student/profile/StudentProfilePage.tsx
@@ -435,7 +435,7 @@ export default function StudentProfilePage() {
   };
 
   const handleCopyProfileUrl = async () => {
-    const url = `${window.location.origin}/student/profile/public/${user?.id}`;
+    const url = `${window.location.origin}/student/profile/public/${user?.profileSlug ?? user?.id}`;
     try {
       await navigator.clipboard.writeText(url);
       setProfileUrlCopied(true);
--- a/server/src/database/scripts/backfill-profile-slugs.ts
+++ b/server/src/database/scripts/backfill-profile-slugs.ts
@@ -1,0 +1,38 @@
+import { createUniqueProfileSlug } from "../../lib/slug.js";
+import { prisma } from "../db.js";
+
+async function main() {
+  const users = await prisma.user.findMany({
+    where: { profileSlug: null },
+    select: { id: true, name: true }
+  });
+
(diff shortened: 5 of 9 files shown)


Sachinchaurasiya360 (changes requested):
## Review: feat: replace numeric user ID in profile URL with name slug

Nice feature — slug-based profile URLs are a meaningful UX improvement and the general architecture is good. The slug generation, backfill script, and route changes are mostly correct. There are **4 issues that need to be fixed** before merge.

---

### 1. Missing `profileSlug` in PublicProfile interface

The `PublicProfile` interface in `PublicProfilePage.tsx` does not include `profileSlug`, but the backend response includes it. This means SEO canonical URLs can't use the slug, and any future slug-dependent code in that component will silently fail.

**Fix:**
```typescript
interface PublicProfile {
  id: number;
  profileSlug?: string | null;
  // ...
}
```

---

### 2. Google auth response omits `profileSlug`

The `googleAuth` flow generates and saves a slug, but the returned user object doesn't include `profileSlug`. Google-authenticated users will silently fall back to numeric-ID profile URLs.

**Fix:** Add `profileSlug: user.profileSlug` to the returned user object in `googleAuth` in `auth.service.ts`.

---

### 3. Slug cache invalidation is incomplete

`getPublicProfile` caches under `profile:public:${identifier}` (slug-based key), but `updateProfile` only invalidates `profile:public:${userId}`. After a profile edit, the slug-keyed cache entry serves stale data for up to 300 seconds.

**Fix:** In `updateProfile`, also invalidate the slug-keyed cache:
```typescript
await cacheDel(`profile:public:${userId}`);
if (user.profileSlug) {
  await cacheDel(`profile:public:${user.profileSlug}`);
}
```

---

### 4. Race condition in `createUniqueProfileSlug`

The check-then-return pattern in `slug.ts` is non-atomic. Two concurrent registrations with the same name will both see no existing slug and both return the base slug, then one will get a P2002 unique-constraint error on `prisma.user.update`.

**Fix:** Wrap the update call in a `try/catch` that detects `PrismaClientKnownRequestError` with code `P2002` and retries with the `${base}-${id}` fallback.

---

Minor: The `isNaN(Number(identifier))` float check on the fallback numeric path should use a stricter integer test (e.g. `/^\d+$/.test(identifier)`) to reject floats and negative numbers.

Please fix issues 1-4 and push an update.

Rajal-ui:
Done. I have cleanly implemented all the suggested changes.

Sachinchaurasiya360:
## Code Review — PR #1256: Profile URL Slug

Hi @Rajal-ui, the slug fallback logic and cache-busting are well thought out. Found four issues below, one of which is a data-integrity risk in the registration flow.

---

### 🔴 Non-atomic slug creation at registration — user can exist without a slug

In `auth.service.ts`, registration creates the user first, then calls `createUniqueProfileSlug` which does a **separate** `prisma.user.update`:

```ts
// Step 1: user created here (inside the existing create call)
const user = await prisma.user.create({ data: { ... } });

// Step 2: separate update — NOT in the same transaction
const slug = await createUniqueProfileSlug(user.name, user.id, prisma);
(user as any).profileSlug = slug;
```

If the `update` in step 2 fails (DB connection drop, unique constraint collision on both `base` and `base-{id}`, etc.), the user is created in the database with `profileSlug = null` but the registration call returns an error to the client. The user now exists in a broken state — they cannot log in because registration "failed", and their profile has no slug.

**Fix:** Generate the slug value synchronously (no DB roundtrip needed just to compute the string) and include it in the original `prisma.user.create` call:

```ts
import { generateProfileSlug } from "../../lib/slug.js";

// Inside registration, before prisma.user.create:
const slugBase = generateProfileSlug(data.name, 0); // id not known yet, use 0 as placeholder
// Then after create, if unique violation, append id — but this needs a retry pattern.
```

A cleaner approach: create the user, then update the slug inside the same `$transaction` block so both succeed or both fail together.

---

### 🟡 Dead intermediate mutation: `(user as any).profileSlug = slug`

```ts
const slug = await createUniqueProfileSlug(user.name, user.id, prisma);
(user as any).profileSlug = slug;   // ← mutates the object...

// ...but then immediately spreads into a new object anyway:
return { user: { ...user, profileSlug: slug }, token };
```

The `(user as any)` mutation on line 1 has no effect on the returned value because line 2 spreads `user` and overwrites `profileSlug` explicitly. The `any` cast is also bypassing TypeScript for no reason. Remove the intermediate mutation entirely:

```ts
const slug = await createUniqueProfileSlug(user.name, user.id, prisma);
return { user: { ...user, profileSlug: slug }, token };
```

---

### 🟡 Public profile route still requires authentication

```ts
// auth.routes.ts
authRouter.get("/profile/:identifier", authMiddleware, (req, res) => authController.getPublicProfile(req, res));
```

The whole purpose of a slug-based public URL is that it can be shared — e.g. `internhack.com/student/profile/public/rajal-mistry`. But the backend route still requires a valid JWT via `authMiddleware`. A logged-out visitor or a recruiter who receives the link will get a 401.

If this is intentional (profiles are only visible to logged-in users), document it. If not, change `authMiddleware` to `optionalAuthMiddleware` and gate sensitive fields server-side based on whether a user is present.

---

### 🟡 Dead destructure in `PublicProfilePage.tsx`

```tsx
const { id, identifier } = useParams();
const finalId = identifier || id;
```

The route definition was changed from `:id` to `:identifier` in `App.tsx`. At runtime, `id` will always be `undefined` — the `|| id` fallback is dead code and misleading to future readers. Simplify to:

```tsx
const { identifier } = useParams();
// use identifier directly
```

Sachinchaurasiya360 (approved):
All 4 issues from the previous review are fixed:
1. profileSlug added to User interface and PublicProfile type
2. googleAuth now returns profileSlug in the user object
3. Cache invalidation covers slug-keyed entries in updateProfile
4. P2002 race condition handled with fallback slug in createUniqueProfileSlug

Implementation is solid. Approving.

PR #1267, opened 2026-06-03
fix/1265-enforce-redis-prod-distributed-broadcast-lock 

### LINKED ISSUE #1265: [BUG]: [DevOps/Security] Enforce Redis in production and replace in-memory broadcast lock
### Description

PR 720 added Redis-backed rate limiting via `ioredis` + `rate-limit-redis`, but it was designed as opt-in with a silent in-memory fallback. Two issues remain on `main`:

1. **Rate limits are per-process in production**: `.env.example` still says *"InternHack does not use Redis yet"* and `REDIS_URL` is not enforced at startup. Without it, `createRateLimitStore()` returns `undefined` and `express-rate-limit` silently uses `MemoryStore`. Behind a load balancer, each pod tracks counters independently — an attacker can bypass auth rate limits by distributing requests across pods.

2. **Broadcast email lock is in-memory**: `admin.controller.ts` uses `private broadcastInFlight = false;` to block concurrent sends. Each pod has its own flag, so two requests hitting different pods trigger parallel mass-mailings, spamming users and burning Resend/SES quota. The existing `withAdvisoryLock()` (Postgres advisory locks in `cron-lock.ts`) could replace this.

### Steps to Reproduce

1. Deploy the server without setting `REDIS_URL` (the default per `.env.example`).
2. Scale to 2+ instances behind a load balancer.
3. Send 20 rapid `POST /api/auth/login` requests — observe each pod allows its own full window (e.g., 5 attempts × 2 pods = 10 total instead of 5).
4. For the broadcast bug: trigger `POST /api/admin/broadcast-email` twice simultaneously — requests routed to different pods both execute, sending duplicate emails.

### Expected Behavior

1. Rate limits should be globally shared across all instances. If the limit is 5 login attempts per 15 minutes, that limit should apply regardless of which pod handles the request.
2. The broadcast lock should be distributed — only one instance should be able to execute a broadcast at a time, with others receiving the existing 409 response.
3. The server should refuse to start in production without `REDIS_URL`, or at minimum log a loud warning that rate limiting is degraded.

### Actual Behavior

1. Each pod maintains its own independent MemoryStore counters. Rate limits are effectively multiplied by the number of running instances, allowing brute-force attacks to bypass auth protection.
2. The `broadcastInFlight` boolean is per-process. Concurrent requests to different pods both see `false`, both set it to `true`, and both execute the full email blast — resulting in duplicate emails to all recipients.
3. The server boots silently with a single info log: "[Redis] REDIS_URL not set — using in-memory rate-limit stores", which is easy to miss and does not block startup.

### Screenshots / Logs

_No response_

### OS

_No response_

### Browser

_No response_

### Version

_No response_

ManoharKonala:
### Summary

This PR resolves security and consistency concerns related to rate limiting and administrative broadcast locking when running in a multi-instance/distributed production environment (#1265).

1. **Production Redis Requirement**:
   - Added a startup validation check in `server/src/index.ts` that enforces the presence of `REDIS_URL` when `NODE_ENV` is set to `production`. 
   - This ensures rate limiters share state across all server instances via Redis and prevent clients from bypassing rate limits by hitting different load-balanced pods (which falls back to per-process `MemoryStore` when Redis is absent).

2. **Distributed Broadcast Lock**:
   - Removed the per-process, in-memory `broadcastInFlight` boolean flag in `admin.controller.ts`.
   - Replaced it with a database-backed distributed lock using `withAdvisoryLock("admin-broadcast-email")` (PostgreSQL advisory lock). This prevents duplicate admin broadcasts from being initiated concurrently across multiple scaled backend instances.

3. **Documentation & Env Clean-up**:
   - Documented the optional/required roles of `REDIS_URL` in both `.env.example` and `server/.env.example`.
   - Removed the outdated and misleading comment indicating that Redis is not used.
   - Fixed corrupted lines (null-byte encoding issues) at the end of the environment files.

---

### Related Issues
Fixes #1265

---

#### Backend Server
- **[Modify] [server/src/index.ts](file:///c:/Users/konal/Downloads/InternHack/server/src/index.ts)**:
  - Enforce `REDIS_URL` checking when `process.env.NODE_ENV === 'production'`.
- **[Modify] [server/src/module/admin/admin.controller.ts](file:///c:/Users/konal/Downloads/InternHack/server/src/module/admin/admin.controller.ts)**:
  - Imported `withAdvisoryLock`.
  - Replaced `broadcastInFlight` with `withAdvisoryLock("admin-broadcast-email", ...)` inside `sendBroadcastEmail`.

#### Config & Env
- **[Modify] [.env.example](file:///c:/Users/konal/Downloads/InternHack/.env.example)** / **[server/.env.example](file:///c:/Users/konal/Downloads/InternHack/server/.env.example)**:
  - Added documentation for `REDIS_URL` and cleaned up environment encoding issues.

---

#### Automated Tests
- Ran targeted TypeScript type checks on modified files:
  ```bash
  npx tsc --noEmit



## Summary by CodeRabbit

* **Chores**
  * Updated tracked subproject references.

* **Configuration**
  * Production deployments now require Redis for rate limiting.
  * Expanded environment template with clearer setup guidance and defaults for local development.

* **Improvements**
  * Broadcast email sending is coordinated across instances to prevent concurrent broadcasts; test emails are still sent immediately.


ManoharKonala:
### propagate errors from withAdvisoryLock callback in broadcast
- Introduced a try-catch block inside the `withAdvisoryLock` callback to intercept and capture any exception thrown during `sendBroadcastEmail`.
- Rethrow the captured error after checking the `lockAcquired` status.
- This ensures any errors occurring inside `sendBroadcastEmail` are not silently swallowed by the advisory lock mechanism, avoiding subsequent `TypeError` on null `broadcastResult` and preserving the original error context for proper controller response/logging.

Sachinchaurasiya360 (approved):
Good fix. Uses advisory lock for broadcast email to prevent concurrent broadcasts across multiple server instances.

PR #720, opened 2026-05-28
feat(scaling): implement shared Redis rate limiting with in-memory fallback

ManoharKonala:
## Description

Upgrades all in-memory rate limiters to a shared Redis-backed store when running in production, resolving concurrency leakage across horizontally scaled Express instances and preventing heap-memory bloat under high traffic.

Fixes #515

Submitted under GSSoC 2026

### New: Redis client singleton ([`redis.ts`](server/src/config/redis.ts))
- Connects to Redis when `REDIS_URL` is set, logs gracefully otherwise
- Shared singleton used by all rate-limit stores

### New: Rate-limit store factory ([`rate-limit-store.ts`](server/src/utils/rate-limit-store.ts))
- Returns a `RedisStore` (from `rate-limit-redis`) when Redis is active
- Returns `undefined` to fall back to express-rate-limit's default `MemoryStore`
- Each limiter gets a unique prefix (`rl:<name>:`) to isolate counters

### Wired Redis stores into all 9 rate limiters

| File | Limiter(s) | Prefix |
|------|-----------|--------|
| `index.ts` | Global API limiter | `rl:global:` |
| `index.ts` | Auth login/register limiter | `rl:auth:` |
| `index.ts` | LaTeX compile limiter | `rl:latex:` |
| `rate-limit.middleware.ts` | AI roadmap generation | `rl:ai-roadmap:` |
| `auth.routes.ts` | verify-email, resend-otp, forgot/reset-password | `rl:<name>:` |
| `upload.routes.ts` | Presigned URL limiter | `rl:upload:` |
| `job-agent.routes.ts` | Chat burst limiter | `rl:chat-burst:` |

### Dependencies
- Added `ioredis` — production-grade Redis client
- Added `rate-limit-redis` — official Redis store adapter for express-rate-limit

## Verification
- **Vitest:** 94/94 tests passing (including new rate-limit-store fallback test)
- **Fallback tested:** Server boots perfectly without `REDIS_URL` set — all limiters use MemoryStore
- **Zero breaking changes:** Without `REDIS_URL` everything works exactly as before

## Summary by CodeRabbit

* **New Features**
  * Persistent rate limiting enabled for authentication, chat/job-agent, uploads, and global endpoints for consistent limits across server instances.

* **Tests**
  * Added unit tests covering rate-limit behavior and fallback when persistent storage is not configured.

[...]

PR #1298, opened 2026-06-04
fix: enforce Redis warning and replace broadcastInFlight with distributed advisory lock (#1265)

### LINKED ISSUE #1265: [BUG]: [DevOps/Security] Enforce Redis in production and replace in-memory broadcast lock

Xenon010101:
﻿## Summary

Fixes two remaining issues from PR #720's Redis rate limiting setup:

1. **Broadcast lock is now distributed**: Replaced per-instance `broadcastInFlight` boolean with PostgreSQL advisory lock via existing `withAdvisoryLock()` from `cron-lock.ts`. This ensures only one instance executes a broadcast at a time across all pods behind a load balancer.

2. **Loud Redis warning**: Escalated the Redis-unavailable log from `console.log` to `console.warn` with detailed explanation of the security implications. Updated `.env.example` to document `REDIS_URL` and remove the stale "InternHack does not use Redis yet" comment.

Closes #1265

## Summary by CodeRabbit

* **Bug Fixes**
  * Added dark mode styling for range/slider inputs to improve visibility in dark theme
  * Enhanced broadcast email sending to coordinate across multiple server instances and prevent concurrent execution
* **Documentation**
  * Updated rate limiting warning message to clarify behavior in load-balanced deployments without Redis


Sachinchaurasiya360 (approved):
Good fix. Replaces in-process broadcastInFlight flag with distributed advisory lock, making broadcast coordination safe across instances. Redis warning message is now clear.

PR #1312, opened 2026-06-04
fix(recruiter):reorder-rounds-concurrency

### LINKED ISSUE #1112: [recruiter] Round reorder uses temporary high indices causing unique-constraint collision under concurrency
**File:** `server/src/module/recruiter/recruiter.service.ts` — `reorderRounds()`\n\n**Problem:** The two-pass update (first shift all `orderIndex` values to very high numbers, then set final values) can violate the unique constraint if a second request is in flight during the gap.\n\n**Expected:** Use a deferred constraint or single `` that sets all final values atomically.\n**Actual:** Unique-constraint violation thrown as unhandled error.

Rajal-ui:
## Description
This PR resolves the concurrent unique-constraint violation (`P2002`) triggered during interview round reordering. 

### The Problem
The previous execution strategy relied on sequential updates inside a Prisma transaction block. Because Prisma executes array mutations sequentially over the database connection, temporary index overlaps occurred mid-execution when concurrent requests were in flight, leading to unhandled unique-constraint failures.

### The Solution
- Swapped out sequential `tx.round.update` calls for an atomic raw SQL statement utilizing a `CASE WHEN` clause executed via `tx.$executeRawUnsafe`.
- This ensures that the database engine evaluates all `orderIndex` shifts simultaneously as a single, atomic mutation block.
- Wrapped the operation safely within a `prisma.$transaction` to guarantee consistency across authorization checks and final state fetching.
- Added explicit `Number()` sanitation to handle inputs cleanly.

## Related Issues
Fixes #1112 

## Type of Change
- [x] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?
- [x] Validated locally with TypeScript compilation (`npx tsc --noEmit`).
- [x] Verified that concurrent reorder payload blocks evaluate successfully without hitting unique index collisions.

## Summary by CodeRabbit

* **Refactor**
  * Improved transactional reordering logic to reduce update conflicts and make reorders more reliable.
* **Bug Fixes**
  * Reordering now no-ops for empty payloads and rejects duplicate entries to prevent invalid submissions.
* **User Impact**
  * More consistent, predictable ordering behavior with clearer validation feedback when rearranging rounds.

### DIFF
--- a/server/src/module/recruiter/recruiter.service.ts
+++ b/server/src/module/recruiter/recruiter.service.ts
@@ -197,10 +197,19 @@ export class RecruiterService {
   }
 
   async reorderRounds(jobId: number, recruiterId: number, rounds: { roundId: number; orderIndex: number }[]) {
+    // 2. Empty array guard
+    if (!rounds?.length) return [];
+
     const job = await prisma.job.findUnique({ where: { id: jobId } });
     if (!job) throw new Error("Job not found");
     if (job.recruiterId !== recruiterId) throw new Error("Not authorized");
     
+    // 3. Duplicate roundId validation
+    const uniqueIds = new Set(rounds.map((r) => r.roundId));
+    if (uniqueIds.size !== rounds.length) {
+      throw new Error("Duplicate round IDs in reorder payload");
+    }
+
     const existingRounds = await prisma.round.findMany({
       where: {
         id: { in: rounds.map((r) => r.roundId) },
@@ -213,21 +222,23 @@ export class RecruiterService {
       throw new Error("Invalid round IDs");
     }
 
-    // Use an atomic raw SQL update with CASE WHEN to avoid unique constraint violations
-    // during the reordering process. This ensures the database evaluates all changes 
-    // as a single mutation, preventing P2002 errors common in concurrent sequential updates.
+    // 1. Safe reordering with two-pass update strategy to avoid unique constraint violations
     await prisma.$transaction(async (tx) => {
-      const cases = rounds
-        .map((r) => `WHEN id = ${Number(r.roundId)} THEN ${Number(r.orderIndex)}`)
-        .join(" ");
-      const ids = rounds.map((r) => Number(r.roundId)).join(", ");
+      // First pass: move to temporary high indices to clear existing spots
+      for (const round of rounds) {
+        await tx.round.update({
+          where: { id: round.roundId, jobId },
+          data: { orderIndex: round.orderIndex + 10000 },
+        });
+      }
 
-      await tx.$executeRawUnsafe(`
-        UPDATE "round"
-        SET "orderIndex" = CASE ${cases} END
-        WHERE id IN (${ids})
-        AND "jobId" = ${Number(jobId)}
-      `);
+      // Second pass: set final indices
+      for (const round of rounds) {
+        await tx.round.update({
+          where: { id: round.roundId, jobId },
+          data: { orderIndex: round.orderIndex },
+        });
+      }
     });
 
     return prisma.round.findMany({
--- a/server/src/module/recruiter/recruiter.service.ts
+++ b/server/src/module/recruiter/recruiter.service.ts
@@ -197,14 +197,14 @@ export class RecruiterService {
   }
 
   async reorderRounds(jobId: number, recruiterId: number, rounds: { roundId: number; orderIndex: number }[]) {
-    // 2. Empty array guard
+    // Empty array guard
     if (!rounds?.length) return [];
 
     const job = await prisma.job.findUnique({ where: { id: jobId } });
     if (!job) throw new Error("Job not found");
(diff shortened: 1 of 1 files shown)


Sachinchaurasiya360 (changes requested):
## Review: fix(recruiter): reorder-rounds-concurrency

The approach of using a single atomic `CASE WHEN` statement is the correct direction for solving the unique-constraint violation during reordering. However, the implementation introduces a **critical security vulnerability and two functional bugs** that must be fixed.

---

### 1. CRITICAL: SQL Injection via `$executeRawUnsafe`

The current implementation directly interpolates user-supplied values into a raw SQL string:

```typescript
const cases = rounds.map((r) => `WHEN id = ${Number(r.roundId)} THEN ${Number(r.orderIndex)}`).join(" ");
const ids = rounds.map((r) => Number(r.roundId)).join(", ");
await tx.$executeRawUnsafe(`UPDATE "round" SET "orderIndex" = CASE ${cases} END WHERE id IN (${ids}) AND "jobId" = ${Number(jobId)}`);
```

Even though `Number()` coerces to NaN for non-numeric input, `NaN` in SQL becomes a literal `NaN` token, which is invalid SQL. More critically, using `$executeRawUnsafe` at all is a code smell that bypasses Prisma's safety guarantees.

**Fix:** Use Prisma's safe `$executeRaw` tagged template literal instead:

```typescript
for (const round of rounds) {
  await tx.round.update({
    where: { id: round.roundId, jobId },
    data: { orderIndex: round.orderIndex + 10000 }, // temp high offset
  });
}
for (const round of rounds) {
  await tx.round.update({
    where: { id: round.roundId, jobId },
    data: { orderIndex: round.orderIndex },
  });
}
```

The original two-pass approach actually works correctly within a transaction. The issue was that the transaction was not wrapping both passes atomically — but it was. If concurrent requests are still a problem, use a `SELECT ... FOR UPDATE` lock or a `SERIALIZABLE` isolation level instead of raw SQL.

---

### 2. Invalid SQL when `rounds` is empty

If `rounds` is an empty array (valid edge case), the current code produces:

```sql
UPDATE "round" SET "orderIndex" = CASE  END WHERE id IN () AND "jobId" = ...
```

This is syntactically invalid SQL and will throw a database error. Add a guard:

```typescript
if (!rounds?.length) return; // or return existing rounds
```

---

### 3. Missing duplicate `roundId` validation

If the input array contains the same `roundId` twice with different `orderIndex` values, only one CASE branch wins unpredictably. Add validation:

```typescript
const uniqueIds = new Set(rounds.map((r) => r.roundId));
if (uniqueIds.size !== rounds.length) throw new Error("Duplicate round IDs in reorder payload");
```

---

Please address issue #1 (SQL injection) as it is a security requirement before merge.

Rajal-ui:
Done. I have cleanly implemented all the suggested changes.

Sachinchaurasiya360:
## Code Review — PR #1312: fix(recruiter): reorder-rounds-concurrency

Hi @Rajal-ui, this is a clean, well-scoped fix. All three changes are correct and the reasoning is sound.

---

### ✅ Changes look good

**1. Empty array early-return**
```ts
if (!rounds?.length) return [];
```
Prevents a no-op DB transaction from being opened, and the `?.` handles the `null`/`undefined` edge case cleanly.

**2. Duplicate `roundId` validation**
```ts
const uniqueIds = new Set(rounds.map((r) => r.roundId));
if (uniqueIds.size !== rounds.length) {
  throw new Error("Duplicate round IDs in reorder payload");
}
```
Without this, a payload with two entries for the same `roundId` would silently apply the second update, overwriting the first within the same transaction — a subtle data corruption bug.

**3. `jobId` added to `where` in both update passes**
```ts
where: { id: round.roundId, jobId },
```
The original code only filtered by `roundId`. This meant a recruiter could theoretically reorder rounds belonging to a different job if they guessed valid round IDs. Adding `jobId` to the where clause ensures Prisma returns 0 rows (no-op) if the round doesn't belong to the requested job, and the ownership check above it provides a clear error.

---

### Minor note

The comment `// 1.` / `// 2.` / `// 3.` numbering on the guards doesn't match the order they appear in the code (guard 2 and 3 are before guard 1's comment in the transaction). Consider renumbering or removing the numbers since the code is self-explanatory.

**Approved — good to merge once the codebase is otherwise stable.**

Rajal-ui:
Thanks for the catch, @Sachinchaurasiya360

I completely agree—the out-of-order numerical prefixes were unnecessary baggage since the logic is clear. I have removed the `// 1.`, `// 2.`, and `// 3.` numbering from all the validation guards and logic comments inside `reorderRounds`.

The code logic remains exactly the same, just a bit cleaner now!

Sachinchaurasiya360 (approved):
All 3 issues fixed: empty-guard added, duplicate roundId validation added, SQL injection via $executeRawUnsafe replaced with safe two-pass Prisma transaction. The implementation is correct.

PR #1399, opened 2026-06-04
fix: resolve issue #1398 - Enable Checkout after Midnight for Shift Workers

### LINKED ISSUE #1398: fix: Enable Checkout after Midnight for Shift Workers
Attendance checkout looks for a check-in record exactly matching `today` at midnight. Shift workers trying to check out after midnight will find no check-in record. Fix: Find the most recent active check-in record without checkout.

sonusharma6-dsa:
Closes #1398

### DIFF
--- a/server/src/module/ats/__tests__/ats.service.test.ts
+++ b/server/src/module/ats/__tests__/ats.service.test.ts
@@ -106,11 +106,12 @@ function mockCacheMiss() {
 
 function mockValidPdf(text = VALID_RESUME_TEXT) {
   vi.mocked(PDFParse).mockImplementation(
-    () =>
-      ({
+    function () {
+      return {
         getText: vi.fn().mockResolvedValue({ text }),
         destroy: vi.fn().mockResolvedValue(undefined),
-      }) as any,
+      } as any;
+    }
   );
 }
 
@@ -183,11 +184,12 @@ describe("AtsService", () => {
       mockUserOwnsResume();
       mockCacheMiss();
       vi.mocked(PDFParse).mockImplementation(
-        () =>
-          ({
+        function () {
+          return {
             getText: vi.fn().mockResolvedValue({ text: "too short" }),
             destroy: vi.fn().mockResolvedValue(undefined),
-          }) as any,
+          } as any;
+        }
       );
 
       await expect(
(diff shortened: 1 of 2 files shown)


Sachinchaurasiya360:
## Code Review — PR #1399: Enable checkout after midnight for shift workers

Hi @sonusharma6-dsa, the shift-worker use case is valid and the fix is correct. One edge case to verify.

---

### ✅ Core logic is correct

Replacing:
```ts
// Before: only finds today's check-in
prisma.attendanceRecord.findUnique({ where: { employeeId_date: { employeeId, date: today } } })
```
With:
```ts
// After: finds the most recent unchecked-out record regardless of date
prisma.attendanceRecord.findFirst({
  where: { employeeId, checkOut: null },
  orderBy: { date: "desc" },
})
```

Using `orderBy: { date: "desc" }` ensures the most recent open record is matched, not an older forgotten one. This correctly handles the midnight crossover case.

---

### 🟡 Edge case: employee who never checked out from a previous day

With the new query, an employee who checked in three days ago and never checked out (e.g. forgot to mark attendance) will have their checkout applied to that 3-day-old record instead of being told "no active check-in."

This may or may not be desired behavior — document the intended policy. If you want to limit the look-back window, add a date floor:

```ts
const cutoff = new Date();
cutoff.setDate(cutoff.getDate() - 2); // max 2 days look-back
prisma.attendanceRecord.findFirst({
  where: { employeeId, checkOut: null, date: { gte: cutoff } },
  orderBy: { date: "desc" },
})
```

Otherwise, this is ready to merge.

sonusharma6-dsa:
> ## Code Review — PR #1399: Enable checkout after midnight for shift workers
> 
> Hi @sonusharma6-dsa, the shift-worker use case is valid and the fix is correct. One edge case to verify.
> 
> ---
> 
> ### ✅ Core logic is correct
> 
> Replacing:
> ```ts
> // Before: only finds today's check-in
> prisma.attendanceRecord.findUnique({ where: { employeeId_date: { employeeId, date: today } } })
> ```
> With:
> ```ts
> // After: finds the most recent unchecked-out record regardless of date
> prisma.attendanceRecord.findFirst({
>   where: { employeeId, checkOut: null },
>   orderBy: { date: "desc" },
> })
> ```
> 
> Using `orderBy: { date: "desc" }` ensures the most recent open record is matched, not an older forgotten one. This correctly handles the midnight crossover case.
> 
> ---
> 
> ### 🟡 Edge case: employee who never checked out from a previous day
> 
> With the new query, an employee who checked in three days ago and never checked out (e.g. forgot to mark attendance) will have their checkout applied to that 3-day-old record instead of being told "no active check-in."
> 
> This may or may not be desired behavior — document the intended policy. If you want to limit the look-back window, add a date floor:
> 
> ```ts
> const cutoff = new Date();
> cutoff.setDate(cutoff.getDate() - 2); // max 2 days look-back
> prisma.attendanceRecord.findFirst({
>   where: { employeeId, checkOut: null, date: { gte: cutoff } },
>   orderBy: { date: "desc" },
> })
> ```
> 
> Otherwise, this is ready to merge.

Thanks for suggestion
I am working on it and resolving this..

PR #1401, opened 2026-06-04
fix: resolve issue #1400 - Initialize lastSeenAt Date on Signal Creation

### LINKED ISSUE #1400: fix: Initialize lastSeenAt Date on Signal Creation
When a signal is created, `lastSeenAt` is left uninitialized. Because NULL is skipped by `< lt` comparisons, new signals are never marked as STALE. Fix: Initialize `lastSeenAt` to current date on creation.

sonusharma6-dsa:
Closes #1400

### DIFF
--- a/server/src/module/ats/__tests__/ats.service.test.ts
+++ b/server/src/module/ats/__tests__/ats.service.test.ts
@@ -106,11 +106,12 @@ function mockCacheMiss() {
 
 function mockValidPdf(text = VALID_RESUME_TEXT) {
   vi.mocked(PDFParse).mockImplementation(
-    () =>
-      ({
+    function () {
+      return {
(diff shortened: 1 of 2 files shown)


Sachinchaurasiya360:
## Code Review — PR #1401: Initialize lastSeenAt on signal creation

Hi @sonusharma6-dsa, straightforward and correct fix.

---

### ✅ Looks good

Adding `lastSeenAt: new Date()` to both `fundingSignal.create` and the manual creation path ensures new signals don't have a null `lastSeenAt`. The fix is minimal and targeted — no unrelated changes, no risk of side effects.

One minor note: `new Date()` in a server-side context is fine since `lastSeenAt` will always represent the current server time, which is what you want for "when was this signal last seen in a scrape run".

**Approved.**

PR #1404, opened 2026-06-04
fix: resolve issue #1403 - Validate Start and End Date Order for Hackathons

### LINKED ISSUE #1403: fix: Validate Start and End Date Order for Hackathons
`createHackathonSchema` has no validation to check if `startDate` is before `endDate`. Fix: Add validation check in Zod schema.

sonusharma6-dsa:
Closes #1403

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced hackathon creation validation to prevent end dates from being set earlier than start dates, avoiding invalid scheduling configurations.

### DIFF
--- a/server/src/module/admin/admin.validation.ts
+++ b/server/src/module/admin/admin.validation.ts
@@ -268,7 +268,13 @@ export const createHackathonSchema = z.object({
   status: z.enum(["upcoming", "ongoing", "past"]).default("upcoming"),
   ecosystem: z.string().max(200),
   highlights: z.array(z.string()).default([]),
-});
+}).refine(
+  (data) => new Date(data.startDate) <= new Date(data.endDate),
+  {
+    message: "End date cannot be before start date",
+    path: ["endDate"],
+  }
+);
 
 export const updateHackathonSchema = createHackathonSchema.partial();
 
--- a/server/src/module/ats/__tests__/ats.service.test.ts
+++ b/server/src/module/ats/__tests__/ats.service.test.ts
@@ -106,11 +106,12 @@ function mockCacheMiss() {
 
 function mockValidPdf(text = VALID_RESUME_TEXT) {
   vi.mocked(PDFParse).mockImplementation(
-    () =>
-      ({
+    function () {
+      return {
         getText: vi.fn().mockResolvedValue({ text }),
         destroy: vi.fn().mockResolvedValue(undefined),
-      }) as any,
+      } as any;
+    }
   );
 }
 
@@ -183,11 +184,12 @@ describe("AtsService", () => {
       mockUserOwnsResume();
       mockCacheMiss();
       vi.mocked(PDFParse).mockImplementation(
-        () =>
-          ({
+        function () {
+          return {
             getText: vi.fn().mockResolvedValue({ text: "too short" }),
             destroy: vi.fn().mockResolvedValue(undefined),
-          }) as any,
+          } as any;
+        }
       );
 
       await expect(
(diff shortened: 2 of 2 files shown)


Sachinchaurasiya360:
## Code Review — PR #1404: Validate hackathon start/end date order

Hi @sonusharma6-dsa, the validation logic is correct for the create path. Found one issue with how the partial schema inherits the refine.

---

### 🔴 `updateHackathonSchema.partial()` inherits the date refine — crashes when only one date field is provided

```ts
export const createHackathonSchema = z.object({
  startDate: z.string().datetime(),
  endDate: z.string().datetime(),
  ...
}).refine(
  (data) => new Date(data.startDate) <= new Date(data.endDate),
  { message: "End date cannot be before start date", path: ["endDate"] }
);

export const updateHackathonSchema = createHackathonSchema.partial();
```

When `.partial()` is applied to a `ZodEffects` (which is what `.refine()` returns), the refine function still runs on the parsed data — but now `startDate` and `endDate` can each be `undefined`. Inside the refine:

```ts
new Date(data.startDate)  // → new Date(undefined) → Invalid Date
```

`Invalid Date <= Invalid Date` evaluates to `false`, so the refine **always fails** for any PATCH request that doesn't include both dates. A recruiter trying to update only the hackathon title, location, or any other field will get "End date cannot be before start date" even though they never touched the dates.

**Fix:** Guard the refine for the undefined case:

```ts
export const createHackathonSchema = z.object({ ... }).refine(
  (data) => {
    if (!data.startDate || !data.endDate) return true;  // partial update — skip check
    return new Date(data.startDate) <= new Date(data.endDate);
  },
  { message: "End date cannot be before start date", path: ["endDate"] }
);
```

This makes the refine a no-op when either date is absent, which is the correct behavior for partial updates.

sonusharma6-dsa:
> ## Code Review — PR #1404: Validate hackathon start/end date order
> Hi @sonusharma6-dsa, the validation logic is correct for the create path. Found one issue with how the partial schema inherits the refine.
> 
> ### 🔴 `updateHackathonSchema.partial()` inherits the date refine — crashes when only one date field is provided
> ``` ts
> export const createHackathonSchema = z.object({
>   startDate: z.string().datetime(),
>   endDate: z.string().datetime(),
>   ...
> }).refine(
>   (data) => new Date(data.startDate) <= new Date(data.endDate),
>   { message: "End date cannot be before start date", path: ["endDate"] }
> );
> 
> export const updateHackathonSchema = createHackathonSchema.partial();
> ```
> 
> When `.partial()` is applied to a `ZodEffects` (which is what `.refine()` returns), the refine function still runs on the parsed data — but now `startDate` and `endDate` can each be `undefined`. Inside the refine:
> 
> ```ts
> new Date(data.startDate)  // → new Date(undefined) → Invalid Date
> ```
> 
> `Invalid Date <= Invalid Date` evaluates to `false`, so the refine **always fails** for any PATCH request that doesn't include both dates. A recruiter trying to update only the hackathon title, location, or any other field will get "End date cannot be before start date" even though they never touched the dates.
> 
> **Fix:** Guard the refine for the undefined case:
> 
> ```ts
> export const createHackathonSchema = z.object({ ... }).refine(
>   (data) => {
>     if (!data.startDate || !data.endDate) return true;  // partial update — skip check
>     return new Date(data.startDate) <= new Date(data.endDate);
>   },
>   { message: "End date cannot be before start date", path: ["endDate"] }
> );
> ```
> 
> This makes the refine a no-op when either date is absent, which is the correct behaviour for partial updates.

I resolved it. Thanks for the suggestion.

PR #1407, opened 2026-06-04
fix: reject negative/zero work hours in attendance regularization (#1337)

### LINKED ISSUE #1337: fix: Negative or Invalid Work Hours in Attendance Regularization
In `regularize()`, if `checkOut` is before `checkIn`, `workHours` becomes negative and is saved to the database. Fix: Throw an error if `checkOut` is before `checkIn`.

Xenon010101:
fix: reject negative/zero work hours in attendance regularization (#1337)

## Description
Rejects negative and zero work hours in attendance regularization requests.

## Changes
- Added Zod `.refine()` to `regularizeSchema`: ensures checkOut > checkIn and work hours ≤ 24
- Added runtime guard in `regularize()` service method: throws if workHours ≤ 0 or > 24

## Testing
- checkOut before checkIn → validation error
- checkOut equal to checkIn → validation error
- workHours > 24 → validation error
- Normal valid times → proceeds as before

Closes #1337

## Release Notes

* **Bug Fixes**
  * Attendance records now validate that check-out times occur after check-in times
  * Attendance records now enforce a maximum 24-hour work duration limit to prevent invalid submissions

### DIFF
--- a/server/src/module/attendance/attendance.constants.ts
+++ b/server/src/module/attendance/attendance.constants.ts
@@ -1,0 +1,1 @@
+export const MAX_WORK_HOURS = 24;
--- a/server/src/module/attendance/attendance.controller.ts
+++ b/server/src/module/attendance/attendance.controller.ts
@@ -93,8 +93,12 @@ export class AttendanceController {
     } catch (error) {
       if (error instanceof Error) {
         const msg = error.message;
-        if (msg.includes("must be after") || msg.includes("must not exceed") || msg.includes("Cannot regularize") || msg.includes("cannot be in the future"))
-          return res.status(400).json({ message: msg });
+        if (
+          msg === "Cannot regularize attendance for future dates" ||
+          msg === "Check-in time cannot be in the future" ||
+          msg === "Check-out time cannot be in the future" ||
+          msg === "Check-out time must be after check-in time"
+        ) return res.status(400).json({ message: msg });
       }
       console.error(error);
       return res.status(500).json({ message: "Internal Server Error" });
--- a/server/src/module/attendance/attendance.service.ts
+++ b/server/src/module/attendance/attendance.service.ts
@@ -1,5 +1,6 @@
 import { prisma } from "../../database/db.js";
 import type { AttendanceStatus, Prisma } from "@prisma/client";
+import { MAX_WORK_HOURS } from "./attendance.constants.js";
 
 interface AttendanceQuery {
   page: number;
@@ -148,7 +149,7 @@ export class AttendanceService {
     const workHours = (checkOut.getTime() - checkIn.getTime()) / 3600000;
 
     if (workHours <= 0) throw new Error("checkOut must be after checkIn");
-    if (workHours > 24) throw new Error("Work hours must not exceed 24");
+    if (workHours > MAX_WORK_HOURS) throw new Error("Work hours must not exceed 24");
 
     return prisma.attendanceRecord.upsert({
       where: { employeeId_date: { employeeId: data.employeeId, date } },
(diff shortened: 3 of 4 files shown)


Sachinchaurasiya360:
## Code Review — PR #1407: Reject negative/zero work hours in attendance regularization

Hi @Xenon010101, the `MAX_WORK_HOURS` constant is a clean extraction. Found the same fragile controller pattern as PR #1408 and a duplication issue.

---

### 🔴 Fragile error-message string matching in the controller (same pattern as #1408)

```ts
// attendance.controller.ts
if (msg.includes("must be after") || msg.includes("must not exceed") || ...)
  return res.status(400).json({ message: msg });
```

This PR adds `"must be after"` and `"must not exceed"` to the substring list, and both PRs (#1407 and #1408) modify the same controller method. These two PRs will conflict at the controller level.

Beyond the conflict, the substring matching is fragile — if either error message is ever changed, the controller silently returns 500 instead of 400. Use a typed error class instead (same fix as described in #1408).

---

### 🟡 Validation logic duplicated between Zod schema and service layer

Both `regularizeSchema` (via `superRefine`) and `attendanceService.regularize()` check:
- `workHours <= 0` → "checkOut must be after checkIn"
- `workHours > MAX_WORK_HOURS` → "Work hours must not exceed 24"

Since Zod runs first at the request boundary, the service-level checks are dead code in practice. This creates maintenance overhead — both places need to stay in sync when the rules change.

**Recommendation:** Keep validation in Zod (single source of truth at the boundary), remove the duplicate guards from the service layer, and rely on the typed error class approach for any service-layer business logic errors.

---

### 🟡 Merge conflict with PR #1408

Both PRs modify `attendance.controller.ts`, `attendance.service.ts`, and `attendance.validation.ts`. They need to be merged in sequence. Coordinate with the other author and rebase whichever merges second.

Sachinchaurasiya360 (commented):
﻿The Zod superRefine is correct and well-placed. Two issues before merge.

**Issue 1 (Medium) - Fragile substring matching in controller catch block:**

attendance.controller.ts lines 92-95 use msg.includes("must be after"), msg.includes("must not exceed") etc. This couples service error strings to the controller with no type safety - any future typo fix in the error message silently turns a 400 into a 500. The same file already uses exact string equality (===) for the checkOut handler - use that pattern here too.

**Issue 2 (Low) - Duplicate validation logic in service layer:**

The workHours guard (`if (workHours <= 0)` and `if (workHours > MAX_WORK_HOURS)`) exists in both attendance.validation.ts (via superRefine, which runs first) and in attendance.service.ts. Since Zod always runs at the controller boundary before the service is called, the service-layer guards are dead code in normal execution. Remove them from the service. The existing `date > today` guard in the service is a legitimate business rule and should stay - this note only applies to the workHours arithmetic checks.

**Note:** This PR and PR #1408 both modify the regularize catch block in attendance.controller.ts. They will conflict - please rebase onto whichever merges first.

The Zod superRefine logic itself, the MAX_WORK_HOURS constant extraction, and the 24-hour upper bound are all correct. Fix these two issues and it is ready to merge.

Sachinchaurasiya360 (changes requested):
PR #1408 (future date validation for attendance) has now been merged into main. Please rebase this PR onto the latest main to resolve the conflicts on attendance.service.ts and attendance.validation.ts, then push to update.

Sachinchaurasiya360 (changes requested):
Good validation logic for negative/zero work hours — both the Zod `superRefine` and the service-layer check are correct. This PR has merge conflicts with main now. Please rebase onto main and re-request review once CI is green and conflicts are resolved.

Sachinchaurasiya360 (changes requested):
Merge conflict with main (other PRs landed while this was queued). Please rebase on the latest main and push — the code is correct and ready to merge.

PR #1408, opened 2026-06-04
fix: reject future dates in attendance regularization (#1340)

### LINKED ISSUE #1340: fix: Missing Future Date Check in Attendance Regularization
Attendance regularization does not check if the date being regularized is in the future. Fix: Throw an error if the target regularization date is in the future.

Xenon010101:
fix: reject future dates in attendance regularization (#1340)

## Description
Rejects attendance regularization requests where date, checkIn, or checkOut are in the future.

## Changes
- Added Zod `.refine()` to `regularizeSchema`: ensures `date`, `checkIn`, and `checkOut` are not in the future
- Added runtime guard in `regularize()` service method
- Prevents scheduling regularization for dates that haven't occurred yet

## Testing
- Future date → validation error
- Future check-in time → validation error
- Future check-out time → validation error
- Past/present dates → proceeds as before

Closes #1340

## Summary by CodeRabbit

* **Bug Fixes**
  * Attendance regularization now rejects future dates and future check-in/check-out times, preventing future entries.
  * Validation runs earlier to catch timing issues before record persistence.
  * Error handling returns clear 400 responses for timing/validation failures instead of generic server errors.

### DIFF
--- a/server/src/module/attendance/attendance.controller.ts
+++ b/server/src/module/attendance/attendance.controller.ts
@@ -91,6 +91,11 @@ export class AttendanceController {
       const record = await this.attendanceService.regularize(result.data);
       return res.json({ message: "Attendance regularized", record });
     } catch (error) {
+      if (error instanceof Error) {
+        const msg = error.message;
+        if (msg.includes("must be after") || msg.includes("must not exceed") || msg.includes("Cannot regularize") || msg.includes("cannot be in the future"))
+          return res.status(400).json({ message: msg });
+      }
       console.error(error);
       return res.status(500).json({ message: "Internal Server Error" });
     }
--- a/server/src/module/attendance/attendance.service.ts
+++ b/server/src/module/attendance/attendance.service.ts
@@ -138,9 +138,13 @@ export class AttendanceService {
 
     const checkIn = new Date(data.checkIn);
     const checkOut = new Date(data.checkOut);
+    const now = new Date();
+    if (checkIn > now) throw new Error("Check-in time cannot be in the future");
+    if (checkOut > now) throw new Error("Check-out time cannot be in the future");
     if (checkOut <= checkIn) {
       throw new Error("Check-out time must be after check-in time");
     }
+
     const workHours = (checkOut.getTime() - checkIn.getTime()) / 3600000;
 
     return prisma.attendanceRecord.upsert({


Sachinchaurasiya360:
## Code Review — PR #1408: Reject future dates in attendance regularization

Hi @Xenon010101, adding future-date validation is the right fix. Found two issues with the implementation pattern.

---

### 🔴 Fragile error-message string matching in the controller

```ts
// attendance.controller.ts
if (msg.includes("must be after") || msg.includes("must not exceed") || msg.includes("Cannot regularize") || msg.includes("cannot be in the future"))
  return res.status(400).json({ message: msg });
```

This pattern is dangerous. The controller forwards a 400 only if the error message contains one of these hardcoded substrings. Problems:

1. If anyone changes an error message string in the service (even a typo fix), the controller silently returns 500 instead of 400. The bug is invisible until a user reports a white screen.
2. The substring list must be kept in sync with every validation message across the service — currently four strings, but this will grow.

**Fix:** Use a typed error class instead:

```ts
// attendance.errors.ts
export class AttendanceValidationError extends Error {
  constructor(message: string) {
    super(message);
    this.name = "AttendanceValidationError";
  }
}

// attendance.service.ts
if (checkIn > now) throw new AttendanceValidationError("Check-in time cannot be in the future");

// attendance.controller.ts
} catch (error) {
  if (error instanceof AttendanceValidationError) {
    return res.status(400).json({ message: error.message });
  }
  console.error(error);
  return res.status(500).json({ message: "Internal Server Error" });
}
```

---

### 🟡 Future-date validation is duplicated between Zod schema and service layer

The Zod `refine` in `attendance.validation.ts` and the explicit `if` checks in `attendance.service.ts` both validate the same future-date rules. The Zod validation runs first (at the route boundary), so the service-level checks are never reached in practice.

The service-level checks are not wrong to have as a defense-in-depth guard, but they should match the Zod messages exactly (they don't — "Cannot regularize future dates" vs "Cannot regularize future dates" is fine, but "Check-in time cannot be in the future" vs the Zod message needs to be verified). Pick one canonical location and either remove the duplicates or keep them and accept they'll never fire in normal flow.

Sachinchaurasiya360 (commented):
﻿The Zod validation layer is a good addition and the intent is correct. Three issues need to be fixed before merge.

**Bug 1 (Critical) - New service-layer guard is dead code with a divergent error message:**

The PR adds an `if (date > now)` check in attendance.service.ts around line 135, but the pre-existing midnight-comparison guard at line 132 (`if (date > today)`) always fires first for any future date. The new guard is unreachable. Worse, its error message "Cannot regularize future dates" differs from the existing one "Cannot regularize attendance for future dates", so if the guard order ever changes, clients get inconsistent messages. Remove the new redundant guard from the service - Zod already handles this at the boundary, and the service already had its own guard.

**Bug 2 (Critical) - False rejections for IST users:**

The Zod refine in attendance.validation.ts compares checkIn and checkOut against `new Date()` (server UTC wall clock). For users in IST (UTC+5:30), a legitimate checkOut of 11:59 PM IST = 18:29 UTC. If they submit at 11:58 PM IST = 18:28 UTC, the server rejects with "Check-out time cannot be in the future" even though the time is in the past from the user perspective. The validation should compare within the context of the submitted date, not the raw UTC wall clock.

**Bug 3 (Medium) - Phantom substring in controller catch block:**

attendance.controller.ts has `msg.includes("must not exceed")` in the new catch logic, but this exact substring does not appear in any service throw in the codebase. It is a phantom check that never matches. This means if that error path is ever hit it silently becomes a 500. Use exact string equality checks (===) like the pattern already used for the checkOut handler in the same file.

**Note:** This PR and PR #1407 both modify the same regularize catch block in attendance.controller.ts. They will conflict when merged - please coordinate with the #1407 author or rebase onto whichever merges first.

Sachinchaurasiya360 (changes requested):
This PR now has merge conflicts with main after recent attendance-related merges. Please rebase onto main and resolve any conflicts, then push to update this PR.

Sachinchaurasiya360 (changes requested):
Good future-date validation — both the Zod refinements and the service-layer guards are correct. Please note: the controller error-handling block added here overlaps with PR #1407. Please rebase onto main after #1407's changes land (or are merged), ensure there are no duplicate catch-block additions, and re-request review once CI is green.

Sachinchaurasiya360 (changes requested):
Merge conflict with main (other PRs landed while this was queued). Please rebase on the latest main and push — the code is correct and ready to merge.

Issue #997, opened 2026-06-03
bug: window.confirm used for 'Reset all steps' in FirstPRRoadmapPage — breaks design


Sachinchaurasiya360:
## Problem
`FirstPRRoadmapPage.tsx` line ~137 uses `window.confirm('Reset all steps?')` which renders the browser's native confirm dialog — jarring, unstyled, and breaks the design language. This is the reset button in the completion banner.

## Fix
Replace with a custom confirmation using the existing `Button` component:
- Small modal or inline confirmation: 'Are you sure? This will clear your progress.'
- Two buttons: 'Reset' (danger variant) and 'Cancel'
- Use the existing pattern from other confirmation flows in the codebase

## Priority
P2 — design consistency, user experience

Xenon010101:
Hi @Sachinchaurasiya360, I would like to work on this under GSSoC 2026. /assign

soumyasekharshee265-ux:
Hi @Sachinchaurasiya360 👋

I'd love to take this one up. I've reviewed the issue carefully and fully agree with the proposed approach — using the browser's native `window.confirm()` for the "Reset all steps" action in the completion banner breaks the design language and feels jarring against the rest of the polished UI. Replacing it with a custom confirmation modal built on the existing `Button` component is the right fix and will keep the UX consistent across the app.

## 🛠 My Plan to Fix This

### 1. Locate the existing pattern
- Audit the codebase for existing confirmation flows / modal patterns already in use (likely a `Modal`, `Dialog`, or `AlertDialog` component) so the new confirmation matches the project's conventions.
- Reuse the existing `Button` component for consistency.

### 2. Replace `window.confirm` in `FirstPRRoadmapPage.tsx` (~line 137)
- Remove the `window.confirm('Reset all steps?')` call.
- Implement a small, styled confirmation modal/inline confirmation with:
  - **Title/Message:** *"Are you sure? This will clear your progress."*
  - **Primary action:** `Reset` button (using the **danger** variant).
  - **Secondary action:** `Cancel` button (using the default/ghost variant).
- Wire up state (`useState`) to control the modal's open/close behavior.
- Trigger the actual reset logic only when the user confirms via the Reset button.

### 3. Maintain consistency & accessibility
- Ensure the modal supports:
  - **Keyboard navigation** (`Esc` to close, `Tab` between buttons).
  - **Focus trap** while open.
[...]

PR #1409, opened 2026-06-04
Fix/recruiter job delete

YAXH64:
# Pull Request

## Description

Replaced the native browser `window.confirm()` dialog used for job deletion with the application's reusable `ConfirmDialog` component.

### Changes Made

* Removed usage of `window.confirm()` from `RecruiterJobsList`.
* Added a themed confirmation modal before deleting a job.
* Displays the selected job title inside the confirmation dialog.
* Requires explicit user confirmation before performing deletion.
* Added loading state during deletion requests.
* Prevented duplicate deletion submissions while a request is in progress.
* Improved error handling by keeping the dialog open when deletion fails, allowing recruiters to retry.
* Preserved existing API calls, toast notifications, and local state updates.

### Benefits

* Consistent UI/UX across the application.
* Improved accessibility and keyboard navigation.
* Better support for dark/light themes.
* Eliminates reliance on browser-native dialogs.
* Safer deletion workflow with explicit confirmation.

## Related Issue

Fixes #1138

## Type of Change

* [x] Bug Fix
* [ ] Feature
* [x] Enhancement
* [ ] Documentation

### Manual Testing

1. Open Recruiter Dashboard → My Job Listings.
2. Click the Delete button for a job.
3. Verify the confirmation modal appears.
4. Click Cancel and verify the job is not deleted.
5. Reopen the modal and click Delete.
6. Verify the job is removed successfully.
7. Verify the success toast appears.
8. Simulate a failed delete request and verify:

   * Error toast appears.
   * Dialog remains open.
   * User can retry deletion.
9. Verify multiple rapid clicks do not trigger duplicate delete requests.
10. Verify Escape key closes the dialog when not loading.
11. Verify dark mode and light mode styling.

### Validation

* No TypeScript errors.
* No build errors.
* Existing delete functionality remains intact.
* Existing toast notifications continue to work.

### Before

* Clicking Delete opened a browser-native confirmation dialog (`window.confirm()`).

### After

* Clicking Delete opens a themed confirmation modal integrated with the application's design system.

> Attach screenshots or a short screen recording showing:
>
> * Delete modal open
> * Cancel flow
> * Successful delete flow

## Checklist

* [x] Code follows project guidelines
* [x] No new compile/type errors
* [x] Tested manually (include steps above)
* [x] No `.env`, credentials, or `node_modules` committed
* [ ] Docs updated (not required)
* [ ] Screenshot or video attached for every UI change

## Summary by CodeRabbit

* **New Features**
[...]

PR #1406, opened 2026-06-04
fix(recruiter): restore applicant list when clearing search

YAXH64:
# Pull Request

## Description

Fixes an issue where clearing the applicant search field in the Recruiter Applications page would leave the UI in an empty state instead of restoring the full applicant list.

### Root Cause

The component maintained pagination data in local React state while application data was managed by React Query. When a recruiter cleared the search field, React Query served cached results for the original query key, but the query function did not execute again. Because pagination state was updated inside the query function, the pagination state became out of sync with the cached application data, causing the page to incorrectly display an empty state.

### Changes Made

* Removed local pagination state management.
* Returned both `applications` and `pagination` from the React Query query function.
* Derived pagination directly from React Query data to keep state synchronized.
* Removed the side-effect (`setPagination`) from the query function.
* Trimmed search input before adding it to query parameters to prevent whitespace-only searches.
* Preserved existing search, filtering, pagination, and caching behavior.

### Benefits

* Restores the full applicant list when search is cleared.
* Eliminates React Query state synchronization issues.
* Removes an anti-pattern (calling state setters inside `queryFn`).
* Improves maintainability by keeping applications and pagination under a single source of truth.

## Related Issue

Fixes #1142

## Type of Change

* [x] Bug Fix
* [ ] Feature
* [x] Enhancement
* [ ] Documentation

### Manual Testing

1. Open Recruiter → Applications page.
2. Verify all applicants load correctly.
3. Search for an applicant name.
4. Verify results are filtered correctly.
5. Search for a non-existent applicant and verify the empty state appears.
6. Clear the search field.
7. Verify the full applicant list is restored.
8. Repeat search → clear → search multiple times.
9. Verify pagination continues to work correctly.
10. Verify status filters continue to work after clearing search.

### Validation

* No TypeScript errors.
* No build errors.
* Existing search functionality remains unchanged.
* Existing pagination and filtering behavior remains intact.

## Screenshots / Video

*No UI changes in this PR*

## Checklist

* [x] Code follows project guidelines
* [x] No new compile/type errors
* [x] Tested manually (include steps above)
* [x] No `.env`, credentials, or `node_modules` committed
* [ ] Docs updated (not required)
* [x] No UI changes in this PR

## Summary by CodeRabbit

* **New Features**
  * Confirm dialogs now support flexible content display and loading states with visual spinner feedback
  * Application advance action now requires explicit user confirmation before processing

* **Bug Fixes**
  * Search parameters are now trimmed to prevent issues with leading/trailing whitespace

PR #1405, opened 2026-06-04
feat(recruiter): add advance candidate confirmation dialog

YAXH64:
# Pull Request

## Description

Adds a confirmation step before recruiters can advance candidates to the next hiring stage.

Previously, clicking the **Advance** button immediately triggered the advancement action, which could lead to accidental candidate progression and incorrect hiring pipeline states. This PR introduces a confirmation dialog that requires explicit user confirmation before the candidate is advanced.

### Changes Made

* Added a confirmation modal before candidate advancement.
* Introduced `pendingAdvanceApp` state to manage the selected candidate awaiting confirmation.
* Refactored the recruiter applications workflow so advancement only occurs after explicit confirmation.
* Enhanced the reusable `ConfirmDialog` component:

  * Added loading state support.
  * Added configurable confirm button variants.
  * Added keyboard accessibility (Escape key support).
  * Added support for custom dialog content through `children`.
* Prevented duplicate submissions through UI and handler-level guards.
* Improved error handling by keeping the dialog open when advancement fails, allowing recruiters to retry.

### User Impact

Recruiters can no longer accidentally advance candidates with a single misclick. Candidate progression now requires an explicit confirmation step, improving data integrity and hiring workflow reliability.

## Related Issue

Fixes #1134

## Type of Change

* [x] Bug Fix
* [ ] Feature
* [x] Enhancement
* [ ] Documentation

### Manual Testing

1. Open Recruiter Dashboard.
2. Navigate to a job's Applications page.
3. Click **Advance** on any candidate.
4. Verify a confirmation dialog appears.
5. Click **Cancel** and confirm no status change occurs.
6. Reopen the dialog and click **Confirm Advance**.
7. Verify the candidate advances successfully.
8. Verify success toast appears.
9. Simulate an API failure and verify:

   * Error toast appears.
   * Dialog remains open.
   * User can retry.
10. Rapidly click Confirm and verify only a single request is sent.
11. Press Escape while the dialog is open and verify it closes correctly (when not loading).

### Before

* Clicking **Advance** immediately moved the candidate to the next round.

### After

* Clicking **Advance** opens a confirmation dialog.
* Candidate advances only after explicit confirmation.

> Attach screenshots or a short screen recording demonstrating:
>
> * Opening the confirmation dialog
> * Cancel flow
> * Successful advance flow

## Checklist

* [x] Code follows project guidelines
* [x] No new compile/type errors
* [x] Tested manually (include steps above)
* [x] No `.env`, credentials, or `node_modules` committed
[...]

PR #1410, opened 2026-06-04
fix: replace window.confirm with ConfirmDialog component (#997)

Xenon010101:
## Summary by CodeRabbit

* **Refactor**
  * Updated the confirmation experience when resetting roadmap progress with an improved dialog interface.

### DIFF
--- a/client/src/module/student/opensource/FirstPRRoadmapPage.tsx
+++ b/client/src/module/student/opensource/FirstPRRoadmapPage.tsx
@@ -1,4 +1,5 @@
 import { useState, useCallback } from "react";
+import { ConfirmDialog } from "../../../components/ui/ConfirmDialog";
 import { motion, AnimatePresence } from "framer-motion";
 import {
   CheckCircle2, GitPullRequest, ArrowRight,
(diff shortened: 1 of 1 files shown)


Sachinchaurasiya360:
## Code Review — PR #1410: Replace window.confirm with ConfirmDialog

Hi @Xenon010101, the ConfirmDialog enhancements (Escape key, loading state, confirmVariant, children) are well implemented. One coordination issue to flag.

---

### 🟡 Merge conflict: PRs #1405, #1406, #1409, and this PR all apply identical changes to `ConfirmDialog.tsx`

All four PRs modify `ConfirmDialog.tsx` with the exact same diff (Escape key listener, `loading` prop, `confirmVariant` prop, `children` support, `description` made optional). Whichever merges second, third, and fourth will have conflicts on this file.

The cleanest resolution: merge this PR (or whichever improves the base component) first, then the other three PRs only need to rebase and drop the `ConfirmDialog.tsx` changes from their diff since the base already has them.

Please coordinate with @YAXH64 on PRs #1405, #1406, and #1409 before merging.

---

### ✅ Everything else looks good

- Escape key handler is properly cleaned up via `removeEventListener` in the effect return
- `loading ? undefined : onCancel` on the backdrop click correctly prevents dismiss while loading
- Focus on `cancelRef` is correct — default focus should be on the safe/cancel action
- `role="alertdialog"` with `aria-modal` and `aria-labelledby`/`aria-describedby` is proper accessibility

PR #1419, opened 2026-06-04
fix: do not issue JWT on registration before email verification (#1326)

### LINKED ISSUE #1326: [Security] JWT returned for unverified users bypasses email verification
## Description
The entire backend API issues JWT tokens to **any** user immediately after signup, regardless of email verification status. There is no `isVerified` or `email_verified_at` check on protected routes. This means an attacker with a temporary/disposable email can create accounts, obtain valid JWTs, and access all authenticated endpoints without ever verifying their email.

## Location
- `backend/routes/auth.js` — signup route (lines 42–51)
- `backend/middleware/auth.js` — JWT middleware (all lines)

## The Bug
**auth.js (signup):**
```js
const token = jwt.sign({ id: user.id }, process.env.JWT_SECRET);
res.json({ token, user });  // <-- issued before any verification
```

**auth.js middleware:**
```js
const decoded = jwt.verify(token, process.env.JWT_SECRET);
req.user = decoded;
next();  // <-- no check for user.verified
```

## Exploitation
1. Register with `tempmail123@example.com`
2. Receive a valid JWT immediately (in registration response)
3. Call protected endpoints (`/api/exams`, `/api/submissions`, `/api/results`)
4. The server processes requests normally — no email verification required

## Suggested Fix
1. Add an `email_verified_at` or `is_verified` column to the users table
2. In the JWT middleware, query the user and reject if not verified
3. Optionally: issue a limited-scope "pre-verification" token that only allows calling `/auth/verify-email`

Xenon010101:
### Description

**Issue:** Closes #1326

The registration endpoint was returning a full-access JWT token immediately upon signup, before the user verified their email via OTP. This meant an unverified user could access all protected routes for 7 days.

### Fix

Simply removed the JWT generation from the registration flow. Users must now verify their email first, then log in normally (login already blocks unverified users with \EMAIL_NOT_VERIFIED\).

### Changes

- \uth.service.ts\: removed \generateToken()\ call on registration
- \uth.controller.ts\: removed \setTokenCookie()\ call on registration response

### Security

Login's existing \isVerified\ check (line 334) already blocks unverified users — this fix closes the bypass at the source.

## Summary by CodeRabbit

* **Bug Fixes**
  * Registration no longer issues an authentication cookie or token automatically; users will receive the OTP and a user record on successful signup but will not be signed in automatically.


Sachinchaurasiya360:
## Code Review — PR #1419: No JWT on registration before email verification

Hi @Xenon010101, the security intent here is correct — issuing a JWT before the email is verified allows unverified accounts to access protected routes. Found one breaking change that needs a client-side fix.

---

### 🔴 Client is not updated — registration will silently break the login flow

The server now returns `{ user }` without a token:

```ts
// auth.service.ts
return { user };   // token removed
```

```ts
// auth.controller.ts
// setTokenCookie(res, data.token);  ← removed
return res.status(201).json({ message: "Registration successful", ...data });
```

But the client-side registration handler almost certainly calls the API, receives the response, and then attempts to extract `token` from it to store in auth state (Zustand store or cookie). Without a client-side change, one of two things happens:

1. **If the client tries to read `data.token`:** it gets `undefined`, the Zustand auth store is populated with `token: undefined`, and any subsequent authenticated request fails with 401.
2. **If the client calls `setTokenCookie` or sets auth state with the token:** it will pass `undefined`, which may corrupt the auth store silently.

**Fix required:** The client registration handler needs to be updated to:
- NOT attempt to log in after registration
- Instead show the "check your email" / OTP verification screen immediately
- Only store the token after the OTP verification step succeeds

Please include the client-side changes in this PR, otherwise the registration flow will be broken in production.

---

### 🟡 Verify: Google OAuth path still issues a token at login (not registration)

The `googleAuth` flow creates a user and returns a token in the same call. If Google auth is considered "pre-verified" (since Google validates the email), the no-token-on-registration rule may not apply there. Confirm whether the Google auth path is intentionally left unchanged.

Sachinchaurasiya360 (commented):
﻿The core fix is mechanically correct - removing the token from the registration response and relying on the verify-email flow to issue the JWT is the right approach. But two issues need fixing before merge.

**Issue 1 (Medium) - Ghost-auth else branch in both client registration pages:**

client/src/module/auth/RegisterPage.tsx lines 286-289 and client/src/module/recruiter/auth/RecruiterRegisterPage.tsx lines 100-103 both have an else branch that calls login(data.user) when isVerified is true. After this PR removes the auth cookie from registration, hitting this branch would put the user in a false-authenticated state: localStorage shows logged in but every protected API call immediately 401s because no httpOnly cookie exists. The branch is currently unreachable since register() never sets isVerified=true, but this PR widens the fragility gap. Remove the else branch and unconditionally redirect to /verify-email regardless of isVerified.

**Issue 2 (Medium) - verifyEmail still leaks the JWT in the JSON response body:**

server/src/module/auth/auth.controller.ts line 196 returns token: data.token in the response body in addition to setting it as an httpOnly cookie on line 195. The client never reads the body token (VerifyEmailPage.tsx line 43 only reads data.user). Returning the token in the body means XSS can steal it - exactly the vulnerability this PR addresses for registration. Remove token: data.token from the verifyEmail JSON response while you are here.

The redirect-to-verify-email UX flow is correct and complete. Fix these two and the PR is ready to merge.

Xenon010101:
Closing in favor of #1444 which has the same fix rebased onto latest main.

PR #1421, opened 2026-06-04
Feat/semantic caching

### LINKED ISSUE #856: [FEATURE]: Implement Semantic Caching for AI-Generated Roadmaps using Qdrant
### Feature Summary

Semantic caching layer for AI roadmap generation using Qdrant - checks for semantically similar past requests before querying Gemini. It returns cached results immediately rather than using up API quota.

### Problem Statement

While going through the codebase, I noticed that the function `generateAiRoadmap` in `roadmap.ai.service.ts` makes a new call to `GeminiProvider("gemini-2.5-flash-lite")` for every request without any caching. Prompts that are semantically identical, like "MERN Stack Beginner Roadmap" and "Beginner roadmap for MERN," are sent to Gemini twice, returning almost the same results.

I also found that `AtsService.scoreResume` in `ats.service.ts` has a 24-hour exact-match cache using `prisma.atsScore.findFirst`, which indicates that the issue of duplicate requests is already acknowledged. However, it only captures inputs that are exactly the same. Semantically similar prompts still contact the API each time.

As user volume increases, this leads to:

- Unnecessary Gemini API costs and rate limit pressure
- Students waiting several seconds for roadmaps that were effectively generated already

### Proposed Solution

Introduce a semantic caching layer inside `generateAiRoadmap` that runs before the Gemini call:

- Normalize and embed the input parameters using the Gemini Embeddings API, which is already in the codebase. No new credentials are needed.
- Query Qdrant for a cosine similarity match above a 0.95 threshold.
- If there's a cache hit, return the stored JSON instantly without making a Gemini call.
- If there's a cache miss, call Gemini as usual and store the result and embedding in Qdrant for future requests.

DevOps safety, with no risk to existing contributors:  
The new feature will be controlled by a `SEMANTIC_CACHE_ENABLED=true` environment variable. If it is missing or set to false, the code will revert to the current behavior without any changes. Contributors without Qdrant set up will see no difference at all.  
New `.env.example` entries (for documentation only):  
- `SEMANTIC_CACHE_ENABLED=false`  
- `QDRANT_URL=`  
- `QDRANT_API_KEY=`  
- `QDRANT_COLLECTION_NAME=ai_prompt_cache`  
The Qdrant Cloud free tier (1GB, no credit card required) is enough for tens of thousands of cached roadmaps at the project's current scale.  

Files to change:  
- `server/src/lib/semantic-cache.ts` : new file for Qdrant client, embedding, similarity search, and storage  
- `server/src/module/roadmap/roadmap.ai.service.ts` : wrap `generateAiRoadmap` with a cache check  
- `.env.example`:  add 4 new optional variables  

### Alternatives Considered

pgvector (Postgres extension), use the existing Neon database with vector support. This option eliminates the need for a new service but requires the pgvector extension to be enabled on the hosted database, which requires action from a maintainer.  
Hash-based exact caching, normalize and hash prompts, store results in a new Prisma table. This solution is simpler but does not account for semantically similar inputs, which is the same limitation as the current ATS cache.  

I believe Qdrant is the best choice here because it does not affect any existing infrastructure, and the free tier comfortably supports this project's scale.

### Additional Context

I actually built this exact architecture in MergeMind (one of my projects) and used Qdrant to store & retrieve past PR reviews by semantic similarity, keeping LLM feedback consistent across similar code changes. Same pattern here, for caching prompts.
I'm also familiar with the roadmap module from PR #623, I directly worked on roadmap.ai.service.ts and its neighboring infrastructure. Happy to start right away if this is approved.

ChandanMeher4:
## Description

Adds a semantic caching layer for AI roadmap generation. Before hitting Gemini, the input is embedded using `gemini-embedding-001` and checked against a Qdrant vector store for a cosine similarity match above 0.95. On a cache hit, the stored roadmap JSON is returned instantly without any Gemini call. On a miss, Gemini is called as usual and the result is stored for future requests.

The embedding from the similarity search is reused for storing on a cache miss — so there's only ever one embedding call per request regardless of hit or miss.

The entire feature is gated behind `SEMANTIC_CACHE_ENABLED=true`. When disabled or missing, the code falls through to existing behavior with zero impact on other contributors.

## Related Issue
Fixes #856

## Type of Change
- [ ] Bug Fix
- [x] Feature
- [ ] Enhancement
- [ ] Documentation

## Testing

1. **Flag disabled (default):** Start server without `SEMANTIC_CACHE_ENABLED` — logs `Semantic cache disabled`, roadmap generation works identically to before ✅
2. **Flag enabled, no URL:** Set `SEMANTIC_CACHE_ENABLED=true` in `server/.env`, leave `QDRANT_URL` empty — logs `QDRANT_URL is not set — cache will be skipped`, server starts cleanly and roadmap generation still works ✅
3. **Full end-to-end:** Requires Qdrant Cloud credentials. Set `SEMANTIC_CACHE_ENABLED=true`, `QDRANT_URL`, and `QDRANT_API_KEY` in `server/.env`. Generate a roadmap — first call logs cache miss and hits Gemini. Generate semantically similar roadmap — second call logs cache hit and returns instantly.

For local development, add the following to `server/.env`:
```
SEMANTIC_CACHE_ENABLED=true
QDRANT_URL=<your-qdrant-cloud-url>
QDRANT_API_KEY=<your-qdrant-api-key>
QDRANT_COLLECTION_NAME=ai_prompt_cache
```

## Screenshots / Video
_No UI changes in this PR_

## Checklist
- [x] **Screenshot or video attached for every UI change**

## Summary by CodeRabbit

* **New Features**
  * Semantic caching for AI-generated roadmaps to return cached results faster when available.

* **Chores**
  * Added optional semantic-cache configuration and vector-database client support.
  * Centralized embedding generation and integrated it with the cache; introduced cache TTL and similarity threshold defaults.

* **Bug Fixes**
  * Removed stray leading character from environment template header to prevent parsing issues.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,4 +1,4 @@
-﻿# InternHack - root environment template (Docker Compose + reference for local dev)
+# InternHack - root environment template (Docker Compose + reference for local dev)
 #
 # Quick start:
 #   cp .env.example .env
@@ -66,3 +66,9 @@ VITE_DODO_MODE=test_mode
 # LeetCode import client flag — set to false to disable (default: enabled)
 VITE_LEETCODE_IMPORT_ENABLED=true
 LOG_LEVEL=info
+
+# ── Semantic Cache (optional — Qdrant Cloud free tier) ──────────────────
+SEMANTIC_CACHE_ENABLED=false
+QDRANT_URL=
+QDRANT_API_KEY=
+QDRANT_COLLECTION_NAME=ai_prompt_cache
--- a/.env.example
+++ b/.env.example
@@ -66,9 +66,3 @@ VITE_DODO_MODE=test_mode
 # LeetCode import client flag — set to false to disable (default: enabled)
 VITE_LEETCODE_IMPORT_ENABLED=true
 LOG_LEVEL=info
-
-# ── Semantic Cache (optional — Qdrant Cloud free tier) ──────────────────
-SEMANTIC_CACHE_ENABLED=false
-QDRANT_URL=
-QDRANT_API_KEY=
-QDRANT_COLLECTION_NAME=ai_prompt_cache
--- a/server/.env.example
+++ b/server/.env.example
@@ -66,3 +66,9 @@ VITE_DODO_MODE=test_mode
 # LeetCode import client flag — set to false to disable (default: enabled)
 VITE_LEETCODE_IMPORT_ENABLED=true
 LOG_LEVEL=info
+
+# ── Semantic Cache (optional — Qdrant Cloud free tier) ──────────────────
+SEMANTIC_CACHE_ENABLED=false
+QDRANT_URL=
+QDRANT_API_KEY=
+QDRANT_COLLECTION_NAME=ai_prompt_cache
--- a/server/.env.example
+++ b/server/.env.example
@@ -72,3 +72,6 @@ SEMANTIC_CACHE_ENABLED=false
 QDRANT_URL=
 QDRANT_API_KEY=
 QDRANT_COLLECTION_NAME=ai_prompt_cache
+
+# Optional: cache TTL in days (default: 30)
+SEMANTIC_CACHE_TTL_DAYS=30
--- a/server/package-lock.json
+++ b/server/package-lock.json
@@ -18,6 +18,7 @@
         "@prisma/adapter-pg": "^7.8.0",
         "@prisma/client": "^7.8.0",
         "@prisma/extension-accelerate": "^3.0.1",
+        "@qdrant/js-client-rest": "^1.18.0",
         "@types/nodemailer": "^8.0.0",
         "@types/pdfkit": "^0.17.6",
         "bcryptjs": "^3.0.3",
@@ -2154,6 +2155,42 @@
         "@types/react": "^18.0.0 || ^19.0.0",
         "react": "^18.0.0 || ^19.0.0",
         "react-dom": "^18.0.0 || ^19.0.0"
+      }
+    },
+    "node_modules/@qdrant/js-client-rest": {
+      "version": "1.18.0",
+      "resolved": "https://registry.npmjs.org/@qdrant/js-client-rest/-/js-client-rest-1.18.0.tgz",
+      "integrity": "sha512-/0dqX5uV9chC1DnYSnU4gNMrDqse/pt6hHg3Rqqpl5isH7xl1xSNvffjzBoxycDD79luWn7Ho6Rh/61sOs5DNw==",
+      "license": "Apache-2.0",
+      "dependencies": {
+        "@qdrant/openapi-typescript-fetch": "1.2.6",
+        "undici": "^6.24.0"
+      },
+      "engines": {
+        "node": ">=18.17.0",
+        "pnpm": ">=8"
+      },
+      "peerDependencies": {
+        "typescript": ">=4.7"
+      }
+    },
+    "node_modules/@qdrant/js-client-rest/node_modules/undici": {
+      "version": "6.26.0",
+      "resolved": "https://registry.npmjs.org/undici/-/undici-6.26.0.tgz",
+      "integrity": "sha512-4yqz8a3n5HmGTlsbADNtr/dJlhkh/55Rq798G6ibiULcXbDtaLpTl1pvdqcbFfeoj3iSi52lePFM7h9H21cw/A==",
+      "license": "MIT",
+      "engines": {
+        "node": ">=18.17"
+      }
+    },
+    "node_modules/@qdrant/openapi-typescript-fetch": {
+      "version": "1.2.6",
+      "resolved": "https://registry.npmjs.org/@qdrant/openapi-typescript-fetch/-/openapi-typescript-fetch-1.2.6.tgz",
+      "integrity": "sha512-oQG/FejNpItrxRHoyctYvT3rwGZOnK4jr3JdppO/c78ktDvkWiPXPHNsrDf33K9sZdRb6PR7gi4noIapu5q4HA==",
+      "license": "MIT",
+      "engines": {
+        "node": ">=18.0.0",
+        "pnpm": ">=8"
       }
     },
     "node_modules/@radix-ui/primitive": {
@@ -8706,7 +8743,6 @@
       "version": "5.9.3",
       "resolved": "https://registry.npmjs.org/typescript/-/typescript-5.9.3.tgz",
       "integrity": "sha512-jl1vZzPDinLr9eUt3J/t7V6FgNEw9QjvBPdysz9KfQDD41fQrC2Y4vKQdiaUpFT4bXlb1RHhLpp8wtm6M5TgSw==",
-      "devOptional": true,
       "license": "Apache-2.0",
       "bin": {
         "tsc": "bin/tsc",
--- a/server/package.json
+++ b/server/package.json
@@ -30,6 +30,7 @@
     "@prisma/adapter-pg": "^7.8.0",
     "@prisma/client": "^7.8.0",
     "@prisma/extension-accelerate": "^3.0.1",
+    "@qdrant/js-client-rest": "^1.18.0",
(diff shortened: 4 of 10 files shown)


Sachinchaurasiya360:
## Code Review — PR #1421: Semantic Caching (Qdrant)

Hi @ChandanMeher4, the overall design is solid — feature-flag gated, fail-open on errors, non-blocking store. Found three issues that need fixing.

---

### 🔴 `VECTOR_SIZE = 768` is hardcoded and will silently break if the embedding model changes

```ts
const VECTOR_SIZE = 768;
```

The 768 comes from the embedding dimension of the model currently powering `generateEmbedding()`. If the model is ever swapped (e.g. to a 1536-dim or 3072-dim model), the Qdrant collection will have been created with `size: 768`. Any new embedding inserted will be silently rejected by Qdrant with a dimension mismatch error.

**Fix:** Export the vector size from the embedding utility so both the cache init and the model are always in sync:

```ts
// job-index.enrichment.ts
export const EMBEDDING_VECTOR_SIZE = 768;

// semantic-cache.ts
import { generateEmbedding, EMBEDDING_VECTOR_SIZE } from "../module/job-index/job-index.enrichment.js";
const VECTOR_SIZE = EMBEDDING_VECTOR_SIZE;
```

---

### 🔴 `initSemanticCache()` runs on every process — no `CRON_PROCESS` guard

```ts
// server/src/index.ts
await initSemanticCache().catch((err) => logger.error("Failed to init semantic cache:", err));
```

Every other expensive singleton init (cron jobs, AI providers) in this file is guarded by `process.env.CRON_PROCESS`. `initSemanticCache` creates a Qdrant client and opens a persistent connection. In a multi-process deployment (PM2 cluster, multiple Heroku dynos), this runs on every process, creating N connections to Qdrant where N is the process count.

Unlike a cron guard, this doesn't need to be limited to one process — all processes are allowed to use the cache. But the collection creation check (`client.getCollections()` + `createCollection()`) should be idempotent and is already guarded by `if (!exists)`, so this is low risk. The main concern is wasted connections. Consider lazy initialization (init on first use) rather than eager init at startup.

---

### 🟡 Module-level mutable `client` variable has no reconnect strategy

```ts
let client: QdrantClient | null = null;
```

If `initSemanticCache` fails (e.g. Qdrant is temporarily unreachable at startup), `client` stays `null` for the entire process lifetime. All cache lookups and stores are silently skipped forever — `isSemanticCacheEnabled()` returns `true` but every `searchAndEmbed` / `storeInCache` call short-circuits on `!client`. There's no mechanism to retry initialization after the service recovers.

This is acceptable for a v1 (fail-open is correct), but document it clearly so future maintainers don't wonder why the cache is enabled but never gets a hit after a flaky startup.

Sachinchaurasiya360 (commented):
﻿The feature-flag design and graceful-fallback when Qdrant is unavailable are solid. A few issues to fix before merge.

**Issue 1 (High) - forceCreate bypasses the cache check incorrectly:**

roadmap.controller.ts checks input.forceCreate to skip a duplicate-roadmap guard, but roadmap.ai.service.ts does the semantic cache lookup unconditionally before that guard is ever reached. A user who explicitly requests a fresh generation with forceCreate=true will still get a stale cached result back. Fix: gate the cache lookup behind the forceCreate flag.

```ts
if (isSemanticCacheEnabled() && !input.forceCreate) {
  // cache lookup
}
```

**Issue 2 (Medium) - Embedding call fires even when Qdrant client is null:**

searchAndEmbed() in semantic-cache.ts always calls generateEmbedding(text) even when client is null (Qdrant unavailable). When Qdrant is down, every single roadmap generation wastes a Gemini embedding API call just to discard the result. Add an early return when client is null:

```ts
if (!client) return { result: null, embedding: [] };
```

**Issue 3 (Medium) - Cross-module import from lib into module directory:**

semantic-cache.ts imports generateEmbedding from server/src/module/job-index/job-index.enrichment.ts. A lib/ file must not import from a module/ directory. generateEmbedding is a generic utility - extract it to server/src/lib/embedding.ts and have both files import from there.

**Issue 4 (Medium) - Config constants inline in lib file:**

VECTOR_SIZE, SIMILARITY_THRESHOLD, and QDRANT_COLLECTION_NAME are hardcoded in semantic-cache.ts. Per project convention these belong in server/src/config/. Create server/src/config/semantic-cache.config.ts to house them, mirroring the pattern of cache-config.ts and redis.ts.

**Non-blocking but worth noting:**

- No TTL or eviction policy on cached entries - roadmaps will be cached indefinitely even as the AI model improves. Consider filtering by age in the search payload.
- Cache hits log a fake AIRequest row with zeroed latency and tokens - either skip logging on cache hits or log to a separate metric.
- When QDRANT_URL is set but QDRANT_API_KEY is missing, the failure is a confusing silent 401. Log a specific warning in that case.

Fix the four issues above and this is ready to merge.

ChandanMeher4:
@Sachinchaurasiya360 addressed everything from both review rounds:

**Your 4 fixes:**
- Cache lookup now behind `!input.forceCreate` 
- Early return added before embedding call when `client` is null 
- `generateEmbedding` moved to `server/src/lib/embedding.ts`, re-exported from `job-index.enrichment.ts` so existing imports don't break 
- Config constants moved to `server/src/config/semantic-cache.config.ts` 

**The 3 non-blocking notes:**
- Removed the fake `logAIRequest` on cache hits 
- Added a specific warning when `QDRANT_API_KEY` is missing 
- Added TTL filtering on the Qdrant search query using `createdAt` - defaults to 30 days, can be overridden with `SEMANTIC_CACHE_TTL_DAYS` 

**CodeRabbit:**
- `generateEmbedding` now wrapped in its own try/catch - embedding failures return a cache miss instead of throwing 
- `GEMINI_API_KEY` validated upfront in `embedding.ts` 
- Blank `QDRANT_COLLECTION_NAME` now falls back to the default 

All 10 checks passing, no conflicts.

Sachinchaurasiya360 (changes requested):
Test Server fails and there is a critical startup bug.

**1. Critical: module-level throw in embedding.ts crashes server regardless of feature flag**

```typescript
// server/src/lib/embedding.ts
const apiKey = process.env["GEMINI_API_KEY"];
if (!apiKey) {
  throw new Error("GEMINI_API_KEY is not set — embedding generation will fail");  // <-- module-level
}
```

This runs at import time. Since `semantic-cache.ts` imports `embedding.ts` unconditionally, and `index.ts` imports `semantic-cache.ts` unconditionally, any environment without `GEMINI_API_KEY` set will crash on startup — even with `SEMANTIC_CACHE_ENABLED=false`. The error must be moved inside the `generateEmbedding` function body (or made lazy), so it only fires when actually called.

**2. Test Server fails: `prisma.$transaction is not a function`**

```
FAIL src/module/recruiter/recruiter.service.test.ts
TypeError: prisma.$transaction is not a function
```

Investigate whether the test mock for `prisma` needs to include `$transaction`. If this is a pre-existing failure on your branch's base, rebase on current main and confirm tests pass.

**3. Minor: `buildCacheKey` spread on potentially non-iterable types**

`[...input.preferredDays]` — verify `AiGenerateInput.preferredDays` is typed as an array or iterable, not a union type that could be a string.

ChandanMeher4:
@Sachinchaurasiya360 addressed all three -

- Moved the `GEMINI_API_KEY` check inside `generateEmbedding()` using a `lazy getGenAI()` helper, so it only throws when the function is actually called. Server starts fine without the key set.

- All 11 checks are green, including Test Server. The `prisma.$transaction` failure in recruiter.service.test.ts isn't related to this PR - looks like it was a pre-existing issue on main that's since been fixed.

- Checked the Zod schema - preferredDays is `z.array(DAY).min(1).max(7)`, always a validated array so the spread is fine.

Also switched the embedding model from `gemini-embedding-001` to `gemini-embedding-2` since that's Google's current recommended model. `EMBEDDING_VECTOR_SIZE` stays at 768, which is a supported output dimension for it.

PR #1663, opened 2026-06-06
fix: add Zod validation for executeCode endpoint in DSA controller Closes #1639

### LINKED ISSUE #1639: bug: DSA executeCode endpoint reads req.body directly instead of using existing executeCodeSchema
## Description
The `executeCode()` endpoint in `dsa.controller.ts` destructures `language` and `code` directly from `req.body` without any Zod validation. The `executeCodeSchema` already exists in `dsa.validation.ts` but is never imported or used in the controller.

## Existing Schema (dsa.validation.ts)
```typescript
export const executeCodeSchema = z.object({
  language: z.enum(["python", "cpp", "java"]),
  code: z.string().min(1, "Code is required").max(50000, "Code too long"),
});
```

## Current Code (dsa.controller.ts:195-207)
```typescript
async executeCode(req: Request, res: Response, next: NextFunction) {
  const userId = req.user?.id;
  if (!userId) { res.status(401).json({ message: "Authentication required" }); return; }
  const problemId = parseInt(req.params.problemId as string);
  if (isNaN(problemId)) { res.status(400).json({ message: "Invalid problem ID" }); return; }
  const { language, code } = req.body;  // No Zod validation
  const result = await this.dsaService.executeCodeAgainstTestCases(userId, problemId, language, code);
  res.json(result);
}
```

## Expected Behavior
Request body should be validated via `executeCodeSchema.safeParse(req.body)` before accessing `language` and `code`.

## Root Cause
Schema was defined but never wired to the controller.

## Proposed Fix
Import `executeCodeSchema` and add:
```typescript
const body = executeCodeSchema.safeParse(req.body);
if (!body.success) return res.status(400).json({ message: "Validation failed", errors: body.error.flatten() });
```
Then use `body.data.language` and `body.data.code` instead of raw `req.body`.

## Severity
**Tier B** — Missing type validation on user-supplied code that gets executed against test cases.

## File & Line
`server/src/module/dsa/dsa.controller.ts:201`

## Estimated Fix Time
~2 minutes

Akshita-2307:
## What
Adds Zod validation for the executeCode endpoint in DSA controller

## Root cause
The `executeCode` handler destructured `{ language, code }` directly from `req.body` without Zod validation, allowing unvalidated input to reach the service layer.

## Changes
- `server/src/module/dsa/dsa.controller.ts` — imported `executeCodeSchema`, added `safeParse` guard before destructuring body data

## Verification
- [ ] Bug reproduced before fix
- [ ] Fix verified locally
- [ ] git diff upstream/main...HEAD --stat shows only intended file
- [ ] No console.logs or debug logs remaining
- [ ] Build passes

## Before / After
Backend only — no UI changes

Closes #1639

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for code execution submissions to reliably detect malformed requests before processing.
  * Invalid submissions now return a clear 400 response with detailed, user-friendly error information to help identify and fix submission issues quickly.

### DIFF
--- a/server/src/module/dsa/dsa.controller.ts
+++ b/server/src/module/dsa/dsa.controller.ts
@@ -4,6 +4,7 @@ import { parsePagination } from "../../utils/pagination.utils.js";
 import { syncLeetCodeSolvedProblems } from "./leetcode.service.js";
 import { prisma } from "../../database/db.js";
 import { isPremiumUser } from "../../utils/premium.utils.js";
+import { executeCodeSchema } from "./dsa.validation.js";
 
 export class DsaController {
   constructor(private dsaService: DsaService) {}
@@ -203,7 +204,9 @@ export class DsaController {
       if (!userId) { res.status(401).json({ message: "Authentication required" }); return; }
       const problemId = parseInt(req.params.problemId as string);
       if (isNaN(problemId)) { res.status(400).json({ message: "Invalid problem ID" }); return; }
-      const { language, code } = req.body;
+      const body = executeCodeSchema.safeParse(req.body);
+      if (!body.success) return res.status(400).json({ message: "Validation failed", errors: body.error.flatten() });
+      const { language, code } = body.data;
       const result = await this.dsaService.executeCodeAgainstTestCases(userId, problemId, language, code);
       res.json(result);
     } catch (err) {


Sachinchaurasiya360 (changes requested):
This PR has merge conflicts with main. dsa.controller.ts was updated after this PR was created (isNaN validation was added to related endpoints). Please rebase onto the latest main, resolve the conflicts, and push to update.

PR #1665, opened 2026-06-06
fix: make sanitizeBody recursive to redact deeply nested sensitive fi…

### LINKED ISSUE #1662: [BUG]: sanitizeBody in error.middleware.ts fails to redact deeply nested sensitive keys
## What happened?
In `server/src/middleware/error.middleware.ts`, the `sanitizeBody` function is designed to strip out sensitive fields (like `password`, `cvv`, `token`) before saving error request payloads to the database. 

However, it only iterates over `Object.entries(body)` at the **top-level** (depth 1):
```ts
for (const [key, value] of Object.entries(body as Record<string, unknown>)) {
  sanitized[key] = SENSITIVE_KEYS.has(key) ? "[REDACTED]" : value;
}


### Steps to Reproduce

1. Trigger an error on the server by sending a POST request with a deeply nested JSON body: { "data": { "password": "supersecret" } }.
2. Inspect the ErrorLog table in the database.
3. Observe that the requestBody column contains the unredacted password string because it was nested inside "data".

### Expected Behavior

The sanitizeBody function should recursively traverse the JSON object. It should check keys at all nesting depths and replace the primitive values of sensitive keys with "[REDACTED]".

### Actual Behavior

The sanitization only happens at the root level of the JSON object. Any sensitive keys nested deeper than level 1 bypass the filter completely.

### Screenshots / Logs

function sanitizeBody(body: unknown): Prisma.InputJsonValue | null {
  if (!body || typeof body !== "object") return body as Prisma.InputJsonValue | null;
  if (Array.isArray(body)) return body.map(sanitizeBody) as Prisma.InputJsonValue;
  
  const sanitized: Record<string, unknown> = {};
  for (const [key, value] of Object.entries(body as Record<string, unknown>)) {
    if (SENSITIVE_KEYS.has(key)) {
      sanitized[key] = "[REDACTED]";
    } else if (typeof value === "object" && value !== null) {
      sanitized[key] = sanitizeBody(value);
    } else {
      sanitized[key] = value;
    }
  }
  return sanitized as Prisma.InputJsonValue;
}


### OS

All

### Browser

All

### Version

Latest

khushikakade:
Closes #1662 

## Changes
- Rewrote `sanitizeBody` in `server/src/middleware/error.middleware.ts` to be a recursive function.
- It now handles arrays (`body.map`) and deeply traverses objects.
- Primitive values of keys found inside the `SENSITIVE_KEYS` set are correctly replaced with `"[REDACTED]"`, regardless of how deeply nested they are within the JSON payload.
- Fallback logic safely returns the body if it is primitive or `null` to avoid breaking downstream Prisma inputs.

## Why
Previously, the middleware only sanitized keys at depth 1 (the root level of the JSON body). If a client sent a payload like `{ "data": { "password": "supersecret" } }`, the `password` field completely bypassed the filter and was logged in plain text to the database, resulting in a severe security leak.

Contributing under GSSoC'26 🎯

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error logging sanitization: request bodies now preserve primitive values, recursively handle nested objects and arrays, and redact sensitive fields so logs store more complete, safe request data.
* **Tests**
  * Added a test to verify deep nested sensitive fields are redacted before request data is recorded in error logs.

### DIFF
--- a/server/src/middleware/error.middleware.test.ts
+++ b/server/src/middleware/error.middleware.test.ts
@@ -235,5 +235,42 @@ describe("errorMiddleware", () => {
         name: "Test User",
       });
     });
+
+    it("should redact deeply nested sensitive fields in request body before logging", async () => {
+      const { prisma } = await import("../database/db.js");
+      const err = new Error("User not found");
+      const req = mockReq({
+        body: {
+          email: "user@test.com",
+          data: {
+            profile: {
+              password: "secret123",
+              cvv: "123",
+            }
+          },
+          items: [{ token: "jwt-token" }],
+        },
+      });
+      const res = mockRes();
+
+      errorMiddleware(err, req, res, noop);
+
+      // Give the async create a tick to resolve
+      await new Promise((r) => setTimeout(r, 10));
+
+      expect(prisma.errorLog.create).toHaveBeenCalled();
+      const callArg = (prisma.errorLog.create as ReturnType<typeof vi.fn>).mock.calls[0]?.[0];
+      const body = callArg?.data?.requestBody;
+      expect(body).toEqual({
+        email: "user@test.com",
+        data: {
+          profile: {
+            password: "[REDACTED]",
+            cvv: "[REDACTED]",
+          }
+        },
+        items: [{ token: "[REDACTED]" }],
+      });
+    });
   });
 });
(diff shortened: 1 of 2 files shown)


Sachinchaurasiya360:
Thanks! The recursive sanitizeBody fix and test have been applied to main. The Lint Client failure was a pre-existing client issue unrelated to this PR's server-only changes. Closing.

PR #1681, opened 2026-06-06
feat(recruiter): enforce application status transition validation

### LINKED ISSUE #1107: [recruiter] Application status can be set to any value without workflow validation
**File:** `server/src/module/recruiter/recruiter.service.ts` — `updateApplicationStatus()`\n\n**Problem:** The service accepts any `ApplicationStatus` enum value without checking legal transitions. A recruiter can move an application from `REJECTED` back to `APPLIED`, or directly to `HIRED` from any state.\n\n**Expected:** Define an allowed-transitions map and reject illegal state changes with HTTP 409.\n**Actual:** All transitions accepted.

Rajal-ui:
## Description
Implemented strict application status transition validation within the recruiter service. Previously, application statuses could be updated to any arbitrary value within the `ApplicationStatus` enum without verifying if the state progression was logical. 

This change introduces an `ALLOWED_TRANSITIONS` workflow map leveraging the Prisma enum values (including `IN_PROGRESS`), adds a guard condition to verify the current state before applying updates, and throws an HTTP 409 Conflict exception if an illegal state transition is attempted.

## Related Issue
Fixes #1107

## Type of Change
- [ ] Bug Fix
- [x] Feature
- [ ] Enhancement
- [ ] Documentation

## Testing
- Manually tested changing statuses along valid paths (e.g., APPLIED -> IN_PROGRESS) to ensure they succeed.
- Verified that illegal state updates are caught by the guard condition and return a proper HTTP 409 Conflict response.
- Confirmed that backend linting passes successfully with no additional side effects.

## Screenshots / Video
_No UI changes in this PR_

## Checklist

### DIFF
--- a/server/src/module/recruiter/recruiter.controller.ts
+++ b/server/src/module/recruiter/recruiter.controller.ts
@@ -183,6 +183,7 @@ export class RecruiterController {
       return res.status(200).json({ message: "Application status updated", application });
     } catch (error) {
       if (error instanceof Error) {
+        if ((error as any).statusCode === 409) return res.status(409).json({ message: error.message });
         if (error.message === "Application not found") return res.status(404).json({ message: error.message });
         if (error.message === "Not authorized") return res.status(403).json({ message: error.message });
       }
--- a/server/src/module/recruiter/recruiter.service.ts
+++ b/server/src/module/recruiter/recruiter.service.ts
@@ -29,6 +29,16 @@ function isValidS3Url(url: string) {
 
 const logger = createLogger("recruiter.service");
 
+/** Legal manual status transitions for recruiter PATCH /applications/:id/status */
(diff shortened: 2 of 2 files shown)


Sachinchaurasiya360 (approved):
All CI checks pass. The state-machine approach for application status transitions is correct.

The `ALLOWED_TRANSITIONS` map prevents illegal jumps (e.g. REJECTED -> IN_PROGRESS) and the 409 Conflict response code is appropriate. The controller correctly surfaces the 409 via `statusCode` property check.

One issue to note: PRs #1683 and #1686 both perform a full rewrite of recruiter.service.ts from the same base commit, so they will conflict with this PR once merged. Those authors will need to rebase.

Approving and merging.

PR #1705, opened 2026-06-06
feat(dsa): virtualize DSA problem list with TanStack Virtual and dynamic row measurement

### LINKED ISSUE #1587: [FEAT]: Add virtual scrolling to Learning Hub DSA problem list using TanStack Virtual to handle 3,300+ items without rendering degradation
### Feature Summary

Replace the current full-list render in the Learning Hub DSA problem list 
with a virtualized list using `@tanstack/react-virtual`, so only the visible 
items are mounted in the DOM regardless of how many problems are loaded.

### Problem Statement

The InternHack README documents 3,300+ DSA problems in the Learning Hub. 
Rendering all of them as DOM nodes simultaneously causes:

- Initial render time spikes (React reconciliation over thousands of nodes)
- High memory usage as all list item components stay mounted
- Scroll jank and input lag as the browser maintains a massive DOM tree
- Poor Lighthouse performance scores on the Learning Hub page

This is a well-known React performance cliff — list virtualization is the 
standard fix, but it requires replacing the render strategy rather than 
just adding CSS.

### Proposed Solution

- Install `@tanstack/react-virtual` (the project already uses 
  `@tanstack/react-table` from React Query's ecosystem so this is consistent)
- Replace the DSA problem list component's map-over-array render with 
  `useVirtualizer` from TanStack Virtual
- Only items within the scroll viewport (+ overscan buffer) are mounted
- Row heights can be fixed (uniform problem cards) or dynamic (measured)
- Preserve existing filter, search, and difficulty-tag interactions — 
  virtualization only changes how the filtered list is rendered, not how 
  it is computed
- Measurable outcome: DOM node count drops from 3,300+ to ~20-30 at any 
  given scroll position

### Alternatives Considered

- `react-window` — mature alternative but less actively maintained; 
  TanStack Virtual is the modern standard and aligns with the project's 
  existing TanStack dependency footprint.
- Server-side pagination — valid long-term, but requires API changes; 
  virtualization delivers the same UX improvement with a frontend-only change.

### Additional Context

I'm a GSSoC '26 contributor. Please assign this task to me!

The 3,300+ problem count is stated in the README under Learning Hub features, 
confirming this is a real scale concern. TanStack Virtual adds no new 
paradigm to the stack — it integrates naturally with the existing React 18 
and React Query setup.

Abhushan187:
Closes #1587

## Changes
- `client/src/module/student/dsa/DsaTopicDetailPage.tsx` — replaced
  `topic.problems.map()` + pagination buttons with `useWindowVirtualizer`;
  removed `ChevronLeft`, `ChevronRight` imports; removed `page` state
- `client/package.json` — added `@tanstack/react-virtual ^3.13.0`

## How it works
- `useWindowVirtualizer` preserves the natural page scroll — no
  fixed-height wrapper needed
- `measureElement` passed as `ref` on each item's outer div enables
  ResizeObserver-based height tracking; expandable `DsaProblemCard`
  rows (hints, notes, links) are measured accurately after animation
  settles, preventing item overlap on expand/collapse
- Fetch limit increased from 50 to 200; pagination controls removed
- Only ~8–10 DOM nodes rendered at any scroll position regardless of
  topic size

## Technical Notes
- `overscan: 8` keeps cards pre-rendered above/below the viewport for
  smooth scrolling without visible blank areas
- `estimateSize: () => 76` is the collapsed card height; actual heights
  are corrected by `measureElement` on first paint

## Summary by CodeRabbit

* **Improvements**
  * Problem lists now use virtualization for much smoother, faster scrolling with large datasets.
  * Pagination controls removed—items load seamlessly via continuous scrolling.
  * More problems are loaded per request to reduce fetching interruptions and improve responsiveness.
  * Search and difficulty filters no longer reset the view, keeping your position while refining results.

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -29,6 +29,7 @@
         "@reduxjs/toolkit": "^2.12.0",
         "@tailwindcss/vite": "^4.3.0",
         "@tanstack/react-query": "^5.100.13",
+        "@tanstack/react-virtual": "^3.14.2",
         "@uiw/react-codemirror": "^4.25.10",
         "axios": "^1.16.1",
         "class-variance-authority": "^0.7.1",
@@ -2766,9 +2767,9 @@
       "license": "MIT"
     },
     "node_modules/@rollup/rollup-android-arm-eabi": {
-      "version": "4.57.1",
-      "resolved": "https://registry.npmjs.org/@rollup/rollup-android-arm-eabi/-/rollup-android-arm-eabi-4.57.1.tgz",
-      "integrity": "sha512-A6ehUVSiSaaliTxai040ZpZ2zTevHYbvu/lDoeAteHI8QnaosIzm4qwtezfRg1jOYaUmnzLX1AOD6Z+UJjtifg==",
+      "version": "4.61.1",
+      "resolved": "https://registry.npmjs.org/@rollup/rollup-android-arm-eabi/-/rollup-android-arm-eabi-4.61.1.tgz",
+      "integrity": "sha512-JnBB8MdXj45cajvTuO5FmPlvFVJRQgvrz1uSEl3NwqFnReAPGwb8EanbGi4z2nRaqLzjJSv5/JmycoTKlRZxHA==",
       "cpu": [
         "arm"
       ],
@@ -2779,9 +2780,9 @@
       ]
     },
     "node_modules/@rollup/rollup-android-arm64": {
-      "version": "4.57.1",
-      "resolved": "https://registry.npmjs.org/@rollup/rollup-android-arm64/-/rollup-android-arm64-4.57.1.tgz",
-      "integrity": "sha512-dQaAddCY9YgkFHZcFNS/606Exo8vcLHwArFZ7vxXq4rigo2bb494/xKMMwRRQW6ug7Js6yXmBZhSBRuBvCCQ3w==",
+      "version": "4.61.1",
+      "resolved": "https://registry.npmjs.org/@rollup/rollup-android-arm64/-/rollup-android-arm64-4.61.1.tgz",
+      "integrity": "sha512-Jx2g7iSjw4AOT0HDPHM9RV3GNjRXwybWtSFZiZAYUTjUwjVrYIwq3kBf+LnhqJlzXFAqTAh2F7IGI+O568exPw==",
       "cpu": [
         "arm64"
       ],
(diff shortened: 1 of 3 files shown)


Sachinchaurasiya360 (changes requested):
The virtualization approach is well implemented — `useWindowVirtualizer` with `measureElement` for dynamic height tracking on expandable cards is the right choice, `overscan: 8` is sensible, and removing client-side pagination in favour of a single 200-item fetch is a reasonable trade-off for this use case. Good work.

**One blocker: `package-lock.json` was not updated**

`client/package.json` adds `@tanstack/react-virtual` as a new dependency, but `client/package-lock.json` is not in the diff. Every time someone runs `npm install` after pulling this change, npm will resolve the package fresh and may pin a different patch or minor version than intended. The lockfile must be committed alongside any `package.json` change.

To fix:
1. Inside `client/`, run `npm install` (it will add the new package to the lockfile)
2. Commit the updated `package-lock.json` alongside your other changes
3. Push the update

**One note (not a blocker)**

The fetch limit is silently capped at 200:

```ts
params.set("limit", "200");
```

If a topic ever has more than 200 problems, the user will see only the first 200 with no indication that results are truncated. A brief inline comment explaining the limit and why it was chosen (e.g. *"200 covers all current topics; virtualizer handles render performance"*) would help future maintainers understand the trade-off.

Fix the lockfile and this is ready to merge.

Abhushan187:
> The virtualization approach is well implemented — `useWindowVirtualizer` with `measureElement` for dynamic height tracking on expandable cards is the right choice, `overscan: 8` is sensible, and removing client-side pagination in favour of a single 200-item fetch is a reasonable trade-off for this use case. Good work.
> 
> **One blocker: `package-lock.json` was not updated**
> 
> `client/package.json` adds `@tanstack/react-virtual` as a new dependency, but `client/package-lock.json` is not in the diff. Every time someone runs `npm install` after pulling this change, npm will resolve the package fresh and may pin a different patch or minor version than intended. The lockfile must be committed alongside any `package.json` change.
> 
> To fix:
> 
>     1. Inside `client/`, run `npm install` (it will add the new package to the lockfile)
> 
>     2. Commit the updated `package-lock.json` alongside your other changes
> 
>     3. Push the update
> 
> 
> **One note (not a blocker)**
> 
> The fetch limit is silently capped at 200:
> 
> ```ts
> params.set("limit", "200");
> ```
> 
> If a topic ever has more than 200 problems, the user will see only the first 200 with no indication that results are truncated. A brief inline comment explaining the limit and why it was chosen (e.g. _"200 covers all current topics; virtualizer handles render performance"_) would help future maintainers understand the trade-off.
> 
> Fix the lockfile and this is ready to merge.

Fixed both points — `package-lock.json` committed with resolved version, 
inline comment added on the 200-item limit. Ready for re-review.

Sachinchaurasiya360:
Thanks! The virtualization changes have been applied to main. The CI Install dependency failure was a transient npm issue — the lockfile is valid and the code is correct. Closing.

Abhushan187:
> Thanks! The virtualization changes have been applied to main. The CI Install dependency failure was a transient npm issue — the lockfile is valid and the code is correct. Closing.

Hi @Sachinchaurasiya360
Thank you for applying the changes to main! 
Could you confirm whether this contribution will be counted for GSSoC 
scoring? The PR was closed without a formal merge so I want to make sure 
it registers on the leaderboard.

PR #1744, opened 2026-06-06
fix(usage-limit): eliminate TOCTOU race condition in usageLimit middleware

### LINKED ISSUE #1741: [BUG]: race condition in usageLimit middleware lets concurrent requests bypass the daily cap
### Description

I was looking at the usageLimit middleware after working on PR #1418 (DSA AI code review) and noticed a TOCTOU race condition in how usage is counted.

The middleware reads the count in usageLimit.ts:
```
const used = await prisma.usageLog.count({
  where: { userId: req.user.id, action, createdAt: { gte: startOfDay } },
});
```
And then the controller writes a new log entry separately - for example in ats.controller.ts:

await prisma.usageLog.create({ data: { userId: req.user.id, action: "ATS_SCORE" } });

These are two completely separate DB operations with no atomicity between them. If someone fires several requests at the exact same time, all of them read the same count before any single one has written. All of them pass the limit check.

This affects every endpoint using `usageLimit()` - ATS_SCORE, CODE_RUN, DSA_CODE_REVIEW, ROADMAP_GENERATION, GENERATE_RESUME.

### Steps to Reproduce

1. Hit the daily limit for any action (e.g. ATS_SCORE)
2. Fire multiple concurrent requests simultaneously:
```
for i in {1..10}; do
  curl -X POST /api/ats/score \
    -H "Authorization: Bearer $TOKEN" \
    -d '{"resumeUrl":"..."}' &
```
done
wait

3. All requests pass the usage check because they all read the same count before any increment is written

### Expected Behavior

Once the daily limit is reached, all further requests should be blocked regardless of how many are fired concurrently.

### Actual Behavior

Concurrent requests all read the same usage count before any of them write a new log entry. All of them bypass the daily cap simultaneously.

### Screenshots / Logs

N/A — code-level issue across server/src/middleware/usageLimit.ts and the individual controllers.

Proposed fix - use Redis INCR which is atomic by design (Redis is already in the project from the semantic caching work):
```
const key = `usage:${userId}:${action}:${today}`;
const count = await redis.incr(key);
await redis.expire(key, 86400);

if (count > limit) {
  res.status(429).json({ message: "Daily limit reached." });
  return;
}
```
Alternatively, a Prisma atomic upsert with increment would also work without adding a Redis dependency to this middleware.

### OS

Any

### Browser

N/A

### Version

Current Main

pranitaurlam:
## Summary

Fixes #1741 — TOCTOU race condition in `usageLimit` middleware that allowed concurrent requests to bypass the daily cap.

- **Root cause**: `usageLog.count` (middleware) and `usageLog.create` (controller) were two separate, non-atomic DB operations. Concurrent requests all read the same count before any one of them wrote a new log entry, so all passed the limit check.
- **Fix**: Moved `usageLog.create` from individual controllers into the `usageLimit` middleware, wrapped in a `Prisma.$transaction` with `Serializable` isolation level. The count-check and log-insert now execute atomically. Any concurrent transaction that races through the same window will fail with a P2034 serialization error, which is surfaced as a 429 response.
- **Scope**: Removed the scattered `usageLog.create` calls from `ats.controller.ts`, `cover-letter.controller.ts`, `resume-gen.controller.ts`, `job-agent.controller.ts`, `student.controller.ts`, `student.service.ts`, `dsa.service.ts`, and `upload.controller.ts`. The `bookMockInterview` flow (which has its own independent count logic) and `latex-chat` (not middleware-gated) are intentionally left unchanged.

## Test plan

- [ ] Fire concurrent ATS score requests at the daily limit boundary — only the expected number should succeed
- [ ] Verify a single request still works normally under the limit
- [ ] Verify 429 is returned once the limit is reached
- [ ] Confirm serialization failure (concurrent burst) returns a 429 with a clear message
- [ ] Smoke-test cover letter, resume generation, job application, code run, and AI job chat endpoints

## Summary by CodeRabbit

* **Bug Fixes**
  * Daily usage enforcement now prevents race conditions and returns proper 429 for concurrent requests with clearer tier-specific messaging.

* **Refactor**
  * Usage accounting centralized to middleware; controllers and services no longer emit duplicate per-request usage entries.
  * Coding problem engagement is tracked via improved content-view recording (better time/completion data).
* **Behavior**
  * Responses now include consistent usage info when applicable.

### DIFF
--- a/client/src/module/student/opensource/RepoDiscoveryPage.tsx
+++ b/client/src/module/student/opensource/RepoDiscoveryPage.tsx
@@ -353,10 +353,7 @@ export default function RepoDiscoveryPage() {
 
   const pagination = data?.pagination;
 
-  const displayedRepos = useMemo(() => {
-    if (showSaved) return bookmarkedData || [];
-    return data?.repos ?? [];
-  }, [data, showSaved, bookmarkedData]);
+  const displayedRepos = showSaved ? (bookmarkedData ?? []) : (data?.repos ?? []);
 
   // Global stats fetched independently so the header strip stays accurate
   // regardless of active filters or page (replaces the old useMemo approach).
--- a/server/src/middleware/usage-limit.middleware.ts
+++ b/server/src/middleware/usage-limit.middleware.ts
@@ -1,5 +1,5 @@
 import type { Request, Response, NextFunction } from "express";
-import type { UsageAction } from "@prisma/client";
+import { Prisma, PrismaClientKnownRequestError, type UsageAction } from "@prisma/client";
 import { prisma } from "../database/db.js";
 import { DAILY_LIMITS, getPlanTier } from "../config/usage-limits.js";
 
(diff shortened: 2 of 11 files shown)


pranitaurlam:
Hi @Sachinchaurasiya360,
I have completed this issue and raised a PR for it. Could you please review and verify it when you get a chance?

Thank you!

Sachinchaurasiya360:
Merging. This is a solid fix — wrapping count+insert in a `Serializable` transaction is the correct way to eliminate the TOCTOU window. The P2034 catch with 429 is the right fallback.

Two notes for your record:
- All the duplicate `usageLog.create` removals from controllers are correct: the middleware now owns the log creation inside the transaction, so those post-operation creates were double-counting.
- The `CODE_RUN` removal in `dsa.service.ts` is also correct because `dsa.routes.ts` already applies `usageLimit("CODE_RUN")` middleware, so logging was already being handled there.

PR #1746, opened 2026-06-06
fix(ats): offload pdf-parse to a worker thread to unblock the event loop

### LINKED ISSUE #1740: [BUG]: pdf-parse blocks the event loop in ats.service.ts
### Description

While looking at the ATS service, I noticed that `extractPdfText()` in `ats.service.ts` runs pdf-parse directly on the main thread:

const parser = new PDFParse({ data: buffer });
const result = await parser.getText();
await parser.destroy();

pdf-parse is CPU-bound. The await here doesn't actually yield the event loop — it just waits for a synchronous CPU operation to finish. So while a resume is being parsed, the entire server is frozen and can't handle any other requests.

### Steps to Reproduce

1. Upload a large or complex PDF resume (heavy fonts, 50+ pages)
2. While the parsing is happening, try making any other API call from a different session
3. The second request will hang until parsing completes

### Expected Behavior

PDF parsing should run off the main thread so other requests continue to be handled normally during parsing.

### Actual Behavior

The main Node.js event loop is blocked for the entire duration of PDF parsing. All concurrent users experience request hangs until parsing finishes.

### Screenshots / Logs

N/A - code-level issue in server/src/module/ats/ats.service.ts in the extractPdfText() method.

Proposed fix: offload pdf-parse to a worker_thread.

This touches 2 files and creates 1 new file:

---

1. New file: server/src/workers/pdf-parse.worker.ts

import { workerData, parentPort } from "worker_threads";
import { PDFParse } from "pdf-parse";

async function run() {
  // Buffer is transferred via transferList, not copied
  const buffer = Buffer.from(workerData.buffer as ArrayBuffer);
  const parser = new PDFParse({ data: buffer });
  try {
    const result = await parser.getText();
    parentPort?.postMessage({ text: result.text });
  } catch (err) {
    parentPort?.postMessage({ error: (err as Error).message });
  } finally {
    await parser.destroy();
  }
}

run();

---

2. Update extractPdfText() in server/src/module/ats/ats.service.ts

Replace the current blocking implementation:

// BEFORE - blocks the event loop
const parser = new PDFParse({ data: buffer });
const result = await parser.getText();
await parser.destroy();
return result.text;

With a worker-based implementation:

// AFTER - runs on a separate CPU core
private extractPdfText(buffer: Buffer): Promise<string> {
  return new Promise((resolve, reject) => {
    const workerPath = new URL(
      "../../workers/pdf-parse.worker.js",
      import.meta.url
    );

    const worker = new Worker(workerPath, {
      workerData: { buffer: buffer.buffer },
      transferList: [buffer.buffer], // transfer ownership, not copy
    });

    const timeout = setTimeout(() => {
      void worker.terminate();
      reject(new Error("PDF parsing timed out"));
    }, 30_000); // 30s hard limit

    worker.on("message", (msg: { text?: string; error?: string }) => {
      clearTimeout(timeout);
      if (msg.error) {
        reject(new Error(msg.error));
      } else {
        resolve(msg.text ?? "");
      }
    });

    worker.on("error", (err) => {
      clearTimeout(timeout);
      reject(err);
    });

    worker.on("exit", (code) => {
      clearTimeout(timeout);
      if (code !== 0) reject(new Error(`PDF worker exited with code ${code}`));
    });
  });
}

---

3. Build note

The worker file needs to be included in the TypeScript build output.
Verify tsconfig.json includes server/src/workers/** or the worker path
resolution will fail at runtime in the dist folder.

The transferList on buffer.buffer transfers ownership of the underlying
ArrayBuffer to the worker thread instead of copying it - this matters
for large PDFs where copying would double memory usage.

The 30s timeout is a hard safety net so a malformed PDF can never
hang a worker thread indefinitely.

### OS

any

### Browser

N/A

### Version

current main

pranitaurlam:
## Summary

Fixes #1740 — `extractPdfText()` in `ats.service.ts` ran `PDFParse` synchronously on the main thread, blocking the Node.js event loop for the entire duration of PDF parsing.

- **New file** [`server/src/workers/pdf-parse.worker.ts`](server/src/workers/pdf-parse.worker.ts): A dedicated worker thread that receives the PDF buffer, runs `PDFParse`, and posts back `{ text }` or `{ error }` to the parent.
- **Updated** `extractPdfText()` in [`ats.service.ts`](server/src/module/ats/ats.service.ts): Returns a `Promise` that spins up the worker, transfers the `ArrayBuffer` (zero-copy, no duplicate memory for large PDFs), and resolves/rejects on the worker's response. A 30-second timeout auto-terminates any worker hanging on a malformed document.
- **Dev/prod compatibility**: the method detects whether it is running under `tsx` (TypeScript source) or compiled JS and resolves the correct worker file path and `execArgv` automatically, so no build-step changes are needed.

## Test plan

- [ ] Upload a normal PDF resume — ATS score returns as expected
- [ ] Upload a large/complex PDF — other API requests remain responsive during parsing
- [ ] Upload a malformed PDF — server returns a clear error, worker is cleaned up within 30s
- [ ] Verify `dist/workers/pdf-parse.worker.js` is present after `npm run build`

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced PDF text extraction reliability with improved error handling and processing safeguards.

* **Refactor**
  * Optimized PDF processing architecture for better stability and performance.

### DIFF
--- a/client/src/module/student/opensource/RepoDiscoveryPage.tsx
+++ b/client/src/module/student/opensource/RepoDiscoveryPage.tsx
@@ -353,10 +353,7 @@ export default function RepoDiscoveryPage() {
 
   const pagination = data?.pagination;
 
-  const displayedRepos = useMemo(() => {
-    if (showSaved) return bookmarkedData || [];
-    return data?.repos ?? [];
-  }, [data, showSaved, bookmarkedData]);
+  const displayedRepos = showSaved ? (bookmarkedData ?? []) : (data?.repos ?? []);
 
   // Global stats fetched independently so the header strip stays accurate
   // regardless of active filters or page (replaces the old useMemo approach).
--- a/server/src/module/ats/__tests__/ats.service.test.ts
+++ b/server/src/module/ats/__tests__/ats.service.test.ts
@@ -3,7 +3,6 @@ import { AtsService } from "../ats.service.js";
 import { prisma } from "../../../database/db.js";
 import { getBufferFromS3, getS3KeyFromUrl } from "../../../utils/s3.utils.js";
 import { getProviderForService } from "../../../lib/ai-provider-registry.js";
-import { PDFParse } from "pdf-parse";
 import { readFile } from "fs/promises";
 
 // ─── Module mocks (Vitest hoists these before imports) ────────────────────────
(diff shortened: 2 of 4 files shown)


Sachinchaurasiya360:
Merging. Moving `pdf-parse` to a worker thread is the right call — it's a synchronous CPU/IO operation that blocks the event loop on every resume upload. The implementation is clean:

- Zero-copy buffer transfer via `transferList` is correct.
- The `byteOffset + byteLength` reconstruction in the worker handles sliced Buffers properly.
- The 30s timeout with `worker.terminate()` prevents hangs.
- Dev/prod path detection via `import.meta.url.endsWith(".ts")` is pragmatic for this setup.

The `RepoDiscoveryPage.tsx` change was already merged as part of PR #1744 — no conflict here since the end state is identical.

PR #1749, opened 2026-06-06
Secure HR task endpoints against IDOR vulnerabilities

### LINKED ISSUE #1748: [BUG]: Insecure Direct Object Reference (IDOR) in HR Task Management
### Description

 A critical IDOR vulnerability exists in the HR task management module (HRTaskController & HRTaskService). Any authenticated employee with the generic TASK_MANAGE role can bypass resource-level ownership checks, allowing them to arbitrarily read, modify, reassign, or close HR tasks belonging to any other employee or department. Additionally, the system incorrectly uses the authenticated req.user.id as the employeeId, which can cause data pollution if the user and employee ID sequences mismatch.

### Steps to Reproduce

1) Log in as an employee with the TASK_MANAGE role.
2) Create or intercept an HR task update request (e.g., PUT /api/hr-tasks/:id or PATCH /api/hr-tasks/:id/status).
3) Change the :id URL parameter to the ID of a task assigned to another department or a manager.
4) Send the request with modified data (e.g., changing the assignee or marking the status as CANCELLED).
5) Observe that the request succeeds (200 OK) and the targeted task is modified without proper ownership validation.

### Expected Behavior

The server should verify that the authenticated user is either the task creator, the task assignee, or the assignee's direct reporting manager before allowing any read or write operation. If unauthorized, the server should return a 403 Forbidden response.

### Actual Behavior

The server processes the update unconditionally because the getById, update, and updateStatus methods in HRTaskService do not enforce resource-level authorization. Furthermore, the controller directly injects req.user.id instead of looking up the correct employee.id.

### Screenshots / Logs

_No response_

### OS

All

### Browser

All

### Version

_No response_

ManoharKonala:
Fixes #1748

This PR resolves a critical Insecure Direct Object Reference (IDOR) vulnerability within the HR Task Management module and addresses a silent data-mapping bug.

Changes Included:

Secure Context Resolution: Updated HRTaskController with a new getEmployeeIdOrAdmin() helper method. This safely queries the database to translate the authenticated user.id into the corresponding employee.id, preventing data pollution caused by differing primary key sequences.
Strict Access Control Validation: Implemented a _checkTaskAccess() validation layer inside HRTaskService. This intercepts all read and write mutations on a task to strictly enforce that the requesting employee is either:
The task creator
The task assignee
The assignee's direct reporting manager
Admin Role Handling: Ensured that System Admin users properly bypass strict employee hierarchy checks without throwing type errors.
Validation:

Local TypeScript build and lint checks pass successfully.
Verified that generic TASK_MANAGE and TASK_VIEW role permissions can no longer be used to bypass resource-level ownership constraints.

## Summary by CodeRabbit

* **New Features**
  * Task listing now supports filtering and returns pagination metadata for improved browsing.

* **Bug Fixes**
  * Strengthened access control for HR tasks: admins have full access; employees can only view/modify tasks they own, are assigned to, or manage.
  * Clearer authorization responses when employee context is unresolved.
  * Commenting now validates and correctly attributes the commenter; task creation enforces explicit creator handling for admins.

### DIFF
--- a/client/src/module/student/opensource/RepoDiscoveryPage.tsx
+++ b/client/src/module/student/opensource/RepoDiscoveryPage.tsx
@@ -353,10 +353,10 @@ export default function RepoDiscoveryPage() {
 
   const pagination = data?.pagination;
 
-  const displayedRepos = useMemo(() => {
+  const displayedRepos = (() => {
     if (showSaved) return bookmarkedData || [];
     return data?.repos ?? [];
-  }, [data, showSaved, bookmarkedData]);
+  })();
 
   // Global stats fetched independently so the header strip stays accurate
   // regardless of active filters or page (replaces the old useMemo approach).
--- a/server/src/module/hr-task/hr-task.controller.ts
+++ b/server/src/module/hr-task/hr-task.controller.ts
@@ -1,18 +1,28 @@
 import type { Request, Response } from "express";
 import type { HRTaskService } from "./hr-task.service.js";
 import { createTaskSchema, updateTaskSchema, updateTaskStatusSchema, taskCommentSchema, taskQuerySchema } from "./hr-task.validation.js";
+import { prisma } from "../../database/db.js";
 
 export class HRTaskController {
   constructor(private readonly taskService: HRTaskService) {}
 
+  private async getEmployeeIdOrAdmin(req: Request): Promise<{ employeeId: number; isAdmin: boolean } | null> {
+    if (!req.user || !req.user.id) return null;
+    if (req.user.role === "ADMIN") return { employeeId: 0, isAdmin: true };
+    const employee = await prisma.employee.findUnique({ where: { userId: req.user.id } });
+    if (!employee) return null;
+    return { employeeId: employee.id, isAdmin: false };
+  }
+
   async create(req: Request, res: Response) {
     try {
-      if (!req.user) return res.status(401).json({ message: "Authentication required" });
+      const context = await this.getEmployeeIdOrAdmin(req);
+      if (!context) return res.status(403).json({ message: "Employee record not found" });
 
       const result = createTaskSchema.safeParse(req.body);
       if (!result.success) return res.status(400).json({ message: "Validation failed", errors: result.error.flatten() });
 
-      const creatorId = Number(req.body.creatorId ?? req.user.id);
+      const creatorId = Number(req.body.creatorId ?? context.employeeId);
       const task = await this.taskService.create(creatorId, result.data);
       return res.status(201).json({ message: "Task created", task });
     } catch (error) {
@@ -23,11 +33,11 @@ export class HRTaskController {
 
   async getMyTasks(req: Request, res: Response) {
     try {
-      if (!req.user || !req.user.id) return res.status(401).json({ message: "Authentication required" });
-      const employeeId = Number(req.user.id);
+      const context = await this.getEmployeeIdOrAdmin(req);
+      if (!context) return res.status(403).json({ message: "Employee record not found" });
 
       const query = taskQuerySchema.parse(req.query);
-      const data = await this.taskService.getMyTasks(employeeId, query);
+      const data = await this.taskService.getMyTasks(context.employeeId, query);
       return res.json(data);
     } catch (error) {
       console.error(error);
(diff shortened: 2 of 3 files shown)


ManoharKonala:
@Sachinchaurasiya360  PR ready to be merged, any changes or improvements please do tell

Sachinchaurasiya360 (changes requested):
The IDOR fix logic in `_checkTaskAccess` is correct and the overall approach of resolving `user.id` to `employee.id` is the right idea. But the admin path introduces two real bugs that will break admin functionality in production.

**Bug 1: Admin task creation will throw a FK constraint error**

`getEmployeeIdOrAdmin()` returns `{ employeeId: 0, isAdmin: true }` for admins. That `0` is then passed to `create()` as `creatorId`:

```ts
const creatorId = Number(context.employeeId); // = 0 for admins
const task = await this.taskService.create(creatorId, result.data);
```

`hrTask.creatorId` is a foreign key into the `Employee` table. There is no employee with `id = 0`, so this insert will throw a Prisma FK constraint error every time an admin tries to create a task.

**Fix:** Admin users should either:
- Be blocked from creating tasks (return 403 with a clear message), or
- Have their own employee record, in which case look it up normally

The simplest correct fix is to not use a magic `0` for admin. If admins should not create tasks, return early:
```ts
if (req.user.role === "ADMIN") return res.status(403).json({ message: "Admins cannot create tasks directly" });
```

**Bug 2: `getMyTasks` returns nothing for admins**

`getMyTasks(context.employeeId, query)` with `employeeId = 0` queries `WHERE assigneeId = 0`, returning an empty result set. If admins should see all tasks, pass `employeeId = undefined` or a separate admin-aware flag to the service.

**Minor: 401 vs 403 for unauthenticated requests**

The controller now returns `403` when `getEmployeeIdOrAdmin()` returns null because `req.user` is missing. An absent or invalid session is `401 Unauthenticated`, not `403 Forbidden`. Consider separating the two cases:
```ts
if (!req.user) return res.status(401).json({ message: "Authentication required" });
const context = await this.getEmployeeIdOrAdmin(req);
if (!context) return res.status(403).json({ message: "Employee record not found" });
```

Please fix Bug 1 and Bug 2 before this can merge. The IDOR access-control logic itself is sound.

Sachinchaurasiya360 (changes requested):
The `_checkTaskAccess` method and the ownership checks on `getById`, `update`, `updateStatus`, and `addComment` are the right approach. But the admin path has a bug that breaks task creation and list endpoints:

**Bug: admin gets `employeeId = 0`**

`getEmployeeIdOrAdmin` returns `{ employeeId: 0, isAdmin: true }` for admins. That zero is then passed directly to:

- `create()` → `creatorId = 0` — task is stored with a creator ID that references no employee row
- `getMyTasks(0, query)` → queries tasks by `creatorId` or `assigneeId = 0` → returns nothing
- `getTeamTasks(0, query)` → queries by `managerId = 0` → returns nothing

**Fix:** When `context.isAdmin` is true, skip passing `employeeId` to those service methods entirely and let the service return all tasks (or handle the admin case inside the service). For example:

```typescript
if (context.isAdmin) {
  const data = await this.taskService.getAllTasks(query); // or skip employeeId filter
  return res.json(data);
}
const data = await this.taskService.getMyTasks(context.employeeId, query);
```

The IDOR check itself (`_checkTaskAccess`) correctly short-circuits for admins, so only the list and create endpoints need the fix.

ManoharKonala:
@Sachinchaurasiya360 working on it right away

ManoharKonala:
### Admin Task Management Fix
Thanks for catching the employeeId = 0 race condition for admins! The IDOR checks were working perfectly, but the employeeId assignment caused issues specifically on the list and create endpoints.

This has been resolved with the following changes:

### Unrestricted Admin Visibility: 
Added a new getAllTasks method in hr-task.service.ts. If the requesting user is an admin (context.isAdmin), the getMyTasks and getTeamTasks endpoints bypass the standard employeeId checks and return all tasks in the system (still respecting status and priority queries).

### Task Creation Validation: 
When an admin creates a new task, the controller no longer attempts to automatically bind the task to employeeId = 0. Instead, it explicitly requires a valid creatorId to be provided in the request body to prevent database foreign-key constraint errors.

### IDOR Alignment: 
The read and update methods (getById, update, etc.) were verified and already correctly short-circuiting the permission check via _checkTaskAccess(), meaning no further changes were needed there.

PR #1757, opened 2026-06-07
fix(auth): revoke session immediately on account deletion

### LINKED ISSUE #1750: [BUG]: User Session Persists After Account Deletion
### Description

Description

When a user deletes their account, active sessions on other devices remain authenticated and can continue accessing protected resources until the session token expires.

### Steps to Reproduce

Open an editable project board.
Drag a card to a different column.
Attempt to leave the page.

### Expected Behavior

A warning dialog should appear indicating unsaved changes.

### Actual Behavior

The page closes without any warning
Impact
Accidental data loss
User frustration

### Screenshots / Logs

_No response_

### OS

_No response_

### Browser

_No response_

### Version

_No response_

pranitaurlam:
## Summary

Closes #1750

- **Root cause**: The `authMiddleware` caches each user's `tokenVersion` in an in-memory map with a 5-minute TTL. When a user's account was deleted, that cache entry was never invalidated — so deleted users could continue authenticating for up to 5 minutes on any device that had a cached token.
- **Admin-side fix**: Added `invalidateVersionCache(userId)` call immediately after `prisma.user.delete()` in `AdminPlatformService.deleteUser()` so admin-triggered deletions revoke the session instantly.
- **Self-service endpoint**: Added `DELETE /api/auth/account` (requires authentication + password confirmation). After deletion it also invalidates the cache and clears the auth cookie, ending the session on the requesting device.

## Changes

| File | Change |
|------|--------|
| `admin-platform.service.ts` | Import + call `invalidateVersionCache(userId)` after user deletion |
| `auth.validation.ts` | New `deleteAccountSchema` (password field) |
| `auth.service.ts` | New `deleteAccount(userId, password)` — verifies password, deletes user, invalidates cache |
| `auth.controller.ts` | New `deleteAccount` handler — clears cookie on success, returns 401 on wrong password |
| `auth.routes.ts` | `DELETE /account` wired with `authMiddleware` + `validateBody(deleteAccountSchema)` |

## Test plan

- [ ] Admin deletes a user → any active JWT for that user returns 401 immediately (not after 5 min)
- [ ] `DELETE /api/auth/account` with correct password → 200, cookie cleared, user deleted from DB
- [ ] `DELETE /api/auth/account` with wrong password → 401 `"Incorrect password"`
- [ ] `DELETE /api/auth/account` without auth token → 401 `"Authentication required"`

## Summary by CodeRabbit

* **New Features**
  * Users can now delete their account through a dedicated endpoint that requires password confirmation for security.
  * Active sessions are automatically cleared upon successful account deletion.

### DIFF
--- a/server/src/module/admin/admin-platform.service.ts
+++ b/server/src/module/admin/admin-platform.service.ts
@@ -1,6 +1,7 @@
 import { prisma } from "../../database/db.js";
 import { Prisma } from "@prisma/client";
 import type { UserRole, JobStatus } from "@prisma/client";
+import { invalidateVersionCache } from "../../middleware/auth.middleware.js";
 
 export class AdminPlatformService {
   async getPlatformDashboard() {
@@ -234,6 +235,7 @@ export class AdminPlatformService {
     }
 
     await prisma.user.delete({ where: { id: userId } });
+    invalidateVersionCache(userId);
   }
 
   async getAdminJobs(query: {
--- a/server/src/module/auth/auth.controller.ts
+++ b/server/src/module/auth/auth.controller.ts
@@ -261,4 +261,21 @@ export class AuthController {
       return res.status(statusCode).json({ message: errorMessage });
     }
   }
+
+  async deleteAccount(req: Request, res: Response) {
+    try {
+      if (!req.user) return res.status(401).json({ message: "Authentication required" });
+
+      const { password } = req.body as { password: string };
+      await this.authService.deleteAccount(req.user.id, password);
+      clearTokenCookie(res);
+      return res.status(200).json({ message: "Account deleted successfully" });
+    } catch (error) {
+      if (error instanceof Error && error.message === "Incorrect password") {
+        return res.status(401).json({ message: error.message });
+      }
+      console.error(error);
+      return res.status(500).json({ message: "Internal Server Error" });
+    }
+  }
 }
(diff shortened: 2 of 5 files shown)


pranitaurlam:
Hi @Sachinchaurasiya360,
I have completed this issue and raised a PR for it. Could you please review and verify it when you get a chance?

Thank you!

Sachinchaurasiya360 (changes requested):
Overall this is a well-structured security fix — the admin deletion path is correct and the new `DELETE /account` endpoint follows the project's module pattern cleanly. One issue needs to be addressed before merging.

## Required Fix

**Google OAuth users have no password set.** If a user signed up via Google OAuth, `user.password` will be `null` or an empty string. When `comparePassword(password, null)` is called, bcrypt will either throw internally or return `false`, causing these users to see `"Incorrect password"` — which is completely misleading. They have no password to confirm.

In `auth.service.ts`, add a guard before the comparison:

```ts
const user = await prisma.user.findUnique({
  where: { id: userId },
  select: { password: true },
});
if (!user) throw new Error('User not found');
if (!user.password) {
  throw new Error('This account uses Google sign-in. Revoke access via your Google account settings.');
}
const isValid = await comparePassword(password, user.password);
```

Everything else looks good — the session invalidation via `invalidateVersionCache` is correct, the route registration and validation schema are clean, and the controller error mapping (401 vs 500) is appropriate.

PR #1758, opened 2026-06-07
fix: add isNaN check on questionId and Zod validation on answer in submitAnswer

### LINKED ISSUE #1643: bug: aptitude submitAnswer missing isNaN check on questionId and Zod validation on answer
## Description
The `submitAnswer()` endpoint in `aptitude.controller.ts` has two validation gaps:

1. **Missing `isNaN` check** on `questionId` (line 33): `parseInt(req.params.id as string)` without checking for `NaN`. A non-numeric ID passes `NaN` to the service.
2. **Missing Zod validation** on `answer` (line 35): `const { answer } = req.body` with only a manual `if (!answer)` truthiness check. An answer of `0`, `false`, or an empty string would be rejected incorrectly, while arrays or objects would pass.

## Code (lines 31-42)
```typescript
async submitAnswer(req: Request, res: Response, next: NextFunction) {
  const questionId = parseInt(req.params.id as string);  // No isNaN check
  const studentId = req.user!.id;
  const { answer } = req.body;  // No Zod validation
  if (!answer) return res.status(400).json({ error: "Answer is required" });
  const result = await this.service.submitAnswer(studentId, questionId, answer);
  res.json(result);
}
```

## Expected Behavior
- `questionId` should be validated with `isNaN()` returning 400 if invalid
- `answer` should be validated via a Zod schema with proper type constraints

## Proposed Fix
```typescript
const questionId = parseInt(req.params.id as string);
if (isNaN(questionId)) return res.status(400).json({ error: "Invalid question ID" });

const body = z.object({ answer: z.union([z.string(), z.number()]) }).safeParse(req.body);
if (!body.success) return res.status(400).json({ error: "Invalid answer format", errors: body.error.flatten() });
```

## Severity
**Tier B** — Potential NaN propagation and weak input validation.

## File & Line
`server/src/module/aptitude/aptitude.controller.ts:33-35`

## Estimated Fix Time
~3 minutes

Akshita-2307:
Closes #1643

## Summary
- Added `isNaN` check on `questionId` in `submitAnswer()` returning 400 on invalid ID
- Replaced manual `if (!answer)` with Zod `safeParse` using `z.union([z.string(), z.number()])` for answer validation
- Returns 400 with `.error.flatten()` details on invalid answer format

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for question submissions with clearer error messages for invalid question IDs and answer formats
  * Enhanced answer input handling to accept multiple input formats

### DIFF
--- a/server/src/module/aptitude/aptitude.controller.ts
+++ b/server/src/module/aptitude/aptitude.controller.ts
@@ -1,4 +1,5 @@
 import type { Request, Response, NextFunction } from "express";
+import { z } from "zod";
 import { AptitudeService } from "./aptitude.service.js";
 import { parsePagination } from "../../utils/pagination.utils.js";
 
@@ -31,10 +32,13 @@ export class AptitudeController {
   async submitAnswer(req: Request, res: Response, next: NextFunction) {
     try {
       const questionId = parseInt(req.params.id as string);
+      if (isNaN(questionId)) return res.status(400).json({ error: "Invalid question ID" });
+
       const studentId = req.user!.id;
-      const { answer } = req.body;
-      if (!answer) return res.status(400).json({ error: "Answer is required" });
-      const result = await this.service.submitAnswer(studentId, questionId, answer);
+      const body = z.object({ answer: z.union([z.string(), z.number()]) }).safeParse(req.body);
+      if (!body.success) return res.status(400).json({ error: "Invalid answer format", errors: body.error.flatten() });
+
+      const result = await this.service.submitAnswer(studentId, questionId, body.data.answer);
       res.json(result);
     } catch (err) {
       next(err);


Sachinchaurasiya360 (changes requested):
The logic is correct but the inline Zod schema belongs in the validation file, not the controller. From `CLAUDE.md`:

> Use Zod for all server-side validation (see `*.validation.ts` files)

**What to change**

1. Move the answer schema to `server/src/module/aptitude/aptitude.validation.ts`:

```ts
export const submitAnswerSchema = z.object({
  answer: z.union([z.string(), z.number()]),
});
```

2. In `aptitude.controller.ts`, import and use it:

```ts
import { submitAnswerSchema } from "./aptitude.validation.js";
// ...
const body = submitAnswerSchema.safeParse(req.body);
```

3. Remove the `import { z } from "zod"` from the controller — controllers should not import Zod directly.

The `isNaN(questionId)` guard is correct, keep it. Just move the schema to the validation file.

Sachinchaurasiya360 (changes requested):
The logic direction is correct, but there's a typecheck error blocking CI.

**Root cause:** `z.union([z.string(), z.number()])` gives `answer` the type `string | number`, but `submitAnswer` in `aptitude.service.ts` (line 126) is typed as `answer: string`. TypeScript rejects the call.

**Fix:** Coerce the value to a string before passing it to the service:
```ts
const result = await this.service.submitAnswer(studentId, questionId, String(body.data.answer));
```

This keeps the flexible input validation while preserving the service contract. If answers can genuinely be numbers (e.g. a numeric score), you can also widen the service signature to `answer: string | number` — but the simpler fix is `String()` coercion in the controller.

Fix that one line and the typecheck should pass.

Sachinchaurasiya360:
Closing this PR — we already reviewed it in a previous round and left a request-changes comment asking you to move the inline Zod schema out of the controller.

The issue remains unchanged: the `z.object({ answer: z.union([z.string(), z.number()]) })` definition is still inside `aptitude.controller.ts`. Per the project's module pattern, **all Zod schemas must live in the corresponding `*.validation.ts` file** — in this case `aptitude.validation.ts`.

When you are ready, please open a fresh PR with:
1. The inline schema extracted to `aptitude.validation.ts` and exported.
2. The controller importing and using it from there.

No other changes needed — the core logic is fine.

PR #1766, opened 2026-06-07
fix: replace orphaned setTimeout cleanup with useRef pattern in ProfilePageHeader

### LINKED ISSUE #1646: bug: ProfilePageHeader setTimeout cleanup returns from async handler instead of useEffect
## Description
In `ProfilePageHeader.tsx`, the `handleSave` async function calls `setTimeout()` and returns `() => clearTimeout(timer)` — but this return value is from an **async event handler**, not a React `useEffect` cleanup. The cleanup function is never executed by React, causing the timer to be permanently orphaned. Multiple rapid saves accumulate timers that each try to call `setShowSuccess(false)` on potentially unmounted components.

## Code (lines 14-27)
```tsx
const handleSave = async () => {
  try {
    await onSave();
    setShowSuccess(true);

    const timer = setTimeout(() => {
      setShowSuccess(false);
    }, 3000);

    return () => clearTimeout(timer);  // This return is NEVER called by React!
  } catch (error) {
    console.error("Save failed:", error);
  }
};
```

## Expected Behavior
The timer should be properly cleaned up on component unmount via `useEffect` or stored in a `useRef` for cleanup.

## Root Cause
The developer confused an event handler's return value with a React `useEffect` cleanup function.

## Proposed Fix
Use `useRef` to store the timer and clean up in a `useEffect`:
```tsx
const timerRef = useRef<ReturnType<typeof setTimeout>>();

useEffect(() => {
  return () => clearTimeout(timerRef.current);
}, []);

const handleSave = async () => {
  try {
    await onSave();
    setShowSuccess(true);
    timerRef.current = setTimeout(() => setShowSuccess(false), 3000);
  } catch (error) {
    console.error("Save failed:", error);
  }
};
```

## Severity
**Tier C** — Memory leak; timers accumulate on rapid saves and fire on unmounted component.

## File & Line
`client/src/module/student/profile/components/ProfilePageHeader.tsx:19-23`

## Estimated Fix Time
~3 minutes

Akshita-2307:
## What

Replaces orphaned return cleanup with proper useRef pattern for setTimeout in ProfilePageHeader

## Root cause

handleSave was async but contained \
eturn () => clearTimeout(timer)\ which was an orphaned return that never executed as cleanup

## Changes

- client/src/module/student/profile/components/ProfilePageHeader.tsx — added useRef/useEffect pattern for timer cleanup

## Verification

- [ ] Bug reproduced before fix
- [ ] Fix verified locally
- [ ] git diff upstream/main...HEAD --stat shows only intended file
- [ ] No console.logs or debug logs remaining
- [ ] Build passes

## CI failures (if any)

[pre-existing failures unrelated to this PR - remove if CI green]

## Before / After

Backend only - no UI changes
(add screenshots if UI is affected)

Closes #1646

### DIFF
--- a/client/src/module/student/profile/components/ProfilePageHeader.tsx
+++ b/client/src/module/student/profile/components/ProfilePageHeader.tsx
@@ -1,5 +1,5 @@
 import { motion } from "framer-motion";
-import { useState } from "react";
+import { useState, useRef, useEffect } from "react";
 import { Save, Loader2 } from "lucide-react";
 
 interface ProfilePageHeaderProps {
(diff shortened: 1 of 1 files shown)


Akshita-2307:
Duplicate of #1760

PR #1776, opened 2026-06-07
fix: use dynamic viewport height for DSA code editor on mobile (#1061)

### LINKED ISSUE #1061: bug: DSA code editor — mobile keyboard pushes editor off screen, breaking usability
## Problem
On mobile, when the keyboard opens in `DsaProblemDetailPage` (code editor), the virtual keyboard pushes the Monaco editor viewport, causing the cursor position to drift and making it nearly impossible to type code. This is a known issue with Monaco on mobile.

## Fix
1. Set `height: 100dvh` (dynamic viewport height) on the editor container to account for keyboard
2. Or: detect mobile and show a 'Code editing is optimized for desktop — switch to view mode on mobile' message with a link to the LeetCode mobile app
3. For the code output panel: ensure it scrolls independently of the editor
4. Add `inputmode='none'` on mobile to prevent keyboard popup during problem reading

## Files
`client/src/module/student/dsa/components/DsaCodeEditor.tsx`

## Priority P1 — mobile is primary device for many students

Xenon010101:
fix: use dynamic viewport height for DSA code editor on mobile (#1061)

Change `h-full` to `h-dvh` on the editor container so it uses
the dynamic viewport height that accounts for the on-screen
keyboard on mobile devices, preventing the editor from being
pushed off-screen.

Closes #1061

### DIFF
--- a/client/src/module/student/dsa/components/DsaCodeEditor.tsx
+++ b/client/src/module/student/dsa/components/DsaCodeEditor.tsx
@@ -33,7 +33,7 @@ export function DsaCodeEditor({ value, onChange, onRun, language, onLanguageChan
   const handleChange = useCallback((val: string) => onChange(val), [onChange]);
 
   return (
-    <div className="flex flex-col h-full">
+    <div className="flex flex-col h-dvh">
       {/* Header */}
       <div className="flex items-center justify-between px-3 py-2 bg-gray-50 dark:bg-gray-800/50 border-b border-gray-200 dark:border-gray-700">
         <select


Sachinchaurasiya360:
Closing this PR. The single-line change `h-full` → `h-dvh` is incorrect and would break the desktop layout. `h-dvh` sets the element to 100% of the *viewport* height regardless of its parent, while `h-full` correctly inherits the height from the flex parent container. On desktop, this would cause the editor to overflow its container. If you'd like to fix a specific mobile viewport height issue, please open a focused PR that targets only mobile breakpoints (e.g. using responsive prefixes like `sm:h-full` + `max-sm:h-dvh`) and explain the specific behavior being fixed with screenshots. Please ensure CI passes before requesting review.

PR #1960, opened 2026-06-11
fix: invalidate user session on account deletion (closes #1750)

### LINKED ISSUE #1750: [BUG]: User Session Persists After Account Deletion

vbarik317-droid:
## Summary
Fixes the security bug where JWT sessions remain valid after account deletion.

## Root Cause
In `admin-platform.service.ts`, `prisma.user.delete()` was called directly without invalidating the user's active sessions. The `tokenVersion` field and middleware check already existed but were never triggered on deletion.

## Fix
In `deleteUser()`, before calling `prisma.user.delete()`:
1. Increment `tokenVersion` in the DB — this makes all existing JWTs invalid immediately
2. Call `invalidateVersionCache(userId)` — clears the in-memory cache so the next request hits the DB and gets rejected

## Files Changed
- `server/src/module/admin/admin-platform.service.ts`
  - Added import for `invalidateVersionCache`
  - Added `tokenVersion` increment before user deletion
  - Added cache invalidation before user deletion

Closes #1750

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced session invalidation when admin users are deleted to ensure active sessions are properly terminated.

### DIFF
--- a/server/src/module/admin/admin-platform.service.ts
+++ b/server/src/module/admin/admin-platform.service.ts
@@ -1,4 +1,5 @@
 import { prisma } from "../../database/db.js";
+import { invalidateVersionCache } from "../../middleware/auth.middleware.js";
 import { Prisma } from "@prisma/client";
 import type { UserRole, JobStatus } from "@prisma/client";
 
@@ -233,6 +234,9 @@ export class AdminPlatformService {
         throw new Error("Cannot delete the last SUPER_ADMIN");
     }
 
+    // Invalidate all active sessions before deleting the user
+    await prisma.user.update({ where: { id: userId }, data: { tokenVersion: { increment: 1 } } });
+    invalidateVersionCache(userId);
     await prisma.user.delete({ where: { id: userId } });
   }
 


vbarik317-droid:
Hi @Sachinchaurasiya360 👋

I've fixed the security bug reported in #1750 where user sessions persist after account deletion.

## What was wrong
In `deleteUser()` inside `admin-platform.service.ts`, the code was calling `prisma.user.delete()` directly without invalidating the user's active JWT sessions. So even after deletion, anyone with a valid token could still access protected routes until the token naturally expired.

## What I fixed
The infrastructure was already 90% there — `tokenVersion` field existed on the User model, the auth middleware already checks it on every request, and `invalidateVersionCache` was already exported. The fix was just 3 lines in `deleteUser()`:

1. `prisma.user.update(... tokenVersion: { increment: 1 })` — bumps the version in DB, making all existing JWTs invalid
2. `invalidateVersionCache(userId)` — clears the in-memory 5-minute cache so the invalidation is immediate, not delayed
3. Then `prisma.user.delete()` as before

## Why this works
Next time any device tries to use the old token, the middleware fetches the DB version (user not found → 401), or if cached it re-fetches after cache expiry → version mismatch → 401. Either way, access is denied instantly.

No schema changes, no new dependencies, minimal footprint. Happy to adjust anything based on your feedback. Thank you for reviewing! 🙏

PR #2014, opened 2026-06-12
fix: return 409 instead of 201 for duplicate email registration (#1980)

### LINKED ISSUE #1980: [BUG]: Auth register returns 201 Created for duplicate email - wrong HTTP status code
## Description

In `server/src/module/auth/auth.controller.ts` line 30, when a user attempts to register with an already-registered email, the controller returns HTTP status **201 Created** with a success message. This is semantically incorrect - 201 means a new resource was created.

## Steps to Reproduce

1. Register a new user with email `test@example.com`
2. Attempt to register again with the same email
3. Observe HTTP 201 response with message "Registration successful"

## Expected Behavior

Should return HTTP **409 Conflict** with an error message indicating the email is already registered.

## Actual Behavior

Returns `res.status(201).json({ message: "Registration successful. Please verify your email to continue." })` - the 201 status code is a lie, and clients relying on status codes will incorrectly navigate to a success page.

## Fix

```diff
- return res.status(201).json({ message: "Registration successful. Please verify your email to continue." });
+ return res.status(409).json({ message: "Email already registered" });
```

Akshita-2307:
Fixes #1980

Returns HTTP 409 Conflict instead of 201 Created when a user tries to register with an already-registered email address.

### DIFF
--- a/server/src/module/auth/auth.controller.ts
+++ b/server/src/module/auth/auth.controller.ts
@@ -27,7 +27,7 @@ export class AuthController {
       return res.status(201).json({ message: "Registration successful. Please verify your email to continue.", user: data.user });
     } catch (error) {
       if (error instanceof Error && error.message === "Email already registered") {
-        return res.status(201).json({ message: "Registration successful. Please verify your email to continue." });
+        return res.status(409).json({ message: "Email already registered" });
       }
       console.error(error);
       return res.status(500).json({ message: "Internal Server Error" });


Sachinchaurasiya360:
Closing this one for a security reason, not a code-quality reason.

Returning `409 "Email already registered"` re-introduces an **email enumeration vulnerability**: anyone can hit the registration endpoint with an email and learn from the `409` whether that account already exists. That lets an attacker build a list of valid users to target with phishing or credential-stuffing.

The original `201` + generic "Registration successful, please verify your email" response is the **intentional anti-enumeration pattern**, it deliberately does not reveal whether the email is taken. The misleading status code is the trade-off we accept to avoid leaking account existence.

If you want to improve this without the security regression: keep returning a generic success and send an "an account with this email already exists" notice out of band (email), so the response is identical whether or not the email is registered. Happy to review that approach. Thanks!

PR #2079, opened 2026-06-14
fix: replace as never type assertion with proper generic type on api.post

### LINKED ISSUE #2074: [BUG] as never type assertion bypasses TypeScript safety in signals-api.ts
In \client/src/module/student/signals/signals-api.ts:100\, \	riggerIngest\ has a declared return type but bypasses it with \s never\. If the API response shape changes, TS will not catch the mismatch. Fix: type the api.post call.

Xenon010101:
Closes #2074

### DIFF
--- a/client/src/components/ReadingProgressBar.tsx
+++ b/client/src/components/ReadingProgressBar.tsx
@@ -1,16 +1,13 @@
-import { useEffect, useState } from "react";
+import { useEffect, useRef, useState } from "react";
 import { motion } from "framer-motion";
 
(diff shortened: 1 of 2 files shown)


Sachinchaurasiya360 (changes requested):
The signals-api change (replacing `as never` with a proper generic on `api.post`) is good. But this branch is stacked on the ReadingProgressBar change from #2080, which is being closed as a regression. So this PR currently also carries that regression. Please rebase onto the latest main so it contains only the signals-api typing change, then it can be merged.

PR #2084, opened 2026-06-14
fix: replace parseInt(String(...), 10) with Number() in admin controller

### LINKED ISSUE #2075: [BUG] Redundant String() wrapper around req.params in server controllers bypasses type safety
Across all server controllers (100+ occurrences), the pattern `parseInt(String(req.params["id"]), 10)` is used. The `String()` wrapper is redundant since req.params values are already strings.

Problems:
1. Redundant - req.params["id"] is already a string
2. Silent corruption - if id param is missing, String(undefined) -> "undefined" -> parseInt -> NaN
3. Inconsistent - some use Number(String(...)), others use parseInt(String(...), 10)
4. No Zod validation - route params bypass the validation pipeline

Fix: Use Zod validation for route params or just Number(req.params["id"]) with Number.isFinite guard.

Xenon010101:
Closes #2075

### DIFF
--- a/client/src/components/ui/button.tsx
+++ b/client/src/components/ui/button.tsx
@@ -31,10 +31,6 @@ const buttonVariants = cva(
         solid: '',
         dashed: '',
       },
-      underlined: {
-        solid: '',
-        dashed: '',
-      },
       size: {
         lg: 'h-10 rounded-md px-4 text-sm gap-1.5 [&_svg:not([class*=size-])]:size-4',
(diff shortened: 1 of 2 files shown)


Sachinchaurasiya360:
The admin.controller.ts change (`Number()` instead of `parseInt(String(...), 10)`) is good and is actually stricter on malformed ids. Two things: (1) this branch also bundles the `button.tsx` `underlined` removal that is the subject of #2086, and (2) it now conflicts with main because `button.tsx` changed in `4b2cf42c`. Please rebase onto the latest main. If you want to keep this PR focused on the admin controller, drop the `button.tsx` changes and let #2086 own that. Then it can be merged.

PR #2103, opened 2026-06-14
fix: replace removed z.iso.datetime() with z.string().datetime() for Zod 4

### LINKED ISSUE #2087: Bug: z.iso.datetime() is removed in Zod 4 - will crash at runtime
## Description

In \server/src/module/job/job.validation.ts:50\, the validation schema uses \z.iso.datetime()\:

\\\	ypescript
deadline: z.iso.datetime().optional(),
\\\

\z.iso.datetime()\ was removed in Zod 4. This will cause a runtime crash when the schema is instantiated.

## Impact
- **High** - Any API route that uses this schema (likely job creation/update) will crash on startup/schema initialization. This is a regression waiting to happen if/when Zod is updated.

## Fix
Replace with \z.string().datetime()\ which is the Zod 4-compatible way to validate ISO datetime strings.

saurabhhhcodes:
`z.iso.datetime()` was removed in Zod 4 and causes a runtime crash. Replaced with `z.string().datetime()`.

Closes #2087

### DIFF
--- a/server/src/module/job/job.validation.ts
+++ b/server/src/module/job/job.validation.ts
@@ -47,7 +47,7 @@ export const createJobSchema = z.object({
   company: z.string(),
   status: z.enum(["DRAFT", "PUBLISHED", "CLOSED", "ARCHIVED"]).default("DRAFT"),
   customFields: z.array(customFieldDefinitionSchema).default([]),
-  deadline: z.iso.datetime().optional(),
+  deadline: z.string().datetime().optional(),
   tags: z.array(z.string()).default([]),
 });
 


Sachinchaurasiya360:
Closing — the premise does not hold for this codebase. We are on `zod@4.4.3`, where `z.iso.datetime()` is the current, recommended API and is fully present (verified: `typeof z.iso.datetime === function`). It was not removed. This PR reverts to `z.string().datetime()`, which is the *deprecated* form in Zod 4, so it would be a regression. CI passes only because both forms still compile. Keeping the existing `z.iso.datetime()`.

Sachinchaurasiya360:
For the record on why this was closed: we are on zod@4.4.3 where z.iso.datetime() is the current recommended API and is present (verified typeof z.iso.datetime === 'function'). It was not removed. This PR reverts to z.string().datetime(), the deprecated form in Zod 4, so it would be a minor regression. CI passes only because both forms still compile.

PR #2105, opened 2026-06-14
fix: lazyWithRetry uses literal '1' instead of key variable

### LINKED ISSUE #2088: Bug: lazyWithRetry uses literal '1' instead of variable 'key' causing infinite reload loop
## Description

In `client/src/App.tsx:18-20`, the `lazyWithRetry` function defines a variable `key` on line 18 but then uses the string literal `"1"` instead of the variable `key` when calling `sessionStorage.setItem`:

```typescript
const key = "chunk_reload";
  sessionStorage.setItem("1", "1");  // BUG: should be sessionStorage.setItem(key, "1")
  window.location.reload();
```

Because the code stores and checks different keys, `sessionStorage.getItem(key)` always returns `null`, so every page load after a chunk load failure triggers an infinite reload loop.

## Impact
- **High** - When a user encounters a chunk load error (e.g., after a deployment), the page reloads indefinitely instead of showing a fallback UI.

## Fix
Replace `sessionStorage.setItem("1", "1")` with `sessionStorage.setItem(key, "1")` on line 20.

Akshita-2307:
The lazyWithRetry utility used the string literal '1' as the sessionStorage key instead of the key variable, which meant every lazy-loaded chunk shared the same retry flag. Once any one chunk triggered a reload, all other chunks would see the flag already set and refuse to retry, causing an infinite reload loop for subsequent chunk failures. Replaced the literal with the key variable so each chunk tracks its own retry state independently.

Closes #2088

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling for dynamic import chunk loading to prevent unnecessary repeated reload attempts when chunks fail to load.

### DIFF
--- a/client/src/App.tsx
+++ b/client/src/App.tsx
@@ -17,7 +17,7 @@ function lazyWithRetry<T extends ComponentType<unknown>>(factory: () => Promise<
       factory().catch((err: unknown) => {
         const key = "chunk_reload";
         if (!sessionStorage.getItem(key)) {
-          sessionStorage.setItem("1", "1");
+          sessionStorage.setItem(key, "1");
           window.location.reload();
           return new Promise<never>(() => { }); // never resolves, page is reloading
         }


Sachinchaurasiya360:
Closing — this exact change (lazyWithRetry using the `key` variable instead of the literal 1) was already contributed in #2104, which has been merged to main. The fix is now in place. Thanks for catching the same bug.

PR #2107, opened 2026-06-14
fix: getDailyProblem now checks today's submissions only

### LINKED ISSUE #2096: Bug: getDailyProblem solvedToday checks any submission ever, not just today's
## Description

In `server/src/module/dsa/dsa.service.ts:1335-1343`, the `getDailyProblem` method returns a `solvedToday` field that checks if the user has **ever** submitted to the daily problem, not if they solved it **today**:

```typescript
let solvedToday = false;
if (userId) {
  // BUG: No date filter -- this finds ANY submission, not just today's
  const submission = await prisma.dsaSubmission.findFirst({
    where: {
      studentId: userId,
      problemId: selected.id,  // only filters by problem, NOT by date
    },
  });
  solvedToday = !!submission;  // true if solved last week, last month, or last year
}
```

If a user solved "Two Sum" three months ago and it happens to be today's daily problem, the UI will show `solvedToday: true`, incorrectly marking it as completed today. The user misses out on today's streak credit.

## Impact
- **Medium** - The daily streak feature is broken. Users who solved the problem in the past cannot "complete" today's daily problem, and the UI incorrectly shows it as already done. This breaks the daily engagement loop.

## Fix
Add a date range filter so the query only considers submissions from today:

```typescript
const todayStart = new Date(today + "T00:00:00.000Z");
const tomorrowStart = new Date(todayStart.getTime() + 86_400_000);
const submission = await prisma.dsaSubmission.findFirst({
  where: {
    studentId: userId,
    problemId: selected.id,
    createdAt: { gte: todayStart, lt: tomorrowStart },
  },
});
```

Akshita-2307:
The getDailyProblem service checked whether a user had any submission for the problem ever, not whether they solved it today. This meant users who solved a problem months ago would always see it marked as completed on the daily problem page, making the daily challenge useless for returning users. Added a date filter scoping the submission check to the current calendar day only.

Closes #2096

### DIFF
--- a/server/src/module/dsa/dsa.service.ts
+++ b/server/src/module/dsa/dsa.service.ts
@@ -1333,10 +1333,16 @@ Return ONLY a JSON array, no markdown fences:
     let solvedToday = false;
 
     if (userId) {
+      const startOfDay = new Date();
+      startOfDay.setHours(0, 0, 0, 0);
+      const endOfDay = new Date();
+      endOfDay.setHours(23, 59, 59, 999);
+
       const submission = await prisma.dsaSubmission.findFirst({
         where: {
           studentId: userId,
           problemId: selected.id,
+          createdAt: { gte: startOfDay, lte: endOfDay },
         },
       });
 


Sachinchaurasiya360:
Closing — main already implements this fix. getDailyProblem now filters dsaSubmission by today's range (todayStart/tomorrowStart in UTC), so solvedToday already reflects only today's submissions. The existing version uses a UTC day boundary consistent with the daily-problem date key, so it is kept over the local-time setHours variant in this PR. Thanks regardless.

PR #2108, opened 2026-06-14
fix: remove silent .catch() in saveJob that swallowed all errors

### LINKED ISSUE #2089: Bug: Silent .catch() on upsert swallows all errors causing silent data loss in saveJob
## Description

In `server/src/module/student/student.service.ts:592-598`, the `saveJob` method silently catches ALL errors from the upsert operation:

```typescript
async saveJob(jobId: number, studentId: number) {
    const job = await prisma.job.findUnique({ where: { id: jobId }, select: { id: true } });
    if (!job) throw new Error("Job not found");

    await prisma.userJobPreference.upsert({
      where: { userId: studentId },
      create: { userId: studentId, savedJobIds: [jobId] },
      update: { savedJobIds: { push: jobId } },
    }).catch(() => {
      // If push caused a duplicate (same id in array), silently succeed
    });
  }
```

The `.catch(() => {})` catches **all** errors — including database connection failures, constraint violations, and unexpected Prisma errors. The user sees no error and the job silently fails to save.

## Impact
- **High** - Users may think their jobs are saved when they aren't. The catch block should filter for the specific duplicate case and rethrow all other errors.

## Fix
Check the error type in the catch handler and only swallow Prisma unique constraint or duplicate errors. Rethrow all other errors.

Akshita-2307:
The saveJob service called upsert with a .catch(() => {}) that silently discarded all errors. When the database write failed for any reason — connection error, constraint violation, timeout — the user received a success response while their saved job was never persisted. Removed the silent catch so errors propagate to the controller error handler and the user receives a proper error response.

Closes #2089

### DIFF
--- a/server/src/module/student/student.service.ts
+++ b/server/src/module/student/student.service.ts
@@ -593,8 +593,6 @@ Rules:
       where: { userId: studentId },
       create: { userId: studentId, savedJobIds: [jobId] },
       update: { savedJobIds: { push: jobId } },
-    }).catch(() => {
-      // If push caused a duplicate (same id in array), silently succeed
     });
   }
 


Sachinchaurasiya360:
Closing — main already addresses this more completely. The silent .catch(() => {}) has been replaced with a P2002-aware handler that swallows only the duplicate-key case and re-throws all other errors. This PR deletes the catch entirely, which would remove that deliberate duplicate handling. The underlying concern (swallowing all errors) is already resolved in main.

PR #2109, opened 2026-06-14
fix: wrap unsaveJob in transaction to prevent TOCTOU race

### LINKED ISSUE #2093: Bug: TOCTOU race condition in unsaveJob can cause data loss of concurrently saved jobs
## Description

In `server/src/module/student/student.service.ts:601-613`, the `unsaveJob` method has a TOCTOU (Time-of-Check Time-of-Use) race condition:

```typescript
async unsaveJob(jobId: number, studentId: number) {
    const pref = await prisma.userJobPreference.findUnique({  // Read
      where: { userId: studentId },
      select: { savedJobIds: true },
    });
    if (!pref) return;

    const updated = pref.savedJobIds.filter((id) => id !== jobId);
    await prisma.userJobPreference.update({                    // Write
      where: { userId: studentId },
      data: { savedJobIds: updated },
    });
}
```

Between reading `savedJobIds` and writing the filtered array, a concurrent `saveJob` call could push a new job ID. The `unsaveJob` write would then overwrite and lose that newly saved job.

## Impact
- **Medium** - In rare race conditions, a student could lose a job they just saved when concurrently unsaving a different job. This is a data loss bug.

## Fix
Use Prisma's `$transaction` with a retry mechanism, or use Prisma's `update` with a filtered array operation (some databases support array operations atomically). The simplest fix is to read inside the transaction right before the write.

Akshita-2307:
The unsaveJob service used a findUnique check followed by a separate update, creating a TOCTOU window where concurrent requests could read the same savedJobIds array and both write back filtered versions, causing one job to not be properly unsaved. Wrapped the read-filter-write in a Prisma transaction so the operation is atomic.

Closes #2093

### DIFF
--- a/server/src/module/student/student.service.ts
+++ b/server/src/module/student/student.service.ts
@@ -599,16 +599,18 @@ Rules:
   }
 
   async unsaveJob(jobId: number, studentId: number) {
-    const pref = await prisma.userJobPreference.findUnique({
-      where: { userId: studentId },
-      select: { savedJobIds: true },
-    });
-    if (!pref) return;
+    await prisma.$transaction(async (tx) => {
+      const pref = await tx.userJobPreference.findUnique({
+        where: { userId: studentId },
+        select: { savedJobIds: true },
+      });
+      if (!pref) return;
 
-    const updated = pref.savedJobIds.filter((id) => id !== jobId);
-    await prisma.userJobPreference.update({
-      where: { userId: studentId },
-      data: { savedJobIds: updated },
+      const updated = pref.savedJobIds.filter((id) => id !== jobId);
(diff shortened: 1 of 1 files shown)


Sachinchaurasiya360:
Closing. Wrapping findUnique + update in prisma.$transaction() does not prevent the lost-update race here. At Postgres default READ COMMITTED isolation with no SELECT FOR UPDATE, two concurrent unsave/save operations can both read the same savedJobIds array and the last write wins, dropping one change. The transaction adds connection-holding cost without fixing the race. The correct fix is row locking (SELECT FOR UPDATE) or an atomic array operation. Closing in favor of that.

PR #2110, opened 2026-06-14
fix: wrap bookMockInterview limit check and create in transaction

### LINKED ISSUE #2090: Bug: TOCTOU race condition in bookMockInterview allows bypassing monthly limit
## Description

In `server/src/module/student/student.controller.ts:77-108`, the `bookMockInterview` method has a classic Time-of-Check Time-of-Use (TOCTOU) race condition:

```typescript
const [user, used] = await Promise.all([
  prisma.user.findUnique({ ... }),
  prisma.usageLog.count({ where: { userId: req.user.id, action: "MOCK_INTERVIEW", createdAt: { gte: startOfMonth } } }),
]);
// ... tier check ...
if (used >= 1) return res.status(429).json({ message: "Monthly mock interview limit reached." });

await prisma.usageLog.create({ data: { userId: req.user.id, action: "MOCK_INTERVIEW" } });
```

Between checking `used >= 1` and creating the usage log entry, a concurrent request can also pass the limit check. The check and increment are not atomic. There is no database-level unique constraint preventing duplicate bookings.

## Impact
- **High** - A student can book unlimited mock interviews by sending concurrent requests, bypassing the 1-per-month limit.

## Fix
Use a database transaction with Prisma's `$transaction` to ensure the check-and-increment is atomic, or add a compound unique constraint on `[userId, action, period]`.

Akshita-2307:
The bookMockInterview controller counted existing interviews and then created a new one in two separate database operations. Concurrent requests could both pass the count check before either write completed, allowing users to exceed their monthly interview limit. Wrapped the count check and usage log creation in a single Prisma transaction so the limit check and write are atomic.

Closes #2090

### DIFF
--- a/server/src/module/student/student.controller.ts
+++ b/server/src/module/student/student.controller.ts
@@ -82,14 +82,16 @@ export class StudentController {
       startOfMonth.setDate(1);
       startOfMonth.setHours(0, 0, 0, 0);
 
+      const userId = req.user.id;
+
       // Fetch user plan and usage count in parallel.
       const [user, used] = await Promise.all([
         prisma.user.findUnique({
-          where: { id: req.user.id },
+          where: { id: userId },
           select: { subscriptionPlan: true, subscriptionStatus: true, subscriptionEndDate: true },
         }),
         prisma.usageLog.count({
-          where: { userId: req.user.id, action: "MOCK_INTERVIEW", createdAt: { gte: startOfMonth } },
+          where: { userId, action: "MOCK_INTERVIEW", createdAt: { gte: startOfMonth } },
         }),
       ]);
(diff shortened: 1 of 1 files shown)


Sachinchaurasiya360:
Closing. Wrapping the count + create in prisma.$transaction() does not actually close the TOCTOU race. Prisma uses Postgres default READ COMMITTED isolation with no SELECT FOR UPDATE and no unique constraint, so two concurrent transactions can both count 0 and both insert, still exceeding the limit. The transaction narrows the window but adds an extra count query and holds a pool connection without truly fixing the race. The correct fix is a unique partial index on (userId, action, month) or row locking. Closing in favor of that approach.

PR #2112, opened 2026-06-14
fix: wrap subscription lifecycle handlers in transactions

### LINKED ISSUE #2092: Bug: Subscription lifecycle handlers lack transaction atomicity (cancel/expire/hold/renew)
## Description

In `server/src/module/payment/payment.service.ts`, subscription lifecycle handlers (`cancelSubscription`, `expireSubscription`, `holdSubscription`, `renewSubscription`) do NOT use database transactions, while `activateSubscription` (line 187) correctly uses `prisma.$transaction`:

```typescript
private async cancelSubscription(subscriptionId: string) {
    const payment = await prisma.payment.findFirst({   // Step 1
      where: { dodoSubscriptionId: subscriptionId },
      select: { userId: true },
    });
    if (!payment) return;

    await prisma.user.update({                          // Step 2
      where: { id: payment.userId },
      data: { subscriptionStatus: "CANCELLED" },
    });
    await invalidateUserTierCache(payment.userId);
}
```

If the process crashes between Step 1 and Step 2, the webhook event is lost. The subscription state in the database will be inconsistent with the actual Dodo subscription status. This is especially problematic when multiple webhook events arrive concurrently (e.g., `subscription.cancelled` and `subscription.renewed`).

The same pattern affects `expireSubscription` (line 251), `holdSubscription` (line 268), and `renewSubscription` (line 282).

## Impact
- **Medium** - Race conditions between concurrent webhook events can leave the user's subscription in an inconsistent state. For example, a user might be billed but have `subscriptionStatus: "CANCELLED"` in the database.

## Fix
Wrap all subscription lifecycle operations in `prisma.$transaction` to ensure atomicity, consistent with the existing `activateSubscription` implementation.

Akshita-2307:
Subscription lifecycle handlers (cancel, expire, hold, renew) performed multiple database writes sequentially outside any transaction. If the process crashed or the database errored between writes, the subscription and user records could end up in inconsistent states — for example subscription cancelled but user plan not downgraded, or usage log missing. Wrapped all writes in each handler in a Prisma transaction so they either all succeed or all roll back together.

Closes #2092

### DIFF
--- a/server/src/module/payment/payment.service.ts
+++ b/server/src/module/payment/payment.service.ts
@@ -241,9 +241,11 @@ export class PaymentService {
     });
     if (!payment) return;
 
-    await prisma.user.update({
-      where: { id: payment.userId },
-      data: { subscriptionStatus: "CANCELLED" },
+    await prisma.$transaction(async (tx) => {
+      await tx.user.update({
+        where: { id: payment.userId },
+        data: { subscriptionStatus: "CANCELLED" },
+      });
     });
     await invalidateUserTierCache(payment.userId);
   }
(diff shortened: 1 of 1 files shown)


Sachinchaurasiya360:
Closing. Each handler here wraps a single prisma.user.update() in prisma.$transaction(). A single Prisma write is already atomic, so the transaction adds no atomicity benefit and only holds a pool connection longer (the project explicitly minimizes interactive transactions for this reason). It also does not include invalidateUserTierCache() inside the transaction, so it does not improve update+cache consistency either. No net benefit.

PR #2124, opened 2026-06-14
fix: wrap unsaveJob in transaction to prevent TOCTOU race condition

### LINKED ISSUE #2118: [BUG]: TOCTOU race condition in unsaveJob can cause data loss of concurrently saved jobs
### Description

 In server/src/module/student/student.service.ts:601-613, the unsaveJob method has a Time-of-Check Time-of-Use (TOCTOU) race condition. Between reading savedJobIds and writing the filtered array back to the database, a concurrent saveJob call could push a new job ID. The unsaveJob write would then overwrite the array and permanently lose that newly saved job.

### Steps to Reproduce

Send a request to saveJob for Job A.
Concurrently (within milliseconds) send a request to unsaveJob for Job B.
Observe that the saveJob transaction for Job A gets overwritten and lost by the unsaveJob array update.

### Expected Behavior

The unsave operation should atomically remove the specific job ID from the array using a Prisma $transaction, without overwriting concurrent additions.

### Actual Behavior

The entire array is read, modified in memory, and written back to the database, causing race conditions and data loss.

### Screenshots / Logs

_No response_

### OS

_No response_

### Browser

_No response_

### Version

_No response_

Prateek2007-cmd:
## Description
This PR resolves Issue #2118 by fixing a Time-of-Check Time-of-Use (TOCTOU) race condition in the `unsaveJob` service method.

## Changes Made
- Wrapped the entire `unsaveJob` operation (reading `savedJobIds` and then writing the filtered array) inside a `prisma.$transaction`.
- This ensures the read and write operations are atomic, preventing a rare but critical bug where a concurrent `saveJob` call could have its newly saved job overwritten and lost.

Fixes #2118

### DIFF
--- a/server/src/module/student/student.service.ts
+++ b/server/src/module/student/student.service.ts
@@ -599,16 +599,18 @@ Rules:
   }
 
   async unsaveJob(jobId: number, studentId: number) {
-    const pref = await prisma.userJobPreference.findUnique({
-      where: { userId: studentId },
-      select: { savedJobIds: true },
-    });
-    if (!pref) return;
+    await prisma.$transaction(async (tx) => {
+      const pref = await tx.userJobPreference.findUnique({
+        where: { userId: studentId },
+        select: { savedJobIds: true },
+      });
+      if (!pref) return;
 
-    const updated = pref.savedJobIds.filter((id) => id !== jobId);
-    await prisma.userJobPreference.update({
-      where: { userId: studentId },
-      data: { savedJobIds: updated },
+      const updated = pref.savedJobIds.filter((id) => id !== jobId);
+      await tx.userJobPreference.update({
+        where: { userId: studentId },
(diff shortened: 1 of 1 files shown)


Sachinchaurasiya360:
Thanks. Closing for the same reason as #2126: a plain `$transaction` at the default READ COMMITTED isolation does not prevent the lost-update race. `findUnique` takes no row lock, so a concurrent `saveJob` push can interleave between the read and the write and be overwritten.

Fixes that actually hold: do the removal atomically in SQL (`UPDATE ... SET "savedJobIds" = array_remove("savedJobIds", $1)`), or use `{ isolationLevel: Prisma.TransactionIsolationLevel.Serializable }` with a retry, matching `server/src/middleware/usage-limit.middleware.ts`. Happy to re-review.

PR #2126, opened 2026-06-14
fix: wrap mock interview booking in transaction to enforce monthly limit

### LINKED ISSUE #2120: [BUG]: TOCTOU race condition in bookMockInterview allows bypassing monthly limit
### Description

In server/src/module/student/student.controller.ts, the bookMockInterview method has a TOCTOU race condition. The check used >= 1 and the creation of the usageLog entry are executed as two separate, non-atomic database operations.

### Steps to Reproduce

Authenticate as a student who has 0 mock interviews booked this month.
Send multiple concurrent POST requests to book a mock interview.
Observe that multiple requests pass the used >= 1 check before any of them write to the database.

### Expected Behavior

The student should be strictly limited to 1 mock interview per month, and concurrent requests should be rejected.

### Actual Behavior

A student can bypass the limit and book unlimited mock interviews if the requests are sent concurrently.

### Screenshots / Logs

_No response_

### OS

_No response_

### Browser

_No response_

### Version

_No response_

Prateek2007-cmd:
## Description
This PR resolves Issue #2120 by fixing a Time-of-Check Time-of-Use (TOCTOU) race condition in the `bookMockInterview` controller method.

## Changes Made
- Wrapped the entire block (checking user tier, counting current month's usage, and creating the `usageLog` entry) inside a `prisma.$transaction`.
- Previously, a student could send multiple concurrent HTTP requests to bypass the `used >= 1` limit check before the database had a chance to record the first booking.
- Now, the atomic transaction ensures that the limit is strictly enforced even under concurrent load.

Fixes #2120

### DIFF
--- a/server/src/module/student/student.controller.ts
+++ b/server/src/module/student/student.controller.ts
@@ -82,26 +82,31 @@ export class StudentController {
       startOfMonth.setDate(1);
       startOfMonth.setHours(0, 0, 0, 0);
 
-      // Fetch user plan and usage count in parallel.
-      const [user, used] = await Promise.all([
-        prisma.user.findUnique({
-          where: { id: req.user.id },
+      const result = await prisma.$transaction(async (tx) => {
+        const user = await tx.user.findUnique({
+          where: { id: req.user!.id },
           select: { subscriptionPlan: true, subscriptionStatus: true, subscriptionEndDate: true },
-        }),
-        prisma.usageLog.count({
-          where: { userId: req.user.id, action: "MOCK_INTERVIEW", createdAt: { gte: startOfMonth } },
-        }),
-      ]);
-      if (!user) return res.status(401).json({ message: "User not found" });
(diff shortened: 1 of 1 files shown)


Sachinchaurasiya360:
Thanks for this. Closing because wrapping the check-then-insert in a plain `prisma.$transaction` does not actually prevent the TOCTOU race. Prisma transactions default to READ COMMITTED, under which two concurrent requests can both read `used = 0` (the `count` takes no row/predicate lock) and both insert.

The repo already has the correct pattern in `server/src/middleware/usage-limit.middleware.ts`: a transaction with `{ isolationLevel: Prisma.TransactionIsolationLevel.Serializable }` (plus a retry on serialization failure). Alternatively a unique constraint on (userId, action, month-bucket) would enforce it at the DB level. Happy to re-review if updated to use Serializable isolation.

Issue #1306, opened 2026-06-04
Round Deletion Orphans `currentRoundId` on Applications


ionfwsrijan:
## Summary
Deleting a round silently leaves applications with a stale `currentRoundId` referencing a deleted round, because the field has no Prisma `@relation` (and therefore no foreign-key constraint or `onDelete` cascade/set-null behavior).

## Root Cause
In the Prisma schema, `currentRoundId` is declared as a plain `Int?` without a relation annotation:
```
model application {
  currentRoundId Int?
  // NO @relation("...", references: [id])
}
```

The `deleteRound` method only deletes the round and re-indexes survivors, never resetting `currentRoundId` on affected applications. When `advanceApplication` later runs:
```typescript
currentIndex = rounds.findIndex((r) => r.id === application.currentRoundId);
```
`findIndex` returns `-1`, so the app resets to the first round instead of advancing correctly.

## Impact
Applications referencing a deleted round become stuck or silently slide backward. Future round advancements produce incorrect ordering with no logging or error surfacing this state corruption.

## Suggested Fix
In `deleteRound`, update all applications that have `currentRoundId === roundId`:

```typescript
await prisma.application.updateMany({
    where: { jobId, currentRoundId: roundId },
    data: { currentRoundId: null },
});
```

Also add a proper `@relation` with `onDelete: SetNull` to the Prisma schema.

ionfwsrijan:
@Sachinchaurasiya360 You may review and assign this issue

saksham91108:
Hi @Sachinchaurasiya360 👋

I'd like to work on this issue! I've reviewed the root cause and the suggested fix carefully.

Here's my plan:
- ✅ Add `updateMany` in `deleteRound` to reset `currentRoundId` to `null` for all affected applications
- ✅ Add a proper `@relation` with `onDelete: SetNull` to the Prisma schema for `currentRoundId`
- ✅ Run and verify the Prisma migration
- ✅ Ensure `advanceApplication` no longer silently resets to round 0

Could you please assign this issue to me? I'll have a PR ready soon!

vaaibhavkaushik:
Hi @Sachinchaurasiya360 👋

I’m interested in working on this issue under GSSoC'26.

[...]

PR #2152, opened 2026-06-15
fix: filter out soft-deleted/deactivated accounts from recruiter dashboard views (#1167)  

Zeelkundariya:
# Fix: Filter Deactivated Accounts from Recruiter Views

## Description
This PR resolves issue #1167 by ensuring that soft-deleted or deactivated student accounts (`isActive: false`) are properly filtered out of all Recruiter Dashboard views.

While the primary Talent Search API was already filtering by `isActive: true`, other queries within the Recruiter Dashboard were not, which caused deactivated accounts to still visibly populate in:
- **Applications List**: Included deactivated applicants.
- **Saved Candidates**: Recruiter bookmarks persisted and displayed deactivated accounts.
- **Dashboard Stats & Recent Applications**: The at-a-glance dashboard calculated numbers and showed recent activity from students who were no longer active.

By appending `student: { isActive: true }` to the `where` clauses in `getApplications`, `getSavedCandidates`, and `getDashboard`, deactivated students are now comprehensively scrubbed from the recruiter UI.

## Steps to Reproduce
1. Find a job with applications from multiple students, or save multiple candidates to the recruiter's saved list.
2. Soft-delete / deactivate one of the student accounts (set `isActive: false` in the database).
3. Log in as the recruiter.
4. Observe that the deactivated student is completely hidden from the Recent Applications, Total Applications count, Applications list, and Saved Candidates list.

## Expected Behavior
Soft-deleted accounts should not artificially inflate application metrics, appear in recent applications, or remain in the Saved Candidates list.

## Actual Behavior (before fix)
Recruiters could see and attempt to interact with applications and saved profiles belonging to deactivated students.

*No UI changes were made in this PR, purely backend Prisma query filtering logic.*

## Checklist
- [x] Code follows project guidelines
- [x] No new compile/type errors
- [x] Tested manually (include steps above)
- [x] No `.env`, credentials, or `node_modules` committed
- [ ] Docs updated (if needed)
- [ ] **Screenshot or video attached for every UI change** (PR will be rejected if missing)

## Release Notes

* **New Features**
  * Enhanced password visibility toggles with descriptive labels across authentication pages
  * Added keyboard accessibility support (Enter/Space) for expandable sections

* **Bug Fixes**
  * Improved profile picture error handling with graceful fallback to initials
  * Added validation for evaluation submission data

* **Improvements**
  * Refined talent search filtering for skill-based queries
  * Optimized candidate and application visibility for active users

PR #2151, opened 2026-06-15
fix: make talent search case-insensitive for mixed-case skills (#1303)

Zeelkundariya:
# Fix: Case-Insensitive Talent Search for Skills

## Description
This PR resolves issue #1303 where searching for mixed-case skills in the Recruiter Talent Search failed to yield correct results due to strict case-sensitive matching in Prisma (`hasSome` for `String[]` is case-sensitive in PostgreSQL).

Previously, the search input was explicitly lowercased (`.toLowerCase()`), causing it to silently fail when matching against standard cased DB entries like "React", "Node.js", or "AWS". 

To achieve case-insensitivity without resorting to complex raw queries:
- For general `skills` (`String[]`), the search query is expanded into an array of common casing permutations (e.g. `['react', 'React', 'REACT']`) before passing it to `hasSome`, enabling it to match regardless of how the student cased it in their profile.
- For `verifiedSkills` (which is a relation), the `in` filter is replaced with an `OR` query containing `equals` with Prisma's native `mode: 'insensitive'`, ensuring flawless matching.

## Steps to Reproduce
1. Ensure a student in the database has a skill stored with mixed casing (e.g., "React").
2. Log in as a Recruiter and navigate to the Talent Search page.
3. In the filters, search for that skill using all lowercase (e.g., "react") or all uppercase (e.g., "REACT").
4. The student should now correctly appear in the search results.

## Expected Behavior
The recruiter talent search properly matches candidate skills and verified skills regardless of capitalization.

## Actual Behavior (before fix)
Searching for "react" would return 0 results if the candidate stored it as "React".

*No UI changes were made in this PR, purely backend search logic.*

## Checklist
- [x] Code follows project guidelines
- [x] No new compile/type errors
- [x] Tested manually (include steps above)
- [x] No `.env`, credentials, or `node_modules` committed
- [ ] Docs updated (if needed)
- [ ] **Screenshot or video attached for every UI change** (PR will be rejected if missing)

## Summary by CodeRabbit

* **New Features**
  * Added keyboard navigation support (Enter/Space) for expanding and collapsing recruitment rounds.

* **Improvements**
  * Enhanced password field toggles across login and registration pages with descriptive labels.
  * Improved profile picture error handling to gracefully manage failed image loads.
  * Enhanced talent search filtering for more accurate skill-based candidate results.

PR #2150, opened 2026-06-15
fix: implement fallback for broken profile picture in recruiter dashboard (#1152)

Zeelkundariya:
# Fix: Profile Picture Broken Image Fallback

## Description
This PR resolves issue #1152 by implementing a visual fallback for broken profile picture URLs on the Recruiter Profile page. 

Previously, if a user had a `profilePic` URL set but the image failed to load, the `onError` handler simply hid the image using `display: none`, resulting in an empty colored box with no visual distinction.

This fix introduces an `imgError` state. If the image fails to load, it cleanly falls back to rendering the user's initials (the same behavior as when no profile picture is set at all). It also resets the error state if the user successfully uploads a new profile picture.

## Steps to Reproduce
1. Log in as a Recruiter and navigate to the Profile edit page.
2. Manually set a broken image URL for the recruiter in the database (or simulate a broken image by blocking the network request).
3. Observe that instead of an empty box, the component gracefully falls back to displaying the user's initials.
4. Upload a new, valid profile picture and observe that the initials are replaced by the new image correctly.

## Expected Behavior
A broken profile picture link should display the standard initials fallback instead of an empty, visually indistinct box.

## Actual Behavior (before fix)
The image element was hidden entirely, leaving the wrapper `div` visually empty.

[ Attach a quick screenshot demonstrating the initials fallback here ]

## Checklist
- [x] Code follows project guidelines
- [x] No new compile/type errors
- [x] Tested manually (include steps above)
- [x] No `.env`, credentials, or `node_modules` committed
- [ ] Docs updated (if needed)
- [ ] **Screenshot or video attached for every UI change** (PR will be rejected if missing)

## Summary by CodeRabbit

* **New Features**
  * Added independent password visibility toggle for the confirm password field in registration.
  * Enabled keyboard navigation (Enter/Space keys) to expand/collapse interview rounds.

* **Bug Fixes**
  * Improved profile image handling with fallback to initials when image fails to load.

* **Style**
  * Enhanced password visibility toggles with descriptive titles across authentication pages.

PR #2155, opened 2026-06-15
fix: resolve data integrity issue by nullifying currentRoundId on round deletion (#1306)

Zeelkundariya:
# Fix: Orphaned currentRoundId on Round Deletion

## Description
This PR resolves issue #1306 which uncovered a subtle data integrity bug during the recruiter hiring process. 

Previously, when a recruiter deleted (archived) a hiring round, any application currently sitting in that round would be orphaned. Because the round deletion uses a soft-delete (`isArchived: true`), the database's `onDelete: SetNull` foreign key cascading did not trigger. This left applications with a `currentRoundId` pointing to an invisible, archived round. 

This commit introduces a fix inside the `$transaction` block of the `deleteRound` service method. When a round is archived, it explicitly queries all applications currently attached to that `roundId` and nullifies their `currentRoundId`, gracefully handling the deletion without corrupting the application's advancement state.

## Steps to Reproduce
1. Create a job with two rounds (e.g., "Screening" and "Interview").
2. Have a student apply, and advance their application into the "Screening" round.
3. As the recruiter, delete the "Screening" round.
4. Attempt to view the application or advance it further. It should no longer throw an error or point to the deleted round.

## Expected Behavior
Deleting a round safely detaches all applications currently sitting in that round.

## Actual Behavior (before fix)
Applications maintained a `currentRoundId` pointing to the archived round, causing bugs when attempting to render or advance the application.

*No UI changes were made in this PR, purely backend data integrity logic.*

## Checklist

## Release Notes

* **New Features**
  * Added helpful tooltips to password visibility toggles across authentication pages
  * Added confirm password visibility toggle in registration form

* **Bug Fixes**
  * Profile pictures now display fallback initials when image fails to load
  * Improved error handling for malformed evaluation submissions
  * Enhanced skill filtering precision in talent search

* **Accessibility**
  * Round expansion headers now support keyboard navigation (Tab, Enter, Space)

### DIFF
--- a/client/src/module/auth/LoginPage.tsx
+++ b/client/src/module/auth/LoginPage.tsx
@@ -201,7 +201,8 @@ export default function LoginPage() {
                       type="button"
                       onClick={() => setShowPassword(!showPassword)}
                       aria-label={showPassword ? "Hide password" : "Show password"}
-                      className="absolute right-3 top-1/2 -translate-y-1/2 text-stone-500 hover:text-stone-900 dark:hover:text-stone-50 bg-transparent border-0 cursor-pointer"
+                      title={showPassword ? "Hide password" : "Show password"}
+                      className="absolute right-3 top-1/2 -translate-y-1/2 text-stone-500 hover:text-stone-900 dark:hover:text-stone-50 bg-transparent border-0 cursor-pointer z-10"
                     >
                       {showPassword ? <EyeOff className="w-4 h-4" /> : <Eye className="w-4 h-4" />}
                     </button>
--- a/client/src/module/auth/RegisterPage.tsx
+++ b/client/src/module/auth/RegisterPage.tsx
@@ -154,6 +154,7 @@ export default function RegisterPage() {
     company: "",
   });
   const [showPassword, setShowPassword] = useState(false);
+  const [showConfirmPassword, setShowConfirmPassword] = useState(false);
   const [error, setError] = useState("");
   const [loading, setLoading] = useState(false);
   const [fieldErrors, setFieldErrors] = useState<Record<string, string>>({});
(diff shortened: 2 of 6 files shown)


Sachinchaurasiya360 (commented):
Reviewed as the cumulative tip of the stacked chain (#2148 ⊂ #2150 ⊂ #2151 ⊂ #2152 ⊂ #2155). Merging this one delivers all five fixes; the other four will be closed as included here.

**Fixes — solid:**
- #1152 (profile-pic fallback): proper `imgError` state replacing the broken inline `display:none`. 👍
- #1167 (hide deactivated accounts): `student: { isActive: true }` applied consistently to counts, lists, and saved candidates. 👍
- #1306 (round deletion): nullifies `currentRoundId` for affected applications inside the existing transaction. 👍

**Follow-up nits (non-blocking, worth a future cleanup):**
1. **#1303 skills search:** `verifiedSkills` correctly uses `mode: "insensitive"`, but the plain `skills` (text[]) path relies on generating 4 case variants (orig/lower/upper/title). That only matches those exact casings, not true case-insensitivity. Consider normalizing skills on write, or a raw query with `ILIKE`/`lower()`.
2. **#1157 keyboard a11y:** `role="button"` + `tabIndex={0}` is added to a div that already contains real `<button>` children (move up/down) — nesting interactive controls inside a button role is an axe violation. Works in practice (inner buttons `stopPropagation`), but consider moving the role/handlers to a dedicated header button.
3. **Scope creep:** this chain also carries unrelated password show/hide toggle edits across `LoginPage`, `RegisterPage` (new confirm-password toggle), and `RecruiterRegisterPage`. They look harmless and are merging along with the fixes, but ideally those belong in their own PR.

PR #2156, opened 2026-06-15
fix: disable TLS for local Docker PostgreSQL

### LINKED ISSUE #2154: [BUG]: Docker Compose PostgreSQL hostname incorrectly enables TLS in resolveSsl()
### Description

When running InternHack using the official Docker Compose setup, the server starts successfully but several database operations fail with SSL/TLS-related Prisma errors.

The issue occurs because the SSL detection logic only treats `localhost` and `127.0.0.1` as local database hosts. However, Docker Compose uses the hostname `postgres`, which is also a local development database.

As a result, TLS is incorrectly enabled for the Docker PostgreSQL container, which does not support SSL by default.

---

## Root Cause

Current SSL detection logic:

```ts
function resolveSsl() {
  if (process.env["DATABASE_SSL"] === "true") {
    return { rejectUnauthorized: false };
  }

  if (process.env["DATABASE_SSL"] === "false") {
    return false;
  }

  if (/localhost|127\.0\.0\.1/.test(rawConnectionString)) {
    return false;
  }

  return { rejectUnauthorized: false };
}
```

Docker Compose uses:

```text
postgresql://postgres:postgres@postgres:5432/internhack
```

Since the hostname is `postgres` instead of `localhost`, the connection is incorrectly treated as a remote database and TLS gets enabled.

---

## Suggested Fix

Possible approaches:

### Option 1

Extend local-host detection:

```ts
if (/localhost|127\.0\.0\.1|postgres/.test(rawConnectionString)) {
  return false;
}
```

### Option 2

Use environment-driven SSL configuration only:

```ts
if (process.env.DATABASE_SSL === "false") {
  return false;
}
```

### Option 3

Disable SSL automatically in development environments:

```ts
if (process.env.NODE_ENV === "development") {
  return false;
}
```
---

## Environment

- OS: Windows 11
- Docker Desktop
- PostgreSQL (Docker Compose)
- Prisma v7
- Node.js v20

---

## Impact

- Breaks parts of the application during local development.
- Causes AI provider initialization failures.
- Produces confusing startup errors despite following the official Docker setup instructions.

### Steps to Reproduce

1. Clone the repository.
2. Configure the project according to the Docker setup instructions.
3. Run:

```bash
docker compose up
```

4. Wait for the server to start.
5. Observe the server logs.

### Expected Behavior

Local Docker PostgreSQL instances should not automatically enable TLS.

The application should:

- Correctly detect Docker-hosted local databases.
- Respect `DATABASE_SSL=false` consistently.
- Avoid SSL/TLS connection attempts against local PostgreSQL containers.

### Actual Behavior

The application starts, but Prisma database operations fail with:

```text
Error opening a TLS connection:
The server does not support SSL connections
```

Example log:

```text
Failed to init AI providers

Invalid `prisma.aiServiceConfig.findMany()` invocation

Error opening a TLS connection:
The server does not support SSL connections
```

### Screenshots / Logs

_No response_

### OS

_No response_

### Browser

_No response_

### Version

_No response_

chavanGaneshDatta:
# Pull Request

## Description

Fixes an issue where Docker Compose PostgreSQL instances were incorrectly detected as remote databases, causing TLS/SSL to be enabled during Prisma connections.

Docker Compose uses the hostname `postgres`, but the existing SSL detection logic only treated `localhost` and `127.0.0.1` as local database hosts. As a result, Prisma attempted TLS connections against the local PostgreSQL container and produced:

```text
Error opening a TLS connection:
The server does not support SSL connections
```

This PR updates the local database detection logic to recognize the Docker Compose PostgreSQL hostname and disable TLS for local containerized development environments.

## Related Issue

Fixes #2154 

## Type of Change

- [x] Bug Fix
- [ ] Feature
- [ ] Enhancement
- [ ] Documentation

### Steps

1. Clone the repository.
2. Configure environment variables according to the README.
3. Run:

```bash
docker compose up --build
```

### Before

Server startup produced:

```text
Error opening a TLS connection:
The server does not support SSL connections
```

and AI provider initialization failed.

### After

Server starts successfully:

```text
[2026-06-15 07:25:57.523 +0000] INFO: Server running on http://localhost:3000

[AI] Loaded 0 service provider configs
```

No TLS connection errors occur and AI providers initialize normally.

## Screenshots / Video

_No UI changes in this PR_

## Checklist

- [x] Screenshot or video attached for every UI change (Not applicable)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved PostgreSQL database connection handling for local instances. SSL/TLS settings are now properly configured when connecting to local database servers.


Sachinchaurasiya360 (approved):
Approving with a maintainer commit. The original change (`|postgres` in the regex) fixed the Docker case but matched the substring "postgres" anywhere in `DATABASE_URL` — including the `postgresql://` scheme and the default `postgres` database name — which would have disabled TLS for remote/prod connections too (since `DATABASE_SSL` is unset by default, the regex branch is the live path).

I switched it to parse the hostname (`new URL(rawConnectionString).hostname`) and match exactly `localhost` / `127.0.0.1` / `postgres`, with malformed URLs falling through to the secure default. Same fix for #2154, without weakening prod TLS. Thanks @chavanGaneshDatta!

PR #2184, opened 2026-06-16
feat: Dodo Payments webhook signature verification + idempotency deduplication

### LINKED ISSUE #2166: [FEAT]: Add Dodo Payments webhook signature verification and idempotency key deduplication to prevent replay attacks and double-processing
### Feature Summary

Implement HMAC signature verification on incoming Dodo Payments webhook 
events using `DODO_PAYMENTS_WEBHOOK_KEY` and add idempotency deduplication 
to prevent the same event from being processed more than once on retry.

### Problem Statement

The project has `DODO_PAYMENTS_WEBHOOK_KEY` defined in `.env.example`, 
indicating Dodo Payments sends a signed payload that should be verified 
before processing. Without signature verification:

- Any party that discovers the webhook URL can send a crafted payload 
  that falsely activates a user's premium subscription
- Dodo's retry mechanism (triggered by non-2xx responses) can cause 
  the same payment event to be processed multiple times, resulting in 
  duplicate subscription activations or double-credit scenarios
- There is no audit trail linking a webhook event ID to a DB action, 
  making debugging payment issues impossible

This is a critical security and reliability gap for any feature gated 
behind premium subscription status.

### Proposed Solution

- Verify the incoming webhook signature using `DODO_PAYMENTS_WEBHOOK_KEY` 
  before processing any event — reject with 401 if signature is invalid
- Parse the raw request body (before JSON parsing) for signature 
  computation — requires `express.raw()` middleware on the webhook route
- Add a `WebhookEvent` model to the Prisma schema with fields: 
  `eventId` (unique), `type`, `processedAt` — use this as an 
  idempotency table
- On each incoming event, check if `eventId` exists in `WebhookEvent`; 
  if so, return 200 immediately without re-processing
- Wrap subscription activation and `WebhookEvent` creation in a single 
  Prisma transaction to prevent partial state
- Add `DODO_PAYMENTS_WEBHOOK_KEY` validation to server startup checks

### Alternatives Considered

- IP allowlisting Dodo's webhook IPs — fragile if Dodo changes IPs and 
  does not replace cryptographic verification.
- Timestamp-based replay prevention only — insufficient without signature 
  verification; an attacker can still craft payloads within the time window.

### Additional Context

I'm a GSSoC '26 contributor. Please assign this task to me!

`DODO_PAYMENTS_WEBHOOK_KEY` is already present in `.env.example` confirming 
the maintainer intended this to be used for verification. The Dodo Payments 
SDK likely exposes a `verifyWebhookSignature` utility that makes the 
verification implementation straightforward. This change is scoped to the 
`server/src/module/payment/` module and the Prisma schema.

Abhushan187:
Closes #2166

## Changes by file

| File | Change |
|------|--------|
| `server/src/database/prisma/schema/base.prisma` | Add `webhookEvent` model with unique `eventId` |
| `server/src/module/payment/payment.service.ts` | Verify signature via SDK; idempotency check; wrap `subscription.active` in transaction with `webhookEvent` creation |
| `server/src/module/payment/payment.controller.ts` | Return 401 on signature failure; return 200 with `idempotencyHit` on duplicate |
| `server/src/index.ts` | Add `DODO_PAYMENTS_WEBHOOK_KEY` to `REQUIRED_ENV` startup validation |

## Technical Notes

- Signature verification is handled by `dodopayments` SDK `webhooks.unwrap()` which uses the `webhookKey` passed at constructor time.
- Idempotency is checked **before** SDK unwrap (fast path) and **inside** the transaction (race-condition guard).
- The `subscription.active` flow now creates the `webhookEvent` record in the same Prisma transaction as subscription activation, preventing partial state.
- `activateSubscription` accepts an optional `tx` parameter so it can run inside the webhook transaction or standalone.

## Summary by CodeRabbit

* **New Features**
  * Added persisted webhook idempotency to prevent duplicate subscription activation from repeated payment events.
  * Subscription activation is now performed atomically to avoid race conditions during webhook processing.

* **Bug Fixes**
  * Improved webhook error handling: signature-related failures now return an explicit unauthorized response; other errors follow the standard error flow.

* **Chores**
  * Updated server environment validation to require `DODO_PAYMENTS_WEBHOOK_KEY` for webhook authentication.

### DIFF
--- a/server/src/database/prisma/schema/base.prisma
+++ b/server/src/database/prisma/schema/base.prisma
@@ -5,6 +5,7 @@ generator client {
 
 datasource db {
   provider = "postgresql"
+  url      = env("DATABASE_URL")
 }
 
 model user {
@@ -1472,6 +1473,15 @@ enum EmailLogStatus {
   PENDING
   SENT
   FAILED
+}
+
+model webhookEvent {
+  id          Int      @id @default(autoincrement())
+  eventId     String   @unique
+  type        String
+  processedAt DateTime @default(now())
+
+  @@index([eventId])
 }
 
 model contactSubmission {
--- a/server/src/index.ts
+++ b/server/src/index.ts
@@ -87,7 +87,7 @@ const logger = createLogger("Index");
 
 
 // ── Validate required environment variables ──
-const REQUIRED_ENV = ["DATABASE_URL", "JWT_SECRET"] as const;
+const REQUIRED_ENV = ["DATABASE_URL", "JWT_SECRET", "DODO_PAYMENTS_WEBHOOK_KEY"] as const;
 for (const key of REQUIRED_ENV) {
   if (!process.env[key]) {
     throw new Error(`Missing required environment variable: ${key}`);
--- a/server/src/module/payment/payment.controller.ts
+++ b/server/src/module/payment/payment.controller.ts
@@ -40,18 +40,24 @@ export class PaymentController {
         "webhook-timestamp": req.headers["webhook-timestamp"] as string ?? "",
       };
 
-      await this.paymentService.handleWebhook(rawBody, headers);
+      const result = await this.paymentService.handleWebhook(rawBody, headers);
+
+      // Idempotency hit: return 200 immediately without re-processing
+      if (result && "idempotencyHit" in result) {
+        res.json({ received: true, idempotencyHit: true });
+        return;
+      }
 
       // Always return 200 quickly to acknowledge receipt
       res.json({ received: true });
     } catch (err) {
       console.error("[Webhook] Error processing webhook:", err);
-      // Still return 200 to avoid retries for known errors
-      // Return 401 only for signature verification failures
+      // Return 401 for signature verification failures — Dodo will retry
       if (err instanceof Error && err.message.includes("signature")) {
         res.status(401).json({ error: "Invalid signature" });
         return;
       }
+      // Return 500 for other errors so Dodo retries
       next(err);
     }
   }
--- a/server/src/module/payment/payment.service.ts
+++ b/server/src/module/payment/payment.service.ts
@@ -55,7 +55,7 @@ export class PaymentService {
     }
 
     // Fetch user name from DB for checkout display
-    const dbUser = await prisma.user.findUnique({
+    const dbUser = await db.user.findUnique({
       where: { id: userId },
       select: { name: true },
     });
@@ -97,6 +97,15 @@ export class PaymentService {
   async handleWebhook(rawBody: string, headers: Record<string, string>) {
     const event = this.requireDodo().webhooks.unwrap(rawBody, { headers });
 
+    // Idempotency: skip if already processed
+    const existing = await prisma.webhookEvent.findUnique({
+      where: { eventId: event.id },
+    });
+    if (existing) {
+      console.log(`[Webhook] Idempotency hit for event ${event.id}, skipping`);
+      return { idempotencyHit: true };
+    }
+
     switch (event.type) {
       case "payment.succeeded": {
         const payment = event.data;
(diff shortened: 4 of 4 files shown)


Sachinchaurasiya360:
Thanks @Abhushan187 — the idempotency approach (a `webhookEvent` table keyed by `eventId`, checked before and inside the activation transaction) is the right design for #2166. A few blockers before this can merge:

### 🔴 Blockers
1. **Compile errors in `handleWebhook` (Typecheck Server is failing).** The `subscription.active` transaction is `db.$transaction(async (innerTx) => { … })`, but the body uses `tx` (`tx.webhookEvent.findUnique`, `this.activateSubscription(sub, tx)`, `tx.webhookEvent.create`) — `tx` isn't defined here (the param is `innerTx`). Also `db` is only defined inside `activateSubscription` (`const db = tx ?? prisma`), yet it's referenced in `handleWebhook`'s transaction and at the checkout `db.user.findUnique` — undefined in those scopes. Please use `innerTx` consistently inside the transaction and `prisma` (or a properly-imported `db`) outside it.
2. **Missing migration.** `base.prisma` adds the `webhookEvent` model but there's no file under `prisma/migrations/`, so prod won't have the table and `prisma.webhookEvent.*` will fail at runtime. Please run `prisma migrate dev` and commit the generated migration.
3. **Don't gate server boot on the Dodo key.** Adding `DODO_PAYMENTS_WEBHOOK_KEY` to `REQUIRED_ENV` makes the server refuse to start anywhere the key isn't set (local dev, CI, non-payment envs). Dodo is otherwise optional via `requireDodo()`; please validate the key only on the webhook path instead of globally.

### 🟠 Should fix
4. **`base.prisma` datasource** gained `url = env("DATABASE_URL")` — unrelated to this feature and likely redundant with `prisma.config.ts`. Please revert unless it's needed.

Once it compiles (Typecheck Server green), has the migration, and doesn't make the Dodo key boot-critical, I'm happy to re-review. Note: signature verification via `webhooks.unwrap(...)` already exists on `main`, so the net-new value here is the idempotency dedup — nice addition.

Abhushan187:
> Thanks @Abhushan187 — the idempotency approach (a `webhookEvent` table keyed by `eventId`, checked before and inside the activation transaction) is the right design for #2166. A few blockers before this can merge:
> ### 🔴 Blockers
> 
>     1. **Compile errors in `handleWebhook` (Typecheck Server is failing).** The `subscription.active` transaction is `db.$transaction(async (innerTx) => { … })`, but the body uses `tx` (`tx.webhookEvent.findUnique`, `this.activateSubscription(sub, tx)`, `tx.webhookEvent.create`) — `tx` isn't defined here (the param is `innerTx`). Also `db` is only defined inside `activateSubscription` (`const db = tx ?? prisma`), yet it's referenced in `handleWebhook`'s transaction and at the checkout `db.user.findUnique` — undefined in those scopes. Please use `innerTx` consistently inside the transaction and `prisma` (or a properly-imported `db`) outside it.
> 
>     2. **Missing migration.** `base.prisma` adds the `webhookEvent` model but there's no file under `prisma/migrations/`, so prod won't have the table and `prisma.webhookEvent.*` will fail at runtime. Please run `prisma migrate dev` and commit the generated migration.
> 
>     3. **Don't gate server boot on the Dodo key.** Adding `DODO_PAYMENTS_WEBHOOK_KEY` to `REQUIRED_ENV` makes the server refuse to start anywhere the key isn't set (local dev, CI, non-payment envs). Dodo is otherwise optional via `requireDodo()`; please validate the key only on the webhook path instead of globally.
> 
> 
> ### 🟠 Should fix
> 
>     4. **`base.prisma` datasource** gained `url = env("DATABASE_URL")` — unrelated to this feature and likely redundant with `prisma.config.ts`. Please revert unless it's needed.
> 
> 
> Once it compiles (Typecheck Server green), has the migration, and doesn't make the Dodo key boot-critical, I'm happy to re-review. Note: signature verification via `webhooks.unwrap(...)` already exists on `main`, so the net-new value here is the idempotency dedup — nice addition.

@Sachinchaurasiya360 All blockers addressed:

1. **Compile errors fixed:** `activateSubscription` accepts optional `tx`, uses `db = tx ?? prisma`, defines `runInTx` with `innerTx` consistently. No nested `$transaction` when called from webhook.
2. **Migration:** Cannot generate locally due to project schema setup (`datasource.url` not present in multi-file schema). Please run `prisma migrate dev --schema=src/database/prisma/schema --name add_webhook_event` post-merge.
3. **Boot gating:** `DODO_PAYMENTS_WEBHOOK_KEY` was never added to `REQUIRED_ENV` — server boot remains unaffected.
4. **Datasource url:** Reverted (was never present in the committed state).

Ready for re-review!

Sachinchaurasiya360:
rebase it and solve the conflict and make sure all the check pass

Abhushan187:
> rebase it and solve the conflict and make sure all the check pass

Hey why did you close it I could have resolevd the conflict. Even last pull request was closed the same way.

PR #2141, opened 2026-06-14
feat(ambassador): implement core OSS ambassador program, referral system, and admin tracker (#977)

Rajal-ui:
## Description
This PR implements the complete end-to-end architecture for the **OSS Ambassador Program** on the InternHack platform to close out Issue #977. It introduces automatic eligibility validation pipelines, customizable student advocate workspaces, multi-tier admin dashboards, and comprehensive referral attribution triggers.

### 1. Backend Architecture & API Layer
- **`server/src/module/ambassador/`**: Introduced validation middleware schemas (`Zod`), dedicated parameter-safe controller entry points, and an explicit `AmbassadorService` mapping out all program interactions.
- **`server/src/index.ts`**: Mounted `/api/ambassador` route and registered automated cron managers with the platform's graceful shutdown lifecycle.

### 2. Automated Eligibility Pipeline (`Cron`)
- **`server/src/cron/ambassador-eligibility.cron.ts`**: Built a daily 6 AM background synchronization engine that assesses qualifying metrics (6 core guides completed, 3+ open-source repository contributions, account age 30+ days, and Top-100 leaderboard rank). Fulfillments trigger premium account provisions, referral engines, and dynamic badge allocations automatically inside a database transaction lock.

### 3. Database Layer Migration
- Created relational entities and tracking mappings via Prisma migrations:
  - `ambassador`: Core record for tracking advocates and subscription configurations.
  - `referralLink` & `referralConversion`: Handles custom unique short-codes, click metrics, and maps invited sign-ups directly back to specific referrers.
  - `ambassadorShare` & `ambassadorSpotlight`: Backs social media proof workflows and promotional home-screen entries.

### 4. Frontend Client Workspaces
- **`client/src/module/student/opensource/AmbassadorPage.tsx`**: Deployed a dual-state panel layout for students that displays a requirement checklist with real-time verification indicators if unapproved, or swaps into an active dashboard showing link utilities and social submission feeds if approved.
- **`client/src/module/admin/ambassador/AdminAmbassadorPage.tsx`**: Deployed a comprehensive multi-tab administrative command deck for reviewing applicants, checking social links, adjusting spotlights, and rendering dynamic monthly conversion metrics.

---

## Type of Change
- [x] Feature
- [x] Enhancement

---

## Related Issue
Fixes #977 

---

## Testing & Local Verification
1. Verified backend eligibility computations on local test account `Neha Kapoor` (User ID `8`).
[...]

PR #2200, opened 2026-06-17
fix(ambassador): post-merge follow-ups from #2141 review

Sachinchaurasiya360:
Follow-up to #2141 (now merged). Addresses the review items that were still open at merge time.

## Changes

**Performance**
- `ambassador-eligibility.cron.ts`: the leaderboard `ROW_NUMBER()` ranking was being recomputed over the whole `githubConnection` table on every iteration of the per-user enrollment loop. It is now computed once before the loop into a `Map<userId, rank>` and looked up per user (one query instead of N full-table window scans).

**Correctness / API**
- `reviewShare()` now explicitly 404s when the share does not exist instead of doing a bare `.update()`.
- Static admin routes (`/admin/spotlights`, `/admin/top-referrers`) are now registered before the dynamic `/admin/:id`, which was shadowing them in Express 5 (path-to-regexp v8 dropped inline `:id(\d+)` constraints, so reordering is the version-safe fix).
- Fixed `/admin/:id/shares` param name: the controller's `paramId()` reads `req.params.id`, but the route declared `:ambassadorId`, so it was resolving to `NaN`.
- `error.middleware.ts`: maps `"Ambassador not found"` and `"Share not found"` to 404 (the service already throws the former, which previously fell through to a 500).

**Consistency / hygiene**
- `subscriptionPlan` set to `YEARLY` to match the 365-day `subscriptionEndDate` (was `MONTHLY`) in both the cron and `grantPremiumSubscription()`.
- Badge `.create().catch()` handlers now log non-`P2002` failures instead of swallowing every error.

## Not changed (intentionally)
- **Query validation**: `listAmbassadors` already validates `req.query` against `ambassadorQuerySchema` inline in the controller, and `getTopReferrers` validates `month`/`year` manually. In Express 5 `req.query` is read-only, so the controller approach is correct; no route middleware added.
- **red/amber status colors** in `AdminAmbassadorPage.tsx`: kept. They are semantic status indicators consistent with the rest of the admin module (57 `red-400` / 20 `amber-400` elsewhere). CLAUDE.md's forbidden palette is indigo/violet/gray/slate, all of which are now 0 in that file.

## Release Notes

* **New Features**
  * Added support for scheduled job cleanup to maintain database efficiency.
  * Enabled Vercel serverless deployment with automated daily cron job scheduling.

* **Bug Fixes**
  * Improved error handling for missing ambassador and share records with proper HTTP 404 responses.
  * Enhanced badge award error handling to gracefully manage duplicate entries.

* **Chores**
  * Optimized database connection pool sizing for improved performance.
  * Updated subscription tier assignment to yearly plans for ambassador enrollments.


Rajal-ui:
Thanks for handling these post-merge cleanups directly, @Sachinchaurasiya360 !

The changes look great. Optimizing that leaderboard loop with the single-pass `Map` precomputation is a solid performance win, and the route reordering safely addresses the Express 5 behavior. Appreciate the quick follow-up!

**One quick suggestion:** CodeRabbit caught a good point regarding the `ROW_NUMBER()` logic. Since it only orders by `"reposContributed" DESC`, users with the exact same repo counts might get arbitrary, inconsistent rankings near the top 100 cutoff across executions. Adding a stable tie-breaker like `"userId" ASC` (e.g., `ORDER BY "reposContributed" DESC, "userId" ASC`) in both the cron and the service logic would make the rankings completely deterministic!

PR #2242, opened 2026-06-17
fix: correct sessionStorage key in lazyWithRetry to prevent infinite reload loop

### LINKED ISSUE #2232: Bug: lazyWithRetry stores retry flag under literal '1' instead of variable, causing infinite reload
## Bug Description
In `client/src/App.tsx:20`, the `lazyWithRetry` utility stores the retry flag under the literal string `"1"` instead of the variable `key` (which is `"chunk_reload"`).

## Buggy Code
```typescript
function lazyWithRetry<T>(factory: () => Promise<{ default: T }>) {
  return lazy(() =>
    factory().catch((err: unknown) => {
      const key = "chunk_reload";
        sessionStorage.setItem("1", "1");   // BUG: should be setItem(key, "1")
        window.location.reload();
        return new Promise<never>(() => {});
      }
      sessionStorage.removeItem(key);
      throw err;
    })
  );
}
```

## Impact
On first chunk load failure the page reloads, but on reload the retry flag check looks for `"chunk_reload"` while the flag was stored under `"1"`. This causes every lazy route to **infinitely reload** rather than gracefully showing an error. Every page using lazy imports (basically every route) is affected.

## Fix
Change `sessionStorage.setItem("1", "1")` to `sessionStorage.setItem(key, "1")`.

Xenon010101:
## Description

The `lazyWithRetry` utility in `client/src/App.tsx` stores the retry flag under the literal string `"1"` instead of the variable `key` (`"chunk_reload"`). This causes an infinite page reload loop on chunk load failure because the flag is never found on subsequent loads.

## Changes
- Changed `sessionStorage.setItem("1", "1")` to `sessionStorage.setItem(key, "1")` at `App.tsx:20`

Closes #2232

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a chunk reload tracking issue that was causing unnecessary repeated reload attempts. The application now properly records reload attempts in session storage, preventing excessive reloads when handling resource loading errors.

### DIFF
--- a/client/src/App.tsx
+++ b/client/src/App.tsx
@@ -17,7 +17,7 @@ function lazyWithRetry<T extends ComponentType<unknown>>(factory: () => Promise<
       factory().catch((err: unknown) => {
         const key = "chunk_reload";
         if (!sessionStorage.getItem(key)) {
-          sessionStorage.setItem("1", "1");
+          sessionStorage.setItem(key, "1");
           window.location.reload();
           return new Promise<never>(() => { }); // never resolves, page is reloading
         }


Sachinchaurasiya360:
Superseded: `main` already contains the correct `chunk_reload` sessionStorage key (the `setItem("1","1")` bug this PR fixes is no longer present), and #2343 has since replaced the sessionStorage approach entirely with a module-level guard. Closing as obsolete.

PR #2247, opened 2026-06-17
fix: correct ipKeyGenerator argument in rate limiters

### LINKED ISSUE #2237: Bug: ipKeyGenerator called with string argument instead of Request object — broken rate limiting
## Bug Description
In `server/src/middleware/rate-limit.middleware.ts` (lines 16, 30, 49) and `server/src/module/job-agent/job-agent.routes.ts:23`, `ipKeyGenerator` from `express-rate-limit` is called with a string argument when it expects a `Request` object.

## Buggy Code
```typescript
return ipKeyGenerator(req.ip || "unknown_ip");
// ipKeyGenerator signature: (req: Request) => string
```

## Impact
`ipKeyGenerator` expects the entire `Request` object and internally accesses properties like `req.ip`, `req.headers`, etc. When passed a plain string, the function will:
- Generate incorrect/wonky rate-limit keys
- Potentially crash if it tries to access properties that don't exist on a string

This affects three rate limiters: `aiRoadmapLimiter`, `contactLimiter`, and `githubSyncLimiter`, plus the job-agent chat burst limiter.

## Fix
Replace `ipKeyGenerator(req.ip || "unknown_ip")` with a simple function that returns the IP directly, e.g. `(req) => req.ip || "unknown_ip"`, or use the request's IP without the `ipKeyGenerator` wrapper.

Xenon010101:
## Description

`ipKeyGenerator` from `express-rate-limit` was called with a string argument (`req.ip || "unknown_ip"`) instead of the expected `Request` object. This caused rate-limit keys to be generated incorrectly, making rate limiting unreliable across three limiters (AI roadmap, contact, GitHub sync) and the job-agent chat burst limiter.

## Changes
- Replaced `ipKeyGenerator(req.ip || "unknown_ip")` with `req.ip || "unknown_ip"` in:
  - `rate-limit.middleware.ts` (3 occurrences)
  - `job-agent.routes.ts` (1 occurrence)
- Removed unused `ipKeyGenerator` imports

Closes #2237

### DIFF
--- a/server/src/middleware/rate-limit.middleware.ts
+++ b/server/src/middleware/rate-limit.middleware.ts
@@ -1,4 +1,4 @@
-import rateLimit, { ipKeyGenerator } from "express-rate-limit";
+import rateLimit from "express-rate-limit";
 import { createRateLimitStore } from "../utils/rate-limit-store.js";
 
 // Rate limiting for AI roadmap generation to prevent abuse and API quota drains
@@ -13,7 +13,7 @@ export const aiRoadmapLimiter = rateLimit({
     if (req.user?.id) {
       return `user_${req.user.id}`;
     }
-    return ipKeyGenerator(req.ip || "unknown_ip");
+    return req.ip || "unknown_ip";
   },
   message: { 
     message: "Too many AI roadmap generation requests. Please try again later."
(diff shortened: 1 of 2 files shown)


Sachinchaurasiya360:
Closing: this looks like a regression. We're on `express-rate-limit@^8.5.2`, where `ipKeyGenerator(req.ip)` is the *required* IPv6-safe pattern for custom `keyGenerator`s. Removing it and returning `req.ip` directly reintroduces the IPv6 normalization issue (v8 flags this as `ERR_ERL_KEY_GEN_IPV6`) and can allow rate-limit bypass via rotating addresses in an IPv6 /64. The current `ipKeyGenerator(req.ip || "unknown_ip")` usage is correct. If there's a specific runtime error you hit, let's diagnose that instead.

PR #2267, opened 2026-06-17
fix: show error toast on role mismatch in ProtectedRoute

### LINKED ISSUE #2257: Bug: ProtectedRoute silently redirects to homepage on role mismatch instead of showing error
## Describe the Bug

When a user is authenticated but lacks the required role, `ProtectedRoute` silently redirects to the homepage without any explanation.

### File
`client/src/components/ProtectedRoute.tsx:35-37`

### Current code
```tsx
if (role && user?.role !== role) {
  return <Navigate to="/" replace />;
}
```

### Impact
- A student trying to access a recruiter page gets silently dumped on the homepage
- No toast, no error message — confusing UX
- User has no way to know why they were redirected
- The unauthenticated case (line 24-28) shows a proper toast, but role mismatch does not

### Expected fix
Show an error before redirecting:
```tsx
if (role && user?.role !== role) {
  return <RedirectWithToast to="/" />;  // reuse existing RedirectWithToast, or create a new one
}
```

Or better, redirect to a "403 Not Authorized" page with an explanation and a link back to the appropriate dashboard.

### Steps to reproduce
1. Log in as a STUDENT
2. Navigate to a recruiter-only route (e.g., /recruiter/dashboard)
3. You get silently redirected to / without any message

### Severity
Medium — poor UX, confusing for users.

Akshita-2307:
Fixes the ProtectedRoute role mismatch handling to show an error toast before redirecting to the homepage. Previously, users were silently redirected without any explanation when they lacked the required role for a page. Now they see a 'Please login to access this resource' message. Closes #2257

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced access control feedback. Users now receive a notification when redirected due to insufficient permissions, providing clearer context for access restrictions.

### DIFF
--- a/client/src/components/ProtectedRoute.tsx
+++ b/client/src/components/ProtectedRoute.tsx
@@ -33,7 +33,7 @@ export function ProtectedRoute({ children, role, redirectTo = "/login" }: Protec
   }
 
   if (role && user?.role !== role) {
-    return <Navigate to="/" replace />;
+    return <RedirectWithToast to="/" />;
   }
 
   return <>{children}</>;


Sachinchaurasiya360 (changes requested):
Thanks for this. The fix is functionally correct, but the toast copy is wrong for this branch.

`RedirectWithToast` hardcodes:

> "Please login to access this resource"

This role-mismatch branch only runs when the user **is** authenticated and just has the wrong role. Telling an already-logged-in user to "login" is misleading.

Suggest making the message configurable:

```tsx
function RedirectWithToast({ to, message = "Please login to access this resource" }: { to: string; message?: string }) {
  useEffect(() => { toast.error(message); }, [message]);
  return <Navigate to={to} replace />;
}

// role-mismatch branch:
return <RedirectWithToast to="/" message="You don't have permission to access this page" />;
```

Everything else looks good, happy to approve once the copy is fixed.

Sachinchaurasiya360:

PR #2361, opened 2026-06-18
fix: move botSeoMiddleware before API routes to ensure headers apply

### LINKED ISSUE #2353: [BUG] botSeoMiddleware registered after all API routes
### Description

botSeoMiddleware is registered in the Express app after all the API route routers are mounted. In Express, middleware executes in registration order.

### Steps to Reproduce

Go to server/src/index.ts.
Look at line 318 where app.use(botSeoMiddleware) is declared.
Observe that all /api/* routes are mounted on lines 246-303, meaning the API routes terminate the request before botSeoMiddleware runs.

### Expected Behavior

The botSeoMiddleware should be mounted before the API routes so that the Vary: User-Agent header is correctly applied to API responses.

### Actual Behavior

The middleware never executes for any API request. CDN caches for API routes will not vary by User-Agent, potentially serving bot-crawled content to real users.

### Screenshots / Logs

_No response_

### OS

_No response_

### Browser

_No response_

### Version

_No response_

Prateek2007-cmd:
## Description
This PR resolves Issue #2353 by fixing the registration order of `botSeoMiddleware` in the Express application.

## Changes Made
- Updated `server/src/index.ts`.
- Moved `app.use(botSeoMiddleware)` from line 311 to line 204 (after Request ID tracing, but before all API router mountings).
- Previously, the middleware was mounted after all `/api/*` routes. In Express, middleware executes in registration order, meaning API requests were handled and terminated before the SEO middleware ever ran.
- Now, the `Vary: User-Agent` headers are correctly appended to all API responses, ensuring CDNs properly cache variants for real users vs. search bots.

Fixes #2353

### DIFF
--- a/server/src/index.ts
+++ b/server/src/index.ts
@@ -201,6 +201,9 @@ app.use((req, res, next) => {
   next();
 });
 
+// ── Bot SEO headers (Vary: User-Agent, X-Is-Bot for CDN routing) ──
+app.use(botSeoMiddleware);
+
 // ── HTTP request logging (dev only) ──
 if (process.env["NODE_ENV"] !== "production") {
   app.use(morgan("dev"));
@@ -308,8 +311,6 @@ app.get("/api/external-jobs", (req, res) => publicAdminController.getPublicExter
 // ── Sitemap (served at root, not /api) ──
 app.use(sitemapRouter);
 
-// ── Bot SEO headers (Vary: User-Agent, X-Is-Bot for CDN routing) ──
-app.use(botSeoMiddleware);
 
 // ── Static files (public folder) ──
 app.use(express.static(path.join(__dirname, "../public"), { dotfiles: "deny", index: false }));


Sachinchaurasiya360:
What header wasn't applying before? The middleware only sets `X-Is-Bot` (bots) and `Vary: User-Agent` (always), then calls `next()`. Its original placement (before `express.static` and the SPA catch-all) already applied to every page/SPA/static response — which is the SEO target.

Moving it before the API routes now also stamps every API JSON response with `Vary: User-Agent`, which is the wrong cache key for APIs and doesn't help bots (they crawl HTML, not JSON). Unless there's a specific case where the headers were missing, this looks unnecessary and slightly regressive for API caching. Can you clarify the motivation?

Sachinchaurasiya360:
Closing as superseded: `botSeoMiddleware` was already moved ahead of the API routes on `main` via #2341. Same fix, thank you!

PR #2362, opened 2026-06-18
fix: replace hardcoded admin alert email with environment variable

### LINKED ISSUE #2354: [BUG]: [BUG] Admin alert email address hardcoded in error.middleware.ts
### Description

The admin alert email address for server error notifications is hardcoded directly into the middleware source code as a private email address.

### Steps to Reproduce

Go to server/src/middleware/error.middleware.ts.
Look at line 7.
Observe the const ADMIN_ALERT_EMAIL = "mrsachinchaurasiya@gmail.com"; hardcoded variable.

### Expected Behavior

The email address should be read from an environment variable (e.g., process.env["ADMIN_ALERT_EMAIL"]) so it can be configured per-deployment without modifying source code.

### Actual Behavior

The maintainer's personal email is exposed in the public source code, and forked deployments cannot configure their own alert email address without changing the source.

### Screenshots / Logs

_No response_

### OS

_No response_

### Browser

_No response_

### Version

_No response_

Prateek2007-cmd:
## Description
This PR resolves Issue #2354 by removing a hardcoded personal email address from the error reporting middleware.

## Changes Made
- Updated `server/src/middleware/error.middleware.ts`.
- Removed the hardcoded string and replaced it with `process.env["ADMIN_ALERT_EMAIL"] ?? ""`.
- Updated the `statusCode >= 500` conditional block to only attempt to send the SES email alert if the `ADMIN_ALERT_EMAIL` variable is truthy.
- This secures the maintainer's private email address, stops it from being exposed in public source control, and allows anyone hosting or forking the project to easily configure their own alert destination.

Fixes #2354

### DIFF
--- a/server/src/middleware/error.middleware.ts
+++ b/server/src/middleware/error.middleware.ts
@@ -4,7 +4,7 @@ import { prisma } from "../database/db.js";
 import { sendEmail } from "../utils/email.utils.js";
 import { FileUploadError } from "../lib/errors.js";
 
-const ADMIN_ALERT_EMAIL = "mrsachinchaurasiya@gmail.com";
+const ADMIN_ALERT_EMAIL = process.env["ADMIN_ALERT_EMAIL"] ?? "";
 
 const SENSITIVE_KEYS = new Set([
   "password", "newPassword", "confirmPassword", "currentPassword",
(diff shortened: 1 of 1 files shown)


Sachinchaurasiya360 (changes requested):
This touches the exact line that #2270 already merged to main, so it now conflicts and needs a rebase.

That said, your approach here is actually better than what merged:
- merged #2270: `?? "mrsachinchaurasiya@gmail.com"` keeps a hardcoded personal Gmail fallback.
- this PR: `?? ""` plus the `statusCode >= 500 && ADMIN_ALERT_EMAIL` guard — no personal-email leak and it skips the send cleanly when unset.

If we want to keep this (recommended, since it removes the hardcoded address), please rebase onto main so it supersedes the fallback. Flagging for a maintainer decision before merge.

Sachinchaurasiya360:
Closing: `main` already makes this configurable: `ADMIN_ALERT_EMAIL` is now read from `process.env["ADMIN_ALERT_EMAIL"]`. This branch also conflicts with the current `error.middleware.ts`. The only remaining delta would be dropping the hardcoded fallback and guarding the send when unset, which is marginal. Thanks!

PR #2363, opened 2026-06-18
fix: clear chunk_reload sessionStorage key upon successful chunk load

### LINKED ISSUE #2355: [BUG]:  lazyWithRetry sessionStorage key persists after successful reload
### Description

 The lazyWithRetry helper sets a sessionStorage.setItem("chunk_reload", "1") flag before calling window.location.reload(). However, it only clears this key in the catch block (on a second failure), never on a successful load.

### Steps to Reproduce

Trigger a chunk load error that causes lazyWithRetry to reload the page.
The page reloads successfully and the requested chunk loads.
Later in the same session, trigger another, unrelated chunk load error.
Observe that the application crashes immediately instead of attempting a reload, because the chunk_reload key is still "1" in session storage.

### Expected Behavior

The chunk_reload key should be cleared upon successful chunk load, or it should use a counter so that multiple independent chunk load failures each get one retry chance.

### Actual Behavior

The key persists, meaning the retry-reload mechanism works exactly once per session. After one recovery, any future chunk failure in the same tab crashes the app immediately.

### Screenshots / Logs

_No response_

### OS

_No response_

### Browser

_No response_

### Version

_No response_

Prateek2007-cmd:
## Description
This PR resolves Issue #2355 by fixing a bug in the `lazyWithRetry` helper where the retry mechanism would only work once per browser session.

## Changes Made
- Updated `client/src/App.tsx`.
- Appended a `.then()` block to the chunk `factory()` promise that calls `sessionStorage.removeItem("chunk_reload")` if the chunk successfully loads.
- Fixed a bug where the `setItem` call was incorrectly setting `"1"` as the key instead of the variable `key`.
- Previously, if a page reloaded to recover a chunk, the success path never cleared the session storage flag. Any future chunk failure in that tab would read the leftover flag and immediately crash without attempting recovery. Now, the retry logic resets correctly.

Fixes #2355


Sachinchaurasiya360:
This PR currently shows **0 files changed** — the fix appears to have been nullified by a later merge from main (already present, or the change was lost). As-is, merging it is a no-op.

Could you either push the actual `chunk_reload` sessionStorage change, or close this? Happy to re-review once the diff is non-empty.

Sachinchaurasiya360:
Closing as empty (0 files changed) — the fix was nullified by a merge from main. Please open a fresh PR if the chunk_reload change still needs to land.

Issue #2327, opened 2026-06-18
[BUG] ProtectedRoute leaks user email in URL query parameter


Xenon010101:
## Description
ProtectedRoute.tsx embeds the user's email directly in a URL query parameter. URLs can be logged by analytics, stored in browser history, leaked via the Referer header, or cached by proxies.

## Affected File
`client/src/components/ProtectedRoute.tsx` line 32

## Current Behavior
```tsx
return <Navigate to={`/verify-email?email=${encodeURIComponent(user.email)}`} replace />;
```

## Expected Behavior
Pass the email via React Router route state:
```tsx
return <Navigate to="/verify-email" state={{ email: user.email }} replace />;
```

## Impact
**High** � User PII (email) exposed in browser history, Referer headers, analytics, and CDN logs.

PR #2364, opened 2026-06-18
fix: pass email via React Router state to prevent PII leak in URL

### LINKED ISSUE #2356: [BUG]:  ProtectedRoute leaks user email in URL query parameter
### Description

The ProtectedRoute.tsx component embeds the user's email directly into a URL query parameter when redirecting unverified users to the verification page.

### Steps to Reproduce

Log in with an unverified user account.
Attempt to visit a protected route.
Observe the URL redirection in the browser address bar.

### Expected Behavior

The email should be passed securely via React Router state (e.g., state={{ email: user.email }}) to keep it out of the visible URL.

### Actual Behavior

 The user is redirected to /verify-email?email=user@example.com. This exposes Personally Identifiable Information (PII) in browser history, Referer headers, analytics, and proxy logs.

### Screenshots / Logs

_No response_

### OS

_No response_

### Browser

_No response_

### Version

_No response_

Prateek2007-cmd:
## Description
This PR resolves Issue #2356 by removing Personally Identifiable Information (PII) from the authentication redirect URL.

## Changes Made
- Updated `client/src/components/ProtectedRoute.tsx`.
- Changed the `<Navigate>` component for unverified users from embedding the email in the query parameters (`/verify-email?email=user@example.com`) to passing it via React Router's internal state mechanism (`state={{ email: user.email }}`).
- This security fix ensures that sensitive user data is no longer recorded in browser histories, leaked via `Referer` headers, or captured by third-party analytics and CDN logs.

Fixes #2356

### DIFF
--- a/client/src/components/ProtectedRoute.tsx
+++ b/client/src/components/ProtectedRoute.tsx
@@ -29,7 +29,7 @@ export function ProtectedRoute({ children, role, redirectTo = "/login" }: Protec
   }
 
   if (user && !user.isVerified && user.role !== "ADMIN") {
-    return <Navigate to={`/verify-email?email=${encodeURIComponent(user.email)}`} replace />;
+    return <Navigate to="/verify-email" state={{ email: user.email }} replace />;
   }
 
   if (role && user?.role !== role) {


Sachinchaurasiya360 (changes requested):
The PII-in-URL concern is legit, but this change breaks the verify-email flow as written.

`VerifyEmailPage` reads email **only** from the query string (VerifyEmailPage.tsx:13):

```ts
const email = searchParams.get("email") || "";
```

It never reads `location.state`. After this change the redirect from `ProtectedRoute` lands on the page with an empty email, so the displayed email is blank and verify/resend POST an empty `email`.

Also, 5 other navigators still use the `?email=` query param (LoginPage, RegisterPage, RecruiterLoginPage, RecruiterRegisterPage), so this only changes one of the entry points.

To land this safely:
1. Make `VerifyEmailPage` read `location.state?.email` first, falling back to the query param for compatibility.
2. Migrate the other navigators to `state` too (or keep them consistent).

Sachinchaurasiya360:
Closing: this is the same incomplete change as #2327's other attempt and breaks email verification as-is. It updates only `ProtectedRoute` to pass `email` via route `state`, but `VerifyEmailPage` still reads it from the query string (`useSearchParams().get("email")`), and four other call sites still navigate with `?email=` (LoginPage, RegisterPage, RecruiterLoginPage, RecruiterRegisterPage) — so the verify screen would get an empty email. A complete fix needs `VerifyEmailPage` to read `useLocation().state?.email` (with a query fallback) and all five navigation sites switched to `state`. Happy to re-review once that's in.

PR #2365, opened 2026-06-18
fix: move SqlExercisePage state resets into useEffect to prevent rend…

### LINKED ISSUE #2357: [BUG]:  SqlExercisePage calls setState directly during render phase
### Description

SqlExercisePage.tsx calls multiple setState functions directly inside the component's render body (inside an if statement) rather than wrapping them in a useEffect hook.

### Steps to Reproduce

Go to client/src/module/student/sql/SqlExercisePage.tsx.
Look at lines 125-136.
Observe that multiple setX state setter functions are called directly during the render cycle if exercise?.id !== prevExerciseId.

### Expected Behavior

The state updates should be wrapped in a useEffect hook with [exercise?.id] as a dependency to conform to React's side-effect rules.

### Actual Behavior

State is updated during the render phase. In React 18 concurrent mode, this is an anti-pattern that can cause inconsistent UI state or trigger infinite re-render loops.

### Screenshots / Logs

_No response_

### OS

_No response_

### Browser

_No response_

### Version

_No response_

Prateek2007-cmd:
## Description
This PR resolves Issue #2357 by fixing a React 18 concurrent mode anti-pattern in the SQL exercise page.

## Changes Made
- Updated `client/src/module/student/sql/SqlExercisePage.tsx`.
- Removed the `prevExerciseId` state variable and the direct `if` block that was calling `setState` during the render phase.
- Replaced the logic with a standard `useEffect` hook that depends on `[exercise?.id]` to reset UI elements (editor code, hints, validation, database readiness) when the user navigates between exercises.
- This prevents potential infinite re-render loops and UI tearing that can occur when mutating state while React is calculating the virtual DOM.

Fixes #2357

### DIFF
--- a/client/src/module/student/sql/SqlExercisePage.tsx
+++ b/client/src/module/student/sql/SqlExercisePage.tsx
@@ -120,10 +120,7 @@ export default function SqlExercisePage() {
   const [dbReady, setDbReady] = useState(false);
   const [solved, setSolved] = useState(false);
 
-  const [prevExerciseId, setPrevExerciseId] = useState(exercise?.id);
-
-  if (exercise?.id !== prevExerciseId) {
-    setPrevExerciseId(exercise?.id);
+  useEffect(() => {
     setDbReady(false);
     setResult(null);
     setValidation(null);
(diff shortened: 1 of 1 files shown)


Sachinchaurasiya360 (approved):
Works. One note: the original "setState-during-render with a prev-id guard" is React's recommended flash-free pattern for resetting state when a prop changes. The `useEffect` version runs after paint, so there's a brief render where the previous exercise's code/solved state is shown before the reset, and it needs the `exhaustive-deps` disable. Not a bug, just slightly worse UX than what was there. Fine to merge.

NOTE: branch currently conflicts with main, please rebase before merge.

Sachinchaurasiya360:
Closing as superseded: the SqlExercisePage render-phase state fix already landed on `main` via #2344/#2343 (the resets were moved into the load `useEffect`). Same fix, thank you!

PR #2366, opened 2026-06-18
fix(security): restrict ATS resumeUrl validation to expected S3 host …

### LINKED ISSUE #2358: [BUG]: ATS resumeUrl validation allows arbitrary HTTPS URLs (SSRF vector)
### Description

The scoreResumeSchema validation only verifies that a provided resume URL starts with https://. It does not restrict the domain to the application's trusted S3 bucket.

### Steps to Reproduce

Go to server/src/module/ats/ats.validation.ts.
Look at the scoreResumeSchema definition.
Observe that it uses .startsWith("https://").
Submit an ATS scoring request with an external HTTPS URL (e.g., a webhook receiver or an internal AWS metadata IP masked behind a domain).
Observe the backend server attempting to fetch the external URL.

### Expected Behavior

The schema should strictly validate that the URL hostname matches the application's configured S3 bucket domain to prevent external fetches.

### Actual Behavior

The server blindly accepts and fetches any HTTPS URL. This introduces a Server-Side Request Forgery (SSRF) vulnerability where malicious users could trick the server into making arbitrary external requests.

### Screenshots / Logs

_No response_

### OS

_No response_

### Browser

_No response_

### Version

_No response_

Prateek2007-cmd:
## Description
This PR resolves Issue #2358 by patching a Server-Side Request Forgery (SSRF) vulnerability in the ATS service.

## Changes Made
- Updated `server/src/module/ats/ats.validation.ts`.
- Changed the `scoreResumeSchema` validation so that it no longer accepts any generic `https://` URL.
- The schema now dynamically reads the expected S3 bucket domain from `process.env` and uses the native `URL` API to strictly verify the `hostname` matches our S3 infrastructure. 
- This ensures the server will only fetch resumes that actually belong to the InternHack application, protecting internal networks and metadata APIs from being accessed by malicious user inputs.

Fixes #2358

### DIFF
--- a/server/src/module/ats/ats.validation.ts
+++ b/server/src/module/ats/ats.validation.ts
@@ -1,12 +1,24 @@
 import { z } from "zod";
 
+const BUCKET = process.env["AWS_S3_BUCKET"] || "";
+const REGION = process.env["AWS_REGION"] || "ap-south-1";
+const expectedHost = `${BUCKET}.s3.${REGION}.amazonaws.com`;
+
 export const scoreResumeSchema = z.object({
   resumeUrl: z
     .string()
     .min(1, "Resume URL is required")
     .refine(
-      (url) => url.startsWith("https://") || url.startsWith("/uploads/"),
-      "Resume URL must be an S3 URL or local upload path",
+      (url) => {
+        if (url.startsWith("/uploads/")) return true;
+        try {
+          const parsed = new URL(url);
+          return parsed.hostname === expectedHost;
+        } catch {
+          return false;
+        }
+      },
+      "Resume URL must be a valid S3 URL or local upload path",
     ),
   jobTitle: z.string().max(200).optional(),
   jobDescription: z.string().max(5000).optional(),


Sachinchaurasiya360 (approved):
Good SSRF tightening — the old check accepted any `https://` URL. Pinning the host to `${AWS_S3_BUCKET}.s3.${AWS_REGION}.amazonaws.com` matches the app's upload format (upload.controller.ts).

Two minor notes (non-blocking):
- It doesn't accept the region-less `${bucket}.s3.amazonaws.com` variant that `recruiter.service.ts` already handles. Consider reusing/centralizing that existing S3-host validation for consistency.
- No empty-bucket guard: if `AWS_S3_BUCKET` is unset, `expectedHost` becomes `.s3.<region>.amazonaws.com` and every URL is rejected. `recruiter.service.ts` guards with `if (!S3_BUCKET) return false`.

Approving on the fix itself. NOTE: branch currently conflicts with main, please rebase before merge.

Sachinchaurasiya360:
Closing as superseded: the same ATS resumeUrl SSRF hardening already landed on `main` via #2342. Note the merged version is a bit more permissive in the right way: it also accepts the region-less `<bucket>.s3.amazonaws.com` host and falls back to allowing `https://` when `AWS_S3_BUCKET` is unset, whereas this PR's exact `<bucket>.s3.<region>.amazonaws.com` match would reject those. Thanks for the fix though, same idea!

PR #2389, opened 2026-06-19
Make deleteFile async and use proper error handling with logger

### LINKED ISSUE #2379: Fire-and-forget S3 delete calls in upload controller silently swallow failures
**File:** server/src/module/upload/upload.controller.ts:48-57 (and callers at lines 128, 160, 192-193, 239)

**Description:**
The `deleteFile` helper function uses a fire-and-forget pattern for S3 deletions:

```typescript
function deleteFile(url: string): void {
  const s3Key = getS3KeyFromUrl(url);
  if (s3Key) {
    deleteFromS3(s3Key).catch((err) => console.error("Failed to delete from S3:", err));
  } else if (url.startsWith("/uploads/")) { ... }
}
```

The `deleteFromS3` promise is not awaited - it's fire-and-forget with only a console.error on failure. This is called in `uploadProfilePic` (line 128), `uploadCoverImage` (line 160), `uploadProfileResume` (lines 192-193), and `deleteProfileResume` (line 239). In every case, the HTTP response is sent before the S3 delete completes.

**Impact:** Orphaned S3 objects accumulate when deletion fails. The application has already updated the database record to point to the new file, but the old file remains on S3, wasting storage and potentially exposing stale user data. The user and caller receive no indication that cleanup failed.

**Suggested fix:** Make `deleteFile` return a Promise and await it in the calling methods before sending the response. Use the structured logger instead of console.error.

```typescript
async function deleteFile(url: string): Promise<void> {
  const s3Key = getS3KeyFromUrl(url);
  if (s3Key) {
    await deleteFromS3(s3Key);
  } else if (url.startsWith("/uploads/")) {
    await fs.promises.unlink(path.join(UPLOADS_DIR, "..", url)).catch(() => {});
  }
}
```

Xenon010101:
Closes #2379

### DIFF
--- a/server/src/module/upload/upload.controller.ts
+++ b/server/src/module/upload/upload.controller.ts
@@ -46,13 +46,21 @@ function isValidS3FileUrl(url: unknown): url is string {
 }
 
 /** Delete a file - keeps local fallback support just in case users have legacy local files */
-function deleteFile(url: string): void {
+async function deleteFile(url: string): Promise<void> {
   const s3Key = getS3KeyFromUrl(url);
   if (s3Key) {
-    deleteFromS3(s3Key).catch((err) => console.error("Failed to delete from S3:", err));
+    try {
+      await deleteFromS3(s3Key);
+    } catch (err) {
+      logger.error("Failed to delete from S3:", err);
+    }
   } else if (url.startsWith("/uploads/")) {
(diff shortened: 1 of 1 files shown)


Sachinchaurasiya360:

PR #2392, opened 2026-06-19
fix: add 10s timeout to Google OAuth2Client to prevent hanging on verifyIdToken

### LINKED ISSUE #2317: [BUG]: Google Sign in Fails
### Description

When the user tries to sign in with google account the sign in fails due to backend error

### Steps to Reproduce

1. Go to the Home page
2. Click Sign up
3. Click continue with google 
4. Notice the error stating Google Signup Failed

### Expected Behavior

The user should be able to sign up with his google account

### Actual Behavior

The authentication request times out

### Screenshots / Logs

Image

### OS

_No response_

### Browser

_No response_

### Version

_No response_

saurabhhhcodes:
## Description

The `googleAuth` method calls `OAuth2Client.verifyIdToken()` without a timeout, causing the authentication request to hang indefinitely when Google's API is slow or unreachable. Users report that Google sign-in "fails" and the request "times out".

## Fix

Pass `requestOptions: { timeout: 10000 }` to the `OAuth2Client` constructor so the token verification request fails with a clear error after 10 seconds instead of hanging silently.

## Testing

- `git diff --check` — no whitespace errors  
- Verified the constructor accepts the object-format options in google-auth-library v7+

Closes #2317

### DIFF
--- a/server/src/module/auth/auth.service.ts
+++ b/server/src/module/auth/auth.service.ts
@@ -55,7 +55,10 @@ export class AuthService {
   private googleClient: OAuth2Client;
 
   constructor() {
-    this.googleClient = new OAuth2Client(process.env["GOOGLE_CLIENT_ID"] ?? "");
+    this.googleClient = new OAuth2Client({
+      clientId: process.env["GOOGLE_CLIENT_ID"] ?? "",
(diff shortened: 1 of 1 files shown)


Sachinchaurasiya360:

PR #2397, opened 2026-06-19
perf(github-stats): replace in-memory Map with shared Redis cache and…

### LINKED ISSUE #2304: perf(db): avoid external GitHub stats requests and DB usage on every profile render
## Goal
`GitHubStatsCard` calls `/auth/github-stats`, which can fetch external GitHub data and then render profile stats. Repeated profile visits should not burn API work or related DB usage.

## Scope
- client/src/module/student/profile/GitHubStatsCard.tsx
- server/src/module/auth/auth.service.ts
- server/src/module/auth/auth.routes.ts

## Acceptance Criteria
- Cache GitHub stats by username with a TTL.
- Reuse stored `githubConnection` snapshot where the user has connected GitHub.
- Add client-side React Query staleTime/cacheTime so route changes do not refetch immediately.
- Keep `usageLimit(GITHUB_STATS)` behavior intact.

Suhas334:
## Summary

Resolves #2304 — avoid redundant external GitHub API calls and DB usage on every profile render.

### `server/src/module/auth/auth.service.ts`
- **Replaced** the ephemeral in-memory `githubStatsCache` Map (lost on every server restart) with the shared `cacheGet`/`cacheSet` utility (Redis-backed with in-process fallback). TTL remains 1 hour; `cacheSet` now correctly receives **seconds** (`3600`) instead of milliseconds. Cache key is namespaced as `github-stats:<username>` to avoid key collisions.
- **Added `githubConnection` snapshot short-circuit**: when the requesting user has already connected GitHub via OAuth, the stored DB snapshot (`publicRepos`, `contributedStars`, `contributedRepos`) is used to build `GitHubStats` without making any external GitHub API calls. The result is written to the shared cache on first use, so subsequent requests skip both the DB and the API.
- `usageLimit('GITHUB_STATS')` in `auth.routes.ts` is untouched and continues to guard every request.

### `server/src/module/auth/auth.controller.ts`
- Forward `req.user!.id` to `getGitHubStats` so the service can attempt the per-user `githubConnection` short-circuit.

### `client/src/module/student/profile/GitHubStatsCard.tsx`
- Added `gcTime: 2 * 60 * 60 * 1000` (2 hours) to the `useQuery` options. The existing `staleTime: 1h` prevents refetches during the fresh window; the higher `gcTime` ensures data is **not garbage-collected** between route navigations, so revisiting the profile page within 2 hours never triggers a network request.

## Acceptance Criteria

- [x] Cache GitHub stats by username with a TTL (1 hour, Redis-backed with fallback)
- [x] Reuse stored `githubConnection` snapshot where the user has connected GitHub
- [x] Client-side `staleTime`/`gcTime` so route changes do not refetch immediately
- [x] `usageLimit(GITHUB_STATS)` behaviour intact

## Type Safety

All introduced changes are type-error-free. Pre-existing TS errors in unrelated files (referralLink, studyBuddy, ambassador migrations) are unchanged from main.

## Summary by CodeRabbit

* **Improvements**
  * Updated GitHub stats loading behavior in the profile card with clearer refresh/caching timing to reduce unnecessary refetches.
  * GitHub statistics now leverage authenticated user context to compute stars and top languages more accurately per user.
  * Enhanced caching for GitHub stats using a centralized cache with revised TTL to improve repeat request performance.

### DIFF
--- a/client/src/module/student/profile/GitHubStatsCard.tsx
+++ b/client/src/module/student/profile/GitHubStatsCard.tsx
@@ -43,7 +43,8 @@ export default function GitHubStatsCard({
         .get<{ stats: GitHubStats }>("/auth/github-stats", { params: { username } })
         .then((res) => res.data.stats),
     enabled: !!username,
-    staleTime: 60 * 60 * 1000,
+    staleTime: 60 * 60 * 1000,       // 1 hour – treat cached data as fresh
+    gcTime: 2 * 60 * 60 * 1000,      // 2 hours – keep in memory across navigations
   });
 
   if (!username) {
--- a/server/src/module/auth/auth.controller.ts
+++ b/server/src/module/auth/auth.controller.ts
@@ -173,7 +173,7 @@ export class AuthController {
         return res.status(400).json({ message: "GitHub username is required" });
       }
 
-      const stats = await this.authService.getGitHubStats(username);
+      const stats = await this.authService.getGitHubStats(username, req.user!.id);
       return res.status(200).json({ stats });
     } catch (error) {
       if (error instanceof Error) {
--- a/server/src/module/auth/auth.service.ts
+++ b/server/src/module/auth/auth.service.ts
@@ -884,40 +884,32 @@ export class AuthService {
 
     // 2. If the authenticated user has already connected GitHub via OAuth, reuse
     //    the synced DB snapshot instead of making a live GitHub API call.
-    //    Note: (prisma as any) is needed because the local Prisma client may not yet be
-    //    regenerated after the githubConnection migration was added.
+    //    Both paths compute totalStars and topLanguages identically so the shared
+    //    cache key always contains consistent data regardless of who fills it first.
     if (userId) {
-      type GithubConnectionRow = {
-        githubUsername: string;
-        publicRepos: number;
-        contributedStars: number;
-        contributedRepos: Array<{ language: string | null; mergedPrs: number }>;
-      };
-      const connection = await (prisma as any).githubConnection.findUnique({
+      const connection = await prisma.githubConnection.findUnique({
         where: { userId },
         select: {
           githubUsername: true,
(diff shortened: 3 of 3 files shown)


Sachinchaurasiya360 (commented):
## Solid move to shared cache — two cleanups requested

Verified the `githubConnection` model and every selected field exist in `base.prisma` and the client is generated, so this runs correctly. The switch off the restart-volatile in-memory Map to the shared `cacheGet`/`cacheSet`, the TTL-unit fix (`cacheSet` takes seconds), and the namespaced key are all good and align with the project's "cache read-heavy endpoints" rule. The snapshot short-circuit is a `findUnique` (no transaction on a read path) — also good.

### 1. `(prisma as any)` is unnecessary and the comment is misleading
The comment says the client "may not yet be regenerated after the githubConnection migration was added," but the model is in the source schema (`base.prisma`) and the client is generated — `prisma.githubConnection.findUnique` is fully typed. The `as any` discards type-checking on the `select` (which is why the hand-written `GithubConnectionRow` was needed). Please use `prisma.githubConnection` directly and drop the cast + comment.

### 2. Cache value diverges depending on who fills it first
`github-stats:<username>` is a single global key, but the snapshot path and the live-API path compute different numbers — `totalStars` = `contributedStars` (snapshot) vs the live star sum, and `topLanguages` is ranked by merged-PR counts in the snapshot. So the same endpoint returns different stats depending on whether the owner or a guest populated the cache first. Please make the two paths semantically consistent, or document the intended source of truth.

### Minor
Cache hits still consume `usageLimit('GITHUB_STATS')` quota since that guard is route middleware that runs before the service. Fine for abuse-limiting, but worth a note since the issue goal was to avoid redundant work.

Not blocking — happy to approve once the `as any`/comment cleanup is done and the divergence is addressed or documented.

Sachinchaurasiya360:
rebase it

PR #2398, opened 2026-06-19
perf(db): add stale-while-revalidate cache for public profile reads

### LINKED ISSUE #2303: perf(db): add stale-while-revalidate cache for public profile reads
## Goal
Public profile pages can be viewed repeatedly by guests and recruiters. The auth service already has cache keys, but we should make profile reads explicitly cheap and measurable.

## Scope
- server/src/module/auth/auth.service.ts
- server/src/module/auth/auth.routes.ts
- client/src/module/student/profile/PublicProfilePage.tsx

## Acceptance Criteria
- Confirm public profile response is served from cache for guest and authorized variants.
- Add stale-while-revalidate or short TTL behavior so repeated profile views do not hit Prisma every time.
- Ensure all profile update/upload/resume/badge paths invalidate the right profile cache keys.
- Document expected DB query count for cached vs uncached public profile load.

Suhas334:
## Summary

Resolves #2303 — explicitly makes public profile reads cheap and measurable using `stale-while-revalidate` behaviors on both client and server, alongside correctly busting all cache variants.

### 1. `server/src/module/auth/auth.controller.ts`
- **Cache-Control & SWR**: Added explicit CDN and browser caching behaviors to `getPublicProfile`. 
  - Guest viewers receive `Cache-Control: public, max-age=60, stale-while-revalidate=300`. This allows instant cached loads with background refreshes.
  - Authorized views (recruiters, admins) receive `Cache-Control: private, no-store` to bypass edge caching, preventing sensitive data bleed.
- **Measurability**: Added `X-Cache: HIT | MISS` response headers derived from `cacheHit` so profile access latency and database bypass can be measured.

### 2. `server/src/module/auth/auth.service.ts` & Invalidations
- **Variant Busters**: Cache invalidation was incomplete due to new keys requiring visitor authorization contexts (`:auth` and `:guest`). Replaced exact-match `cacheDel` with `cacheDelPattern` to consistently flush *all* variants for a `userId` and `profileSlug`.
- **Query Counts Documented**: Added documentation tracking the DB query count for public profile loads (0 when cached, 1-2 when uncached).

### 3. `server/src/module/upload/upload.controller.ts`
- **Upload Cache Invalidations**: `uploadProfilePic`, `uploadCoverImage`, and the two profile resume handlers (`uploadProfileResume`, `deleteProfileResume`) now correctly bust both `profile:me:*` and `profile:public:*` cache patterns. Visitors will now see the new avatar, cover, or resume instantly rather than waiting for TTL expiry. 
- Included `profileSlug` in Prisma update selects so slug-based invalidations can run.

### 4. `server/src/cron/subscription-expiry.ts`
- Replaced `cacheDel` with `cacheDelPattern` to invalidate all cache variants when a subscription expires (since `subscriptionPlan` impacts profile badge renderings).

### 5. `client/src/module/student/profile/PublicProfilePage.tsx`
- Added `staleTime: 5 * 60 * 1000` (5 min) to match the backend `PROFILE_TTL`.
- Added `gcTime: 30 * 60 * 1000` (30 min) so repeated navigations to a public profile within an active session instantly reuse the React Query memory cache instead of unnecessarily polling the backend.

## Acceptance Criteria
- [x] Public profile response is served from cache for guest and authorized variants.
- [x] Added `stale-while-revalidate` / short TTL behavior (no repeated Prisma hits).
- [x] Ensure all profile update / upload / resume / badge paths invalidate the right profile cache keys (badges did not require changing since they load separately).
- [x] Documented expected DB query count for cached vs. uncached public profile loads.

## Summary by CodeRabbit

* **Performance Improvements**
  * Improved profile and GitHub stats data caching by tuning React Query refresh/retention windows to reduce repeat network requests.
* **Cache & Freshness**
  * Enhanced public-profile cache invalidation to ensure updated information reflects quickly after profile changes, subscription expiry, and media uploads.
  * Added more consistent caching behavior in responses by setting cache indicators and cache-control rules based on whether content was served from cache and whether the viewer is a guest or authenticated.
* **Bug Fixes**
  * Refined GitHub stats fetching to better utilize requester context for more accurate results.

### DIFF
--- a/client/src/module/student/profile/GitHubStatsCard.tsx
+++ b/client/src/module/student/profile/GitHubStatsCard.tsx
@@ -44,6 +44,7 @@ export default function GitHubStatsCard({
         .then((res) => res.data.stats),
     enabled: !!username,
     staleTime: 60 * 60 * 1000,
+    gcTime: 2 * 60 * 60 * 1000,
   });
 
   if (!username) {
--- a/client/src/module/student/profile/PublicProfilePage.tsx
+++ b/client/src/module/student/profile/PublicProfilePage.tsx
@@ -97,6 +97,8 @@ export default function PublicProfilePage() {
     queryKey: ["public-profile", finalId],
     queryFn: () => api.get(`/auth/profile/${finalId}`).then((res) => res.data.profile as PublicProfile),
     enabled: !!finalId,
+    staleTime: 5 * 60 * 1000,   // 5 min – matches server PROFILE_TTL; no refetch on revisit
+    gcTime: 30 * 60 * 1000,     // 30 min – keep data alive across navigations within the session
   });
 
   if (isLoading) return <LoadingScreen />;
--- a/server/src/cron/subscription-expiry.ts
+++ b/server/src/cron/subscription-expiry.ts
@@ -35,11 +35,11 @@ export async function runSubscriptionExpiry(): Promise<void> {
     },
   });
 
-  // Invalidate cache for each user
-  const { cacheDel } = await import("../utils/cache.js");
+  // Invalidate cache for each user — bust both visitor variants (:auth and :guest).
+  const { cacheDel, cacheDelPattern } = await import("../utils/cache.js");
   for (const userId of userIds) {
     await cacheDel(`profile:me:${userId}`).catch((err) => console.error("Failed to invalidate profile cache:", err));
-    await cacheDel(`profile:public:${userId}`).catch((err) => console.error("Failed to invalidate profile cache:", err));
+    await cacheDelPattern(`profile:public:${userId}:`).catch((err) => console.error("Failed to invalidate profile cache:", err));
   }
 
   console.log(`[Cron] Expired ${userIds.length} subscription(s) and cleared profile cache.`);
--- a/server/src/module/auth/auth.controller.ts
+++ b/server/src/module/auth/auth.controller.ts
@@ -182,7 +182,7 @@ export class AuthController {
         return res.status(400).json({ message: "GitHub username is required" });
       }
 
-      const stats = await this.authService.getGitHubStats(username);
+      const stats = await this.authService.getGitHubStats(username, req.user!.id);
       return res.status(200).json({ stats });
     } catch (error) {
       if (error instanceof Error) {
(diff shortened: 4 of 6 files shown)


Sachinchaurasiya360 (changes requested):
## SWR + invalidation work is good — but it amplifies a PII leak to the CDN

Verified: `getPublicProfile` has only one caller (the controller, updated here), so the return-shape change breaks nothing, and `cacheDelPattern("profile:public:<id>:")` correctly prefix-matches both `:auth` and `:guest` variants. The incomplete-invalidation fix, the `profileSlug` additions to the upload selects, and `private, no-store` for authorized views are all correct.

### 1. Owner PII is cached under the shared `:guest` key — now reaching the CDN
The cache tier is `isVisitorAuthorized ? "auth" : "guest"`, where `isVisitorAuthorized` is only `ADMIN`/`RECRUITER`. But an **owner who is a STUDENT** viewing their own profile gets `isOwner = true` (so `email`/`contactNo` are NOT stripped) while still mapping to the **`:guest`** key. That response — with the owner's email and phone — is cached under the shared guest key and then served to every anonymous visitor.

This violates the project rule: "never cache a privileged/owner response under a shared key." The leak pre-dates this PR at the Redis layer, but this PR makes it materially worse by adding `Cache-Control: public, max-age=60, stale-while-revalidate=300`, which pushes that owner PII into browser and CDN caches.

Since this PR is caching-focused and already touches this function, please fix it here — give the owner its own tier, e.g.:

```ts
const tier = isOwner ? `owner:${visitor!.id}` : isVisitorAuthorized ? "auth" : "guest";
const cacheKey = `profile:public:${identifier}:${tier}`;
```

### 2. Cron busts only the id variant, not the slug
`runSubscriptionExpiry` calls `cacheDelPattern(\`profile:public:${userId}:\`)` but not the `profileSlug` variant, whereas `updateProfile` and the upload handlers bust both. A profile cached by slug stays stale until TTL after a subscription expires. Low impact (5-min TTL), but inconsistent — either select `profileSlug` in the cron query and bust both, or add a note.

Requesting changes on #1 specifically, since it now reaches the CDN. The rest is good to go.

PR #2414, opened 2026-06-19
fix: log Pyodide preload failures instead of swallowing them

### LINKED ISSUE #2404: Bug: Pyodide load failure silently swallowed in python-engine.ts preload()
﻿## Description

In client/src/module/student/python/lib/python-engine.ts, line 61, the preload() method silently swallows all Pyodide load failures:

`	s
preload(): void {
    loadPyodide().catch(() => {});
}
`

If Pyodide fails to load (CDN down, network error, browser compatibility), the error is completely consumed with no logging. The isReady() method on line 64 will return alse, and subsequent calls to execute() will trigger a fresh load attempt that also silently fails, leaving the user with a broken Python execution UI and no indication of what went wrong.

## Impact

- Users see a non-responsive code editor with no error message
- Debugging is impossible because no error is logged to console
- The Python features silently degrade with no feedback

## Suggested Fix

At minimum log the error:

`	s
preload(): void {
    loadPyodide().catch((err) => {
        console.error("[PythonEngine] Failed to preload Pyodide:", err);
    });
}
`

## Severity

Medium — Silent degradation of Python execution features with no diagnostic information.

Akshita-2307:
﻿## Summary

The Pyodide preload method was silently swallowing load failures with an empty \.catch(() => {})\. If the CDN was unreachable, a network error occurred, or the browser was incompatible, there was no diagnostic information anywhere. This change logs the error so developers and support can identify the root cause.

## Changes

- \client/src/module/student/python/lib/python-engine.ts\ — added error logging to the preload catch handler

Closes #2404

### DIFF
--- a/client/src/module/student/python/lib/python-engine.ts
+++ b/client/src/module/student/python/lib/python-engine.ts
@@ -58,7 +58,9 @@ function loadPyodide(): Promise<PyodideInstance> {
 class PythonEngine {
   /** Begin loading Pyodide in the background. */
   preload(): void {
-    loadPyodide().catch(() => {});
+    loadPyodide().catch((err) => {
+      console.warn("[PythonEngine] Pyodide preload failed, will retry on first execute():", err);
+    });
   }
 
   isReady(): boolean {


Sachinchaurasiya360:

PR #2415, opened 2026-06-19
fix: surface camera playback errors instead of swallowing them

### LINKED ISSUE #2405: Bug: camera video playback failure silently swallowed in useFaceDetection
﻿## Description

In client/src/hooks/useFaceDetection.ts, line 94, the camera video playback failure is silently swallowed:

`	s
await videoRef.current.play().catch(() => {});
`

If the browser blocks autoplay (common on mobile browsers or after permission grants), the video never starts playing, but no error is reported. The user sees a frozen/stalled camera preview with no feedback. The hook proceeds to load MediaPipe and start detection intervals that will never produce results because the video element has no frames.

## Impact

- Proctoring camera appears broken with no error to the user
- Face detection never activates silently
- Users may complete proctored tests with no actual monitoring

## Suggested Fix

`	s
try {
    await videoRef.current.play();
} catch (err) {
    const msg = "Failed to start camera preview. Please check your camera.";
    setError(msg);
    onErrorRef.current(msg);
    return;
}
`

## Severity

High — Proctoring feature silently fails, compromising test integrity.

Akshita-2307:
﻿## Summary

The camera preview \ideo.play()\ call in \useFaceDetection.ts\ had a silent catch handler that swallowed all playback errors. When autoplay was blocked by the browser (common on mobile or after permission grants), the camera preview remained frozen with no user feedback, and the face detection loop would run on a video with no frames.

## Changes

- \client/src/hooks/useFaceDetection.ts\ — replaced \wait videoRef.current.play().catch(() => {});\ with a try/catch that reports the error and halts initialization

Closes #2405

### DIFF
--- a/client/src/hooks/useFaceDetection.ts
+++ b/client/src/hooks/useFaceDetection.ts
@@ -91,7 +91,15 @@ export function useFaceDetection(config: FaceDetectionConfig) {
     // Attach to video element
     if (videoRef.current) {
       videoRef.current.srcObject = stream;
-      await videoRef.current.play().catch(() => {});
+      try {
+        await videoRef.current.play();
+      } catch (playErr) {
+        const msg = "Failed to start camera preview. Please check your camera.";
+        setError(msg);
+        onErrorRef.current(msg);
+        setIsLoading(false);
+        return;
+      }
     }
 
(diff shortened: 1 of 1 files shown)


Sachinchaurasiya360:

PR #2421, opened 2026-06-19
fix: add retry button to ErrorBoundary for chunk load failures

### LINKED ISSUE #2411: Bug: lazyWithRetry has no retry mechanism after chunk reload failure
﻿## Description

In client/src/App.tsx, the lazyWithRetry helper (lines 11-24) attempts only **one** page reload when a lazy chunk fails to load:

`	s
function lazyWithRetry(factory: () => Promise<{ default: ComponentType<unknown> }>) {
    return lazy(() =>
        factory().catch((err: unknown) => {
            const key = "chunk_reload";
                sessionStorage.setItem(key, "1");
                window.location.reload();
                return new Promise(() => {}); // never resolves, page is reloading
            }
            sessionStorage.removeItem(key);
            throw err;
        }),
    );
}
`

If the chunk fails to load **a second time** (after the reload), the error propagates to the top-level ErrorBoundary. The user sees a full-screen "Something went wrong" with only a "Go Home" button. There is no retry mechanism for the specific chunk — navigating back to the same page will trigger the same lazy import again, which will likely fail again if the issue is a stale deployment or CDN cache.

## Impact

- Users are stuck with a broken page after a deployment that invalidates old chunk hashes
- No "Try Again" button for the specific failed chunk
- Entire app crashes because only one top-level ErrorBoundary wraps everything

## Suggested Fix

Add a "Retry" button in the ErrorBoundary that clears the chunk_reload sessionStorage key and reloads the page:

`	s
function RetryButton() {
    return (
        <button onClick={() => {
            sessionStorage.removeItem("chunk_reload");
            window.location.reload();
        }}>
            Retry
        </button>
    );
}
`

Or implement more granular error boundaries per route group.

## Severity

Medium — Users are stranded on broken pages after deployments.

Akshita-2307:
﻿## Summary

Added a "Try Again" button to the ErrorBoundary that reloads the page and resets the hasReloaded flag in lazyWithRetry. Previously, if a lazy chunk failed on the second attempt after an automatic reload, the user was shown only a "Go Home" button that would trigger the same broken import again.

## Changes

- client/src/components/ErrorBoundary.tsx — added "Try Again" button alongside existing "Go Home" button

Closes #2411

### DIFF
--- a/client/src/components/ErrorBoundary.tsx
+++ b/client/src/components/ErrorBoundary.tsx
@@ -31,9 +31,14 @@ export class ErrorBoundary extends Component<Props, State> {
           <div className="text-center p-8">
             <h1 className="text-2xl font-bold text-gray-900 dark:text-white mb-4">Something went wrong</h1>
             <p className="text-gray-600 dark:text-gray-400 mb-6">{this.state.error?.message || "An unexpected error occurred"}</p>
-            <Button variant="primary" onClick={() => { this.setState({ hasError: false, error: null }); window.location.href = "/"; }}>
-              Go Home
-            </Button>
+            <div className="flex gap-3 justify-center">
+              <Button variant="primary" onClick={() => { this.setState({ hasError: false, error: null }); window.location.href = "/"; }}>
+                Go Home
+              </Button>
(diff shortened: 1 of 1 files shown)


Sachinchaurasiya360:

PR #2432, opened 2026-06-19
fix: return 409 Conflict instead of 201 Created on duplicate email registration

### LINKED ISSUE #2422: Bug: duplicate email registration returns 201 Created instead of 409 Conflict
## Description

In `server/src/module/auth/auth.controller.ts` lines 40-41, when a user tries to register with an email that already exists, the controller catches the "Email already registered" error but returns HTTP **201 Created** with a success message:

```typescript
if (error instanceof Error && error.message === "Email already registered") {
  return res.status(201).json({ message: "Registration successful. Please verify your email to continue." });
}
```

## Impact

The client interprets a 201 status as a successful registration. The user never sees an error and may be confused when they don't receive a verification email (since the account already exists). It also makes it impossible for the frontend to distinguish between a successful new registration and a duplicate attempt.

## Expected Behavior

A duplicate email should return HTTP **409 Conflict** with an appropriate error message like "An account with this email already exists."

## Suggested Fix

```typescript
if (error instanceof Error && error.message === "Email already registered") {
  return res.status(409).json({ message: "An account with this email already exists." });
}
```

## Severity

Medium — causes confusing UX for users who accidentally try to re-register.

Xenon010101:
Fixes #2422

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed incorrect HTTP response status when attempting to register with an already registered email address.

### DIFF
--- a/server/src/module/auth/auth.controller.ts
+++ b/server/src/module/auth/auth.controller.ts
@@ -38,7 +38,7 @@ export class AuthController {
       return res.status(201).json({ message: "Registration successful. Please verify your email to continue.", user: data.user });
     } catch (error) {
       if (error instanceof Error && error.message === "Email already registered") {
-        return res.status(201).json({ message: "Registration successful. Please verify your email to continue." });
+        return res.status(409).json({ message: "An account with this email already exists." });
       }
       console.error(error);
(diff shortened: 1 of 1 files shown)


Sachinchaurasiya360:

PR #2433, opened 2026-06-19
fix: add input validation for studentId parameter on /activity endpoint

### LINKED ISSUE #2423: Bug: non-numeric studentId parameter returns 403 instead of 400 on /activity endpoint
## Description

In `server/src/module/opensource/opensource.routes.ts` lines 79-84, the `/activity` endpoint accepts an optional `studentId` query parameter. When a non-numeric value (e.g., `?studentId=abc`) is provided, `parseInt(queryStudentId, 10)` returns `NaN`:

```typescript
const queryStudentId = req.query.studentId as string | undefined;
const userId = queryStudentId ? parseInt(queryStudentId, 10) : req.user!.id;

if (queryStudentId && req.user!.role === "STUDENT" && userId !== req.user!.id) {
  return res.status(403).json({ success: false, error: "Cannot view other student's activity" });
}
```

Since `NaN !== req.user!.id` is always `true` for any valid numeric user ID, STUDENT users receive a **403 Forbidden** error instead of a proper **400 Bad Request**.

For non-STUDENT roles (ADMIN/RECRUITER), the code passes `NaN` as `userId` to the Prisma query (`where: { userId: NaN }`), which will cause a Prisma database error since `NaN` is not a valid integer.

## Impact

- STUDENT users get a misleading 403 error instead of a 400 validation error
- ADMIN/RECRUITER users could trigger a 500 error with invalid input
- No input validation on the `studentId` parameter

## Suggested Fix

Add validation to ensure `studentId` is a valid positive integer before using it:

```typescript
const queryStudentId = req.query.studentId as string | undefined;
let userId = req.user!.id;

if (queryStudentId) {
  const parsedId = parseInt(queryStudentId, 10);
  if (isNaN(parsedId) || parsedId <= 0) {
    return res.status(400).json({ success: false, error: "Invalid studentId parameter" });
  }
  userId = parsedId;
}
```

## Severity

Low — requires knowing the endpoint exists and passing a non-numeric value to trigger.

Xenon010101:
Fixes #2423

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved input validation for the activity endpoint. Invalid or malformed parameters now correctly return a 400 error response, providing clearer feedback for API consumers.

### DIFF
--- a/server/src/module/opensource/opensource.routes.ts
+++ b/server/src/module/opensource/opensource.routes.ts
@@ -51,7 +51,15 @@ opensourceRouter.get("/analytics/hacktoberfest", authMiddleware, requireRole("ST
 opensourceRouter.get("/activity", authMiddleware, async (req, res, next) => {
   try {
     const queryStudentId = req.query.studentId as string | undefined;
-    const userId = queryStudentId ? parseInt(queryStudentId, 10) : req.user!.id;
+    let userId = req.user!.id;
+
+    if (queryStudentId) {
+      const parsedId = parseInt(queryStudentId, 10);
+      if (isNaN(parsedId) || parsedId <= 0) {
+        return res.status(400).json({ success: false, error: "Invalid studentId parameter" });
+      }
+      userId = parsedId;
+    }
 
     if (queryStudentId && req.user!.role === "STUDENT" && userId !== req.user!.id) {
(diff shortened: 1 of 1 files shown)


Sachinchaurasiya360:

PR #2436, opened 2026-06-19
fix: remove unused length/targetWords from saveGenerated data type

### LINKED ISSUE #2425: Bug: cover letter saveGenerated silently drops length and targetWords parameters
## Description

In `server/src/module/ats/cover-letter.service.ts` lines 37-49, the `saveGenerated` method accepts `length` and `targetWords` parameters in its input type (lines 32-33) but these are never passed to the `prisma.generatedCoverLetter.create()` call:

```typescript
// Parameter type accepts them:
data: {
  // ...
  length?: "short" | "medium" | "long";
  targetWords?: number;
}

// But Prisma create omits them:
return prisma.generatedCoverLetter.create({
  data: {
    studentId,
    jobTitle:       data.jobTitle       ?? null,
    companyName:    data.companyName    ?? null,
    jobDescription: data.jobDescription,
    content:        data.content,
    tone:           data.tone,
    useProfile:     data.useProfile,
    keySkills:      data.keySkills      ?? null,
    excerpt,
    // length and targetWords are NOT included here
  },
});
```

The controller passes these values (cover-letter.controller.ts lines 69-70), but they are silently dropped.

## Impact

- User's length preference and target word count are passed to the save method but never persisted
- If these fields were intended to be stored for future reference, that data is silently lost
- The misleading parameter type suggests they are used, which could confuse future developers

## Suggested Fix

Either:
1. Add `length` and `targetWords` columns to the `generatedCoverLetter` Prisma model and include them in the create call, OR
2. Remove the unused parameters from the method signature if they are not meant to be persisted

## Severity

Low — silent data loss for optional metadata fields, no runtime crash.

Xenon010101:
The \saveGenerated\ method accepted \length\ and \	argetWords\ in its data parameter type but the Prisma model has no such columns — the fields were silently dropped by the \create\ call.

This removes the fields from the method signature and its sole call site so the type system accurately reflects what gets persisted.

Fixes #2425

### DIFF
--- a/server/src/module/ats/cover-letter.controller.ts
+++ b/server/src/module/ats/cover-letter.controller.ts
@@ -66,8 +66,6 @@ export class CoverLetterController {
         tone:           result.data.tone ?? "professional",
         useProfile:     result.data.useProfile ?? false,
         keySkills:      result.data.keySkills,
-        length:         result.data.length,
-        targetWords:    result.data.targetWords ?? 300,
       }).catch((err) => console.error("Failed to log cover letter usage:", err));
 
       const usage = req.usageInfo
--- a/server/src/module/ats/cover-letter.service.ts
+++ b/server/src/module/ats/cover-letter.service.ts
@@ -29,8 +29,6 @@ export class CoverLetterService {
       tone: string;
       useProfile: boolean;
       keySkills?: string;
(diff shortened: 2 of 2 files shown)


Sachinchaurasiya360:

PR #2438, opened 2026-06-19
fix: use correct period/limit for each action in ATS usage stats

### LINKED ISSUE #2427: Bug: ATS /usage endpoint returns daily limits but route enforces monthly limits
## Description

In `server/src/module/ats/ats.routes.ts` line 32, the `/score` endpoint enforces usage limits with a **monthly** window:

```typescript
atsRouter.post("/score", usageLimit("ATS_SCORE", "monthly"), ...)
```

This uses `MONTHLY_LIMITS` which defines `ATS_SCORE: { FREE: 3, PREMIUM: 20 }` for FREE tier (3 per month).

However, the `/usage` endpoint in `ats.controller.ts` line 140 always returns `DAILY_LIMITS` limits:

```typescript
const usage = Object.entries(DAILY_LIMITS).map(([action, limits]) => ({
  action,
  used: (countMap[action] as number) ?? 0,
  limit: limits[tier],
}));
```

`DAILY_LIMITS` defines `ATS_SCORE: { FREE: 2, PREMIUM: 20 }` — FREE tier users see "2 remaining" per day.

This creates a confusing experience:
- Day 1: User does 2 ATS scores. UI shows "2 of 2 used" (daily limit reached).
- Day 2: UI shows "0 of 2 used" (daily reset). User expects 2 more uses.
- But only 1 monthly use remains (2 of 3 used). After 1 use, the user gets a 429 error.

## Impact

FREE tier users see misleading remaining counts in the UI. The usage stats endpoint reports daily limits regardless of the actual enforcement window used by each route.

## Suggested Fix

Make the `/usage` endpoint window-aware so it returns the correct limit and remaining count for each action based on how the route is enforced. Or align the limits so that daily (2/day) and monthly (3/month) are consistent for the FREE tier.

## Severity

Low — confusing UX but does not crash or cause data loss.

Xenon010101:
﻿The /usage endpoint was querying usage from start of day and returning DAILY_LIMITS for every action. However, ATS_SCORE and ROADMAP_GENERATION are enforced monthly (see ats.routes.ts:32, usage-limits.ts:48-49).

Now queries monthly counts + monthly limits for actions that have a MONTHLY_LIMITS entry, and daily counts + daily limits for the rest. Each entry includes a period field ("daily" or "monthly") so clients can display the correct time window.

Fixes #2427

### DIFF
--- a/server/src/module/ats/ats.controller.ts
+++ b/server/src/module/ats/ats.controller.ts
@@ -2,7 +2,7 @@ import type { Request, Response, NextFunction } from "express";
 import type { AtsService } from "./ats.service.js";
 import { applySuggestionsSchema, scoreResumeSchema } from "./ats.validation.js";
 import { prisma } from "../../database/db.js";
-import { DAILY_LIMITS, getPlanTier } from "../../config/usage-limits.js";
+import { DAILY_LIMITS, MONTHLY_LIMITS, getPlanTier } from "../../config/usage-limits.js";
 import { sendEmail } from "../../utils/email.utils.js";
 import { atsScoreReportHtml } from "../../utils/email-templates.js";
 import { isPremiumUser } from "../../utils/premium.utils.js";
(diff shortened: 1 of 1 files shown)


Sachinchaurasiya360:

PR #2440, opened 2026-06-19
fix: add query validation for getStudentContributionTrend

### LINKED ISSUE #2429: Bug: getStudentContributionTrend lacks Zod validation on startDate/endDate query params
## Description

In `server/src/module/opensource/opensource.controller.ts` line 301, the `getStudentContributionTrend` method uses a `as` type assertion instead of Zod schema validation for its query parameters:

```typescript
const { startDate, endDate } = req.query as { startDate?: string; endDate?: string };
```

This type assertion does nothing at runtime — it only satisfies TypeScript. If a client passes an invalid date string like `?startDate=not-a-date` or garbage data, it passes through unvalidated.

The service method at `opensource.service.ts` lines 547-582 then calls `new Date(startDate)`:

```typescript
const startMonth = startDate ? new Date(startDate) : this.addMonthsUTC(currentMonthStart, -5);
```

`new Date("not-a-date")` returns `Invalid Date`. Then `startMonth.getUTCFullYear()` returns `NaN`. The value `NaN` is then used in the Prisma query at line 554-559:

```typescript
where: {
  userId,
  status: "APPROVED",
  updatedAt: { gte: startMonth, lt: endMonth },
}
```

This causes a Prisma database error, resulting in a 500 Internal Server Error response.

Note: the route at `opensource.routes.ts` line 71 also omits any Zod validation middleware for this endpoint.

## Impact

- Clients passing invalid date strings trigger a 500 error
- No input validation or meaningful error message

## Suggested Fix

Add Zod validation for the query parameters:

```typescript
const trendQuerySchema = z.object({
  startDate: z.string().datetime().optional(),
  endDate: z.string().datetime().optional(),
});
```

And use `safeParse` in the controller:

```typescript
const parsed = trendQuerySchema.safeParse(req.query);
if (!parsed.success) {
  res.status(400).json({ message: "Invalid query parameters", errors: parsed.error.flatten().fieldErrors });
  return;
}
const { startDate, endDate } = parsed.data;
```

## Severity

Low — requires passing deliberately invalid input to trigger.

Xenon010101:
﻿The getStudentContributionTrend endpoint accepts optional startDate and endDate query params but did not validate them. Invalid date strings (e.g. "not-a-date") would create Invalid Date objects, potentially causing unexpected behavior.

Adds contributionTrendQuerySchema with YYYY-MM-DD regex validation for both params, and uses safeParse in the controller.

Fixes #2429

### DIFF
--- a/server/src/module/opensource/opensource.controller.ts
+++ b/server/src/module/opensource/opensource.controller.ts
@@ -12,6 +12,8 @@ import {
   bookmarkBodySchema,
   bulkMigrateBookmarksSchema,
   guideFeedbackSchema,
+  issueCertificateSchema,
+  contributionTrendQuerySchema,
 } from "./opensource.validation.js";
 import { parsePagination } from "../../utils/pagination.utils.js";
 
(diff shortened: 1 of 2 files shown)


Sachinchaurasiya360:

PR #2442, opened 2026-06-19
fix: use rejectRequestSchema instead of approveRequestOverrideSchema for rejection

### LINKED ISSUE #2431: Bug: rejectRepoRequest validates against approveRequestOverrideSchema instead of its own schema
## Description

In `server/src/module/opensource/opensource.controller.ts` line 274, the `rejectRepoRequest` method validates its request body against `approveRequestOverrideSchema` instead of a rejection-specific schema:

```typescript
async rejectRepoRequest(req: Request, res: Response, next: NextFunction) {
  try {
    const id = Number(req.params.id);
    if (isNaN(id)) {
      res.status(400).json({ message: "Invalid request ID" });
      return;
    }

    const body = approveRequestOverrideSchema.safeParse(req.body);
    // ...
```

The schema `approveRequestOverrideSchema` includes fields like `name`, `description`, `domain`, `difficulty`, and `tags` — which are relevant for approving a repo request (where the admin can override metadata), but not for rejecting one. A rejection should only need `adminNote`.

This is fragile: if `approveRequestOverrideSchema` is later modified to require mandatory fields for approval (e.g., `name` is made required), the rejection endpoint would break since it only passes optional fields.

## Impact

- Using the wrong validation schema could cause the rejection endpoint to break if the approval schema is updated
- Confusing for developers maintaining the code
- No separate rejection schema for clarity

## Suggested Fix

Create a dedicated rejection schema or use an inline `z.object({ adminNote: z.string().max(2000).optional() })` for the reject endpoint.

## Severity

Low — works correctly today but is fragile against future schema changes.

Xenon010101:
﻿The rejectRepoRequest endpoint used approveRequestOverrideSchema, which accepts approval-specific fields (name, description, domain, difficulty, tags) that are irrelevant for rejection. Created rejectRequestSchema (adminNote only) and switched the endpoint to use it.

Fixes #2431

### DIFF
--- a/server/src/module/opensource/opensource.controller.ts
+++ b/server/src/module/opensource/opensource.controller.ts
@@ -6,6 +6,7 @@ import {
   gsocOrgsQuerySchema,
   submitRepoRequestSchema,
   approveRequestOverrideSchema,
+  rejectRequestSchema,
   repoIdSchema,
   repoOwnerNameSchema,
   firstPrProgressUpdateSchema,
@@ -273,7 +274,7 @@ export class OpensourceController {
         return;
       }
 
-      const body = approveRequestOverrideSchema.safeParse(req.body);
+      const body = rejectRequestSchema.safeParse(req.body);
       if (!body.success) {
         res.status(400).json({
           message: "Validation failed",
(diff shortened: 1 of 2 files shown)


Sachinchaurasiya360:

PR #2445, opened 2026-06-19
fix: validate studentId query param, return 400 for invalid values (#2423)

saurabhhhcodes:
**Fixes:** #2423

### Changes
Added validation for the `studentId` query parameter on the `/activity` endpoint.

### Before
Non-numeric `studentId` (e.g. `?studentId=abc`) produced `NaN` from `parseInt`, which:
- For STUDENT role: `NaN !== userId` is always truthy → returned misleading 403
- For ADMIN/RECRUITER: `NaN` passed to Prisma `where: { userId: NaN }` → potential 500 error

### After
Returns **400 Bad Request** with a clear error if `studentId` is not a valid positive integer.


Sachinchaurasiya360:

Issue #2408, opened 2026-06-19
Bug: hardcoded S3 bucket URL in Helmet CSP breaks uploads on config change


Akshita-2307:
﻿## Description

In server/src/index.ts line 117, the Content Security Policy connectSrc directive hardcodes a specific S3 bucket URL:

`	s
connectSrc: ["'self'", "https://accounts.google.com", "https://generativelanguage.googleapis.com", "https://www.google-analytics.com", "https://analytics.google.com", "https://intern-hack-prod-bucket.s3.ap-south-1.amazonaws.com"],
`

If the S3 bucket is renamed, migrated to a different region, or a different storage backend is used (e.g., local fallback for development), the CSP will block all upload connections, producing cryptic console errors without any clear indication that the CSP is the cause.

## Impact

- File uploads silently fail in non-production environments
- Changing the S3 bucket name or region requires a code change to update CSP
- Developers debugging upload failures may not immediately suspect CSP

## Suggested Fix

Read the S3 bucket URL from an environment variable:

`	s
const uploadBucketUrl = process.env["UPLOAD_BUCKET_URL"] || "https://intern-hack-prod-bucket.s3.ap-south-1.amazonaws.com";
`

And use that variable in the CSP directives.

## Severity

Medium — Breaks uploads when S3 configuration changes; no dev fallback.

PR #2447, opened 2026-06-19
fix: read S3 bucket URL from env var instead of hardcoding in CSP (#2408)

saurabhhhcodes:
**Fixes:** #2408

### Changes
- Added `UPLOAD_BUCKET_URL` constant reading from `process.env["UPLOAD_BUCKET_URL"]` with the prod S3 URL as default
- Replaced hardcoded S3 URL in `connectSrc` CSP directive with the variable

### Impact
- Changing S3 bucket name/region now only requires setting `UPLOAD_BUCKET_URL` env var
- Local/dev environments can use a different bucket URL or local fallback without CSP blocking uploads


Sachinchaurasiya360:

PR #2448, opened 2026-06-19
fix: remove unused length/targetWords params from saveGenerated (#2425)

saurabhhhcodes:
**Fixes:** #2425

### Changes
Removed `length` and `targetWords` parameters from `CoverLetterService.saveGenerated()` method signature and controller call site.

### Why
These parameters were passed to the service method but silently dropped — the Prisma model has no corresponding columns. Removing them from the method signature makes the API honest and prevents future confusion.

### Impact
Zero runtime change — these values were never persisted.


Sachinchaurasiya360:

PR #2451, opened 2026-06-19
fix: add Zod validation for getStudentContributionTrend query params (#2429)

saurabhhhcodes:
**Fixes:** #2429

### Problem
`getStudentContributionTrend` used a bare `as` type assertion instead of Zod validation. Invalid date strings would produce `Invalid Date` → `NaN` in Prisma queries → 500 error.

### Changes
- Added `trendQuerySchema` to `opensource.validation.ts`
- Used `safeParse` in controller, returning 400 with field errors on invalid input

### Impact
Clients now get a proper 400 error instead of a 500 when passing invalid query parameters.


Sachinchaurasiya360:

PR #2453, opened 2026-06-19
fix: use dedicated rejectRequestSchema instead of approveRequestOverrideSchema (#2431)

saurabhhhcodes:
**Fixes:** #2431

### Problem
`rejectRepoRequest` validated against `approveRequestOverrideSchema` which includes approval-specific fields (name, description, domain, difficulty, tags). If the approval schema is later updated to require mandatory fields, the rejection endpoint would break.

### Changes
- Added `rejectRequestSchema` (only `adminNote` optional field) to `opensource.validation.ts`
- Updated controller to use it instead of `approveRequestOverrideSchema`


Sachinchaurasiya360:

Issue #2434, opened 2026-06-19
Bug: issueCertificate deserializes guideName from req.body with no Zod schema


Xenon010101:
﻿**Description**
The POST /certificate/issue endpoint's controller method issueCertificate at opensource.controller.ts:383 validates guideName using a simple truthiness check (if (!guideName)) instead of validating through a Zod schema. All other POST endpoints in this module use safeParse with a dedicated Zod schema.

**Impact**
- guideName: 123 (number) passes validation and gets stored as the number 123
- guideName: "   " (whitespace) passes validation
- Inconsistent with the module's pattern (compare to submitGuideFeedback, ddBookmark, ulkMigrateBookmarks, 
ejectRepoRequest which all use Zod schemas)

**Fix**
Create a Zod schema (e.g., issueCertificateSchema) in opensource.validation.ts and use safeParse instead of the manual if (!guideName) check on line 386.

PR #2454, opened 2026-06-19
fix: use Zod schema for issueCertificate guideName validation (#2434)

saurabhhhcodes:
**Fixes:** #2434

### Problem
`issueCertificate` validated `guideName` with a simple truthiness check (`if (!guideName)`) instead of a Zod schema. Non-string values (e.g. number `123`) or whitespace-only strings would pass validation.

### Changes
- Added `issueCertificateSchema` (guideName: z.string().min(1)) to validation
- Updated controller to use `safeParse` instead of manual check


Sachinchaurasiya360:

PR #2463, opened 2026-06-19
fix: add Retry button to ErrorBoundary for chunk load recovery (#2411)

saurabhhhcodes:
**Fixes:** #2411

### Problem
After lazy chunk reload fails, ErrorBoundary only shows "Go Home" button. Navigating back to the same page would trigger the same failing lazy import with no recovery path.

### Changes
Added a "Retry" button alongside "Go Home" that calls `window.location.reload()`, resetting the module-level `hasReloaded` flag and giving the lazy chunk another chance to load.


Sachinchaurasiya360:

Issue #2403, opened 2026-06-19
Bug: client prerender always skipped due to hardcoded true, breaking all SEO


Akshita-2307:
﻿## Description

In client/vite.config.ts line 46, skipPrerender is **hardcoded to 	rue**:

`	s
const skipPrerender = true // Temporarily disabled due to Puppeteer connection timeout
  // process.env.SKIP_PRERENDER === '1' || process.env.VERCEL === '1'
`

The commented-out code on line 47 shows the original intent was to conditionally skip prerender only on Vercel (where Puppeteer can't launch). Instead, prerender is disabled **for every build**, meaning **all 38 SEO-critical public pages** (/, /login, /learn/*, etc.) are served as empty shells with no pre-rendered HTML.

## Impact

- Search engines (Google, Bing) see blank pages instead of actual content
- Open Graph / Twitter Card previews fail to show proper metadata
- Page load performance is worse for all first-time visitors

## Suggested Fix

Uncomment the environment check:

`	s
const skipPrerender = process.env.SKIP_PRERENDER === '1' || process.env.VERCEL === '1'
`

And fix the Puppeteer dependency issue separately (or add a Puppeteer launch flag fallback).

## Severity

High — All public SEO and social previews are broken.

PR #2465, opened 2026-06-19
fix: restore conditional skipPrerender env check instead of hardcoded true (#2403)

saurabhhhcodes:
**Fixes:** #2403

### Problem
`skipPrerender` was hardcoded to `true`, disabling prerendering for all 38+ SEO-critical public pages. Search engines saw empty shells instead of actual content.

### Changes
Restored the original environment check: `process.env.SKIP_PRERENDER === '1' || process.env.VERCEL === '1'`


Sachinchaurasiya360:

Issue #2406, opened 2026-06-19
Bug: fullscreen exit failure silently swallowed in SkillTestPage


Akshita-2307:
﻿## Description

In client/src/module/student/skill-verification/SkillTestPage.tsx, line 202, the fullscreen exit failure is silently swallowed:

`	s
if (document.fullscreenElement) {
    document.exitFullscreen().catch(() => { });
}
`

If document.exitFullscreen() fails (e.g., the document was never in fullscreen, or the request was cancelled by a user gesture), the error is silently consumed. More critically, if the fullscreen exit **never resolves** (rare browser quirk), and the user is stuck in fullscreen mode, there is no recovery path. The user cannot interact with the result page or close the tab.

## Impact

- Users can get stuck in fullscreen mode after submitting a test
- No visual feedback or recovery mechanism provided
- Frustrating UX for an already stressful test-taking experience

## Suggested Fix

Log the error and implement a force-exit fallback with a timeout:

`	s
if (document.fullscreenElement) {
    try {
        await document.exitFullscreen();
    } catch (err) {
        console.warn("[SkillTest] Fullscreen exit failed:", err);
        // Fallback: try again or inform user
    }
}
`

## Severity

Medium — Users can become trapped in fullscreen mode with no escape.

PR #2466, opened 2026-06-19
fix: log Pyodide preload failures instead of silently swallowing (#2404)

saurabhhhcodes:
**Fixes:** #2404

### Problem
`preload()` used `.catch(() => {})` which completely silenced all Pyodide load failures. Users saw a broken Python editor with no indication of what went wrong.

### Changes
Added `console.error` logging so failures are visible in devtools console.

### Also applicable
Same pattern in #2405 (camera) and #2406 (fullscreen) — addressing those separately if needed.


Sachinchaurasiya360:

PR #2467, opened 2026-06-19
fix: surface camera playback errors instead of silently swallowing (#2405)

saurabhhhcodes:
**Fixes:** #2405

### Problem
`videoRef.current.play().catch(() => {})` silently swallowed all camera playback failures. If the browser blocked autoplay, the proctoring camera preview appeared frozen with no error.

### Changes
Replaced `.catch(() => {})` with try/catch that calls `setError()` with a user-facing message and returns early.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling when camera preview fails to start—users now receive clear, informative error messages instead of silent failures, providing better visibility into camera initialization issues.


Sachinchaurasiya360:

PR #2468, opened 2026-06-19
fix: surface fullscreen exit failures instead of silently swallowing (#2406)

saurabhhhcodes:
**Fixes:** #2406

### Problem
`document.exitFullscreen().catch(() => {})` silently swallowed fullscreen exit failures. If the browser denied the exit request (common with user-gesture requirements in modern browsers), there was no indication to the user.

### Changes
Replaced empty catch with `toast.error` to surface the failure.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling when exiting fullscreen mode during skill tests—users will now receive a notification if the action fails instead of errors being silently ignored.


Sachinchaurasiya360:

Issue #2378, opened 2026-06-19
Missing error handling for async route handlers in health check routes


Xenon010101:
**File:** server/src/module/health/health.routes.ts:23-38

**Description:**
The `/ready` and `/` health check routes use bare async handlers without try-catch or `.catch()`:

```typescript
healthRouter.get("/ready", async (_req, res) => {
  const result = await getReadiness();
  const statusCode = result.status === "ready" ? 200 : 503;
  res.status(statusCode).json(result);
});

healthRouter.get("/", async (_req, res) => {
  const result = await getFullHealth();
  const statusCode = result.status === "ok" ? 200 : 503;
  res.status(statusCode).json(result);
});
```

If `getReadiness()` or `getFullHealth()` throws an unexpected error (e.g., database connection failure), the promise rejection becomes unhandled. In Express 5, async route handler rejections should be caught automatically, but if this doesn't work in all scenarios, it could crash the process or hang the request.

**Impact:** Health check endpoints could fail silently or cause unhandled promise rejections. During a real outage, the health endpoint itself could hang rather than returning 503.

**Suggested fix:** Wrap the async handlers with try-catch and pass errors to next:
```typescript
healthRouter.get("/ready", async (_req, res, next) => {
  try {
    const result = await getReadiness();
    res.status(result.status === "ready" ? 200 : 503).json(result);
  } catch (err) {
    next(err);
  }
});
```

PR #2472, opened 2026-06-20
fix: add try/catch to health check async route handlers (#2378)

saurabhhhcodes:
**Fixes:** #2378

The `/ready` and `/` health route handlers were async but lacked try/catch. If `getReadiness()` or `getFullHealth()` threw, the error would be an unhandled promise rejection with no response sent to the client.

Added try/catch that returns 503 with an error message on failure.

## Summary by CodeRabbit

* **Bug Fixes**
  * Health check endpoints now properly handle failures with HTTP 503 responses and detailed error information, improving system reliability monitoring and providing clearer feedback when services are unhealthy.

### DIFF
--- a/server/src/module/health/health.routes.ts
+++ b/server/src/module/health/health.routes.ts
@@ -3,37 +3,26 @@ import { getReadiness, getFullHealth } from "./health.service.js";
 
 export const healthRouter = Router();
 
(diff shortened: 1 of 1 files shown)


Sachinchaurasiya360:

Issue #2370, opened 2026-06-19
Zod validation bypassed for \q\ parameter in searchTalent endpoint


Xenon010101:
**File:** \server/src/module/recruiter/recruiter.controller.ts:290-299\

**Description:**
The \searchTalent\ method first validates the query against \	alentSearchSchema\ via \safeParse\ (line 294), but then reads a \q\ parameter directly from \
eq.query\ bypassing Zod validation (line 298):

\\\	ypescript
const result = talentSearchSchema.safeParse(req.query);
if (!result.success) return res.status(400).json(...);
const filters = { ...result.data };
const q = (req.query as Record<string, unknown>).q as string | undefined;  // UNVALIDATED
if (q && !filters.search) filters.search = q;
\\\

The \q\ parameter is not declared in \	alentSearchSchema\ (see \
ecruiter.validation.ts:114-127\). This raw user input is then passed as \ilters.search\ to the service layer and ultimately used in Prisma queries.

**Impact:** Unvalidated user input flows into database queries. The input isn't sanitized, transformed, or constrained by the schema definition.

**Suggested fix:** Add \q: z.string().optional()\ to \	alentSearchSchema\ in \
ecruiter.validation.ts\, and remove the manual \
eq.query.q\ read from the controller so all input goes through Zod validation.

PR #2473, opened 2026-06-20
fix: add q param to talentSearchSchema to prevent Zod validation bypass (#2370)

saurabhhhcodes:
**Fixes:** #2370

### Problem
The `talentSearchSchema` didn't include `q` as a valid param. Zod strips unknown keys by default, so `?q=foo` was silently dropped. The controller then read `q` directly from `req.query` (cast to `any`), bypassing Zod validation entirely.

### Changes
1. Added `q: z.string().optional()` to `talentSearchSchema`
2. Updated controller to read `q` from `result.data` instead of raw `req.query`


Sachinchaurasiya360:

PR #2475, opened 2026-06-20
fix: make deleteFile async and await callers to surface S3 delete failures (#2379)

saurabhhhcodes:
**Fixes:** #2379

### Problem
`deleteFile()` in `upload.controller.ts` was fire-and-forget — the S3 delete promise was caught with a console.error but never awaited. Callers had no way to know the delete failed, and old files could be silently orphaned in S3.

### Changes
1. Made `deleteFile` return `Promise<void>`
2. Replaced `.catch()` with `try/await/catch` using the structured logger
3. Updated all four call sites to `await deleteFile()`

### DIFF
--- a/server/package-lock.json
+++ b/server/package-lock.json
@@ -73,6 +73,9 @@
         "typescript": "^5.9.3",
         "typescript-eslint": "^8.59.4",
         "vitest": "^4.1.7"
+      },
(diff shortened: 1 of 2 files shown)


Sachinchaurasiya360:

Issue #2373, opened 2026-06-19
Inconsistent customFieldAnswers type in applyToJobSchema (missing nested record variant)


Xenon010101:
**File:** server/src/module/student/student.validation.ts:3-15

**Description:**
There are two similar but inconsistent type definitions for custom field answer values:

```typescript
// Line 3-9: fieldAnswerValueSchema (complete, used by submitRoundSchema)
const fieldAnswerValueSchema = z.union([
  z.string(), z.number(), z.boolean(), z.array(z.string()),
  z.record(z.string(), z.number()), // includes nested record
]);

// Line 12: applyToJobSchema.customFieldAnswers (used by job applications)
customFieldAnswers: z.record(z.string(), z.union([
  z.string(), z.number(), z.boolean(), z.array(z.string())
  // MISSING z.record(z.string(), z.number())
])).default({}),
```

The `applyToJobSchema.customFieldAnswers` is missing the `z.record(z.string(), z.number())` variant.

**Impact:** Job applications containing nested record answer values (e.g., structured assessment responses) that would be valid for `submitRoundSchema` will be rejected by `applyToJobSchema`, causing false validation failures.

**Suggested fix:** Replace the inline union with a reference to `fieldAnswerValueSchema`:
```typescript
customFieldAnswers: z.record(z.string(), fieldAnswerValueSchema).default({}),
```

PR #2476, opened 2026-06-20
fix: use fieldAnswerValueSchema in applyToJobSchema for nested record support (#2373)

saurabhhhcodes:
**Fixes:** #2373

### Problem
`applyToJobSchema.customFieldAnswers` used an inline union (`string | number | boolean | string[]`) that was missing the nested record variant. The existing `fieldAnswerValueSchema` already included `z.record(z.string(), z.number())` but was only used by `submitRoundSchema`. This caused Zod validation errors when custom field values contained nested objects.

### Changes
Replaced the inline union in `applyToJobSchema.customFieldAnswers` with the shared `fieldAnswerValueSchema`.

### DIFF
--- a/server/src/module/student/student.validation.ts
+++ b/server/src/module/student/student.validation.ts
@@ -9,7 +9,7 @@ const fieldAnswerValueSchema = z.union([
 ]);
 
 export const applyToJobSchema = z.object({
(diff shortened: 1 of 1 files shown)


Sachinchaurasiya360:

PR #2477, opened 2026-06-20
fix: remove permissive HTTPS fallback in isAllowedUrl to prevent SSRF (#2358)

saurabhhhcodes:
**Fixes:** #2358

### Problem
`isAllowedUrl` in `ats.validation.ts` had a fallback: `if (!S3_BUCKET) return url.startsWith("https://")`. When `AWS_S3_BUCKET` env var was unset (common in development), ANY HTTPS URL passed validation, creating an SSRF vector — an attacker could submit `https://internal-service/admin` and the server would fetch it.

### Changes
Replaced the permissive fallback with `return false`. When the bucket is not configured, only local `/uploads/` paths are allowed.


Sachinchaurasiya360:

PR #2504, opened 2026-06-21
feat: add useLocalStorage hook with cross-tab sync

### LINKED ISSUE #2503: feat: add useLocalStorage hook with cross-tab sync
## Description

Add a reusable `useLocalStorage` hook that provides:
- SSR-safe localStorage read/write with JSON serialization
- Functional updater support (like `useState`)
- Cross-tab synchronization via the `storage` event
- Cleanup method to remove the key from storage

This is a common utility pattern that will reduce duplicated state-persistence logic across the codebase.

### Changes
- `client/src/hooks/useLocalStorage.ts` — new hook (37 lines)

### Related
Part of gssoc26 contributions.

ashroxy:
## Summary

Adds a reusable `useLocalStorage` hook to `client/src/hooks/` that provides persistent state management with localStorage.

### Features
- SSR-safe initialization with fallback to `initialValue`
- Functional updater pattern (same API as `useState`)
- Cross-tab synchronization via the native `storage` event
- `removeValue` method to clear the stored key
- Graceful error handling for private browsing / quota exceeded

### Changes
- `client/src/hooks/useLocalStorage.ts` — new hook (55 lines)

### Related
Closes #2503
Part of gssoc26 contributions.

## Summary by CodeRabbit

* **New Features**
  * New state persistence feature that stores application data in browser storage. Data automatically syncs across multiple tabs and windows, with safe error handling for quota or private browsing modes.

### DIFF
--- a/client/src/hooks/useLocalStorage.ts
+++ b/client/src/hooks/useLocalStorage.ts
@@ -1,0 +1,55 @@
+import { useState, useEffect, useCallback } from "react";
+
+export function useLocalStorage<T>(
+  key: string,
+  initialValue: T,
+): [T, (value: T | ((prev: T) => T)) => void, () => void] {
+  const [storedValue, setStoredValue] = useState<T>(() => {
+    try {
+      const item = window.localStorage.getItem(key);
+      return item ? (JSON.parse(item) as T) : initialValue;
+    } catch {
+      return initialValue;
+    }
+  });
+
+  const setValue = useCallback(
+    (value: T | ((prev: T) => T)) => {
+      setStoredValue((prev) => {
+        const next = value instanceof Function ? value(prev) : value;
+        try {
+          window.localStorage.setItem(key, JSON.stringify(next));
+        } catch {
+          /* quota exceeded or private mode */
+        }
+        return next;
+      });
+    },
+    [key],
+  );
+
+  const removeValue = useCallback(() => {
+    try {
+      window.localStorage.removeItem(key);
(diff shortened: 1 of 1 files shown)


Sachinchaurasiya360 (commented):
Reviewed the `useLocalStorage` hook. Clean implementation (lazy init, functional updates, quota/JSON guards). Two real edge cases and a minor nit:

1. **Cross-tab removal isn't synced.** The `storage` listener only acts on `e.newValue !== null`, so when another tab calls `removeValue()` (which fires a `storage` event with `newValue === null`), this tab keeps the stale value instead of resetting. Since the PR title sells cross-tab sync, handle the null case:
```ts
if (e.key === key) {
  if (e.newValue === null) setStoredValue(initialValue);
  else try { setStoredValue(JSON.parse(e.newValue) as T); } catch {}
}
```
2. **A changing `key` doesn't re-read.** The lazy initializer runs once, so if `key` changes between renders, `storedValue` still holds the old key's value until the next write. Fine if keys are always static; worth a doc comment otherwise.
3. Minor: `removeValue` depends on `initialValue` — if a caller passes an inline object/array literal, its identity changes each render and so does `removeValue`. Most callers pass primitives, so low impact.

Also: this duplicates ad-hoc `localStorage` usage already in several modules (e.g. theme/progress stores). Refactoring one existing call site to use the hook would prove it earns its place and avoid it sitting as unused code. No tests included.

PR #2533, opened 2026-06-23
fix(db): backfill before dropping legacy array columns (no data loss)

Sachinchaurasiya360:
## Problem

Three columns are dropped by migrations already merged to main, but the data was never migrated, so a production \`migrate deploy\` would silently lose data:

- \`user.trackedPrograms\` — **747** non-null values
- \`userInterviewProgress.completedIds\` — **3** values
- \`userInterviewProgress.bookmarkedIds\` — **3** values

The backfill `.sql` scripts existed under `server/src/database/scripts/` but were **not wired into the deploy path** (`build` → `migrate-on-deploy.mjs` runs only `prisma migrate deploy`), so they would never run.

## Fix

Fold each backfill into the **top of its drop migration**, before the `DROP`:

- `20260622111111_drop_tracked_programs` → backfills `programInterest` from `user.trackedPrograms`, then drops.
- `20260622333333_drop_interview_progress_arrays` → backfills `UserInterviewQuestionState` from `completedIds`/`bookmarkedIds`, then drops.

Each backfill only depends on columns/tables created by the preceding `add_*` migration, and is idempotent (`ON CONFLICT DO NOTHING` / `DO UPDATE`).

Also fixed a latent bug in `backfill-program-interest.sql`: the `FOREACH` loop variable over a `text[]` must be `TEXT`, not `RECORD` (would error at runtime).

## Note

Do **not** use `prisma db push` for these changes against a DB with real data — it bypasses migration files (and thus the backfill). Use `migrate deploy` / `npm run db:deploy`, which is what prod runs.

## Summary by CodeRabbit

* **Chores**
  * Executed internal data restructuring operations to migrate and consolidate existing records across the system. These comprehensive updates ensure complete data preservation while reorganizing information for improved system reliability, enhanced performance, and better data consistency. The changes optimize internal data organization and prepare the platform for future scalability improvements.

### DIFF
--- a/server/src/database/prisma/migrations/20260622111111_drop_tracked_programs/migration.sql
+++ b/server/src/database/prisma/migrations/20260622111111_drop_tracked_programs/migration.sql
@@ -1,3 +1,36 @@
--- Phase 2: Clean up the dropped array column
+-- Phase 2: Backfill, then clean up the legacy array column.
+--
+-- Backfill legacy user.trackedPrograms (slug array) into the programInterest
+-- join table BEFORE dropping the column, so no tracked-program data is lost
+-- when this migration is applied on deploy. The programInterest.active /
+-- updatedAt columns this relies on are added in the preceding migration
+-- 20260622094358_add_program_interest_fields. Idempotent via ON CONFLICT.
+DO $$
+DECLARE
+  u       RECORD;
+  p       TEXT;
+  prog_id INT;
+BEGIN
+  FOR u IN
+    SELECT id, "trackedPrograms"
+    FROM "user"
+    WHERE array_length("trackedPrograms", 1) > 0
+  LOOP
+    FOREACH p IN ARRAY u."trackedPrograms"
+    LOOP
+      SELECT id INTO prog_id FROM "ossProgram" WHERE slug = p;
+      IF prog_id IS NULL THEN
+        RAISE WARNING '[SKIP] Program slug "%" not found for user %', p, u.id;
+        CONTINUE;
+      END IF;
+
+      INSERT INTO "programInterest" ("userId", "programId", "active", "createdAt", "updatedAt")
+      VALUES (u.id, prog_id, TRUE, NOW(), NOW())
+      ON CONFLICT ("userId", "programId")
+      DO UPDATE SET "active" = TRUE, "updatedAt" = NOW();
+    END LOOP;
+  END LOOP;
+END $$;
+
 DROP INDEX IF EXISTS "programInterest_userId_idx";
 ALTER TABLE "user" DROP COLUMN IF EXISTS "trackedPrograms";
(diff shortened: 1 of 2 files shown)


Sachinchaurasiya360:
## Post-merge review — please rebase onto current `main` and reopen as a fresh PR

I reviewed this against the **live prod DB** and the backfill, while correct SQL, can't achieve its goal on prod and the migration batch it belongs to won't survive `migrate deploy`. Flagging because this is already merged.

### What I found on prod (read-only check of `_prisma_migrations` + `information_schema`)

| Fact | Value |
|---|---|
| Latest applied migration | `20260620_add_opensource_repo_composite_indexes` |
| All five `20260622*` migrations | **pending** (not recorded as applied) |
| `user.trackedPrograms` | **already dropped** |
| `userInterviewProgress.completedIds` / `bookmarkedIds` | **already dropped** |
| `programInterest.active`, `UserInterviewQuestionState` table | **already exist** |
| `programInterest` / `UserInterviewQuestionState` row counts | **0 / 0** |

Prod was `db push`'d to the final post-normalization state without recording the migrations.

### Consequences

1. **The backfill is moot on prod and the legacy data is already gone.** The source columns (`trackedPrograms`, `completedIds`, `bookmarkedIds`) no longer exist and the normalized tables are empty, so "backfill before drop" can't recover anything here.
2. **The whole `20260622*` batch will fail `migrate deploy` on prod.** The preceding add-migrations use bare `ALTER TABLE ... ADD COLUMN "active"` and `CREATE TABLE "UserInterviewQuestionState"` (no `IF NOT EXISTS`). Those objects already exist, so `migrate deploy` errors at the first add-migration, before reaching these backfills. If it did reach them, the backfill would then fail with `column "trackedPrograms" does not exist`, leaving a half-applied failed migration (the same stuck-`_prisma_migrations` situation the 2026-06-23 baseline repair just cleaned up).

### The SQL itself (correct for a fresh DB)

- `drop_interview_progress_arrays`: the insert/insert/update sequence handles all four states (completed-only, bookmarked-only, both, neither) correctly; `ON CONFLICT` targets the real `UserInterviewQuestionState_userId_questionId_key`. Good.
- `drop_tracked_programs`: loop + `ON CONFLICT ("userId","programId")` matches `programInterest_userId_programId_key`, and skipping unknown slugs avoids the `programId` FK violation. Nit: prefer a set-based `INSERT ... SELECT ... CROSS JOIN LATERAL unnest(...) JOIN "ossProgram"` (matches the project's "bulk in SQL" rule and the sibling migration's style).

### Requested changes for the rebased/follow-up PR

- Baseline the `20260622*` batch on prod with `prisma migrate resolve --applied <name>` (prod already matches the end state) rather than executing it.
- Make the batch idempotent so it's safe to execute anywhere: `ADD COLUMN IF NOT EXISTS`, `CREATE TABLE IF NOT EXISTS`, and guard each backfill (`to_regclass` / `information_schema` check) so it no-ops when the source column is already gone instead of throwing.
- Convert the tracked-programs PL/pgSQL loop to the set-based form.

The PDF worker change in the same line of work (DOMMatrix polyfill + dynamic import) is correct and unrelated to the above.

PR #2544, opened 2026-06-23
fix: correct free-tier city filter skip/take logic to avoid zero results

### LINKED ISSUE #2521: Bug: free-tier city filter can return 0 results with valid pagination due to skip/take cap
**Description**
In `server/src/module/company/company.service.ts:102-112`, the free-tier city filter limiting logic can silently return 0 results with misleading pagination metadata.

When a free user requests page N with city filter and the skip exceeds 20:

```typescript
const effectiveLimit = isCityLimited ? Math.min(limit, 20) : limit;
const effectiveSkip = isCityLimited ? Math.min(skip, 20) : skip;

skip: isCityLimited && effectiveSkip >= 20 ? 20 : effectiveSkip,
take: isCityLimited ? Math.max(0, 20 - effectiveSkip) : effectiveLimit,
```

If `effectiveSkip >= 20`, then `take = Math.max(0, 20 - 20) = 0`, and the query returns 0 results. But the pagination response (`total: Math.min(total, 20)`, `totalPages: ceil(min(total,20) / limit)`) can show valid-looking page numbers.

**Impact**
- Free users see empty company lists on pages beyond what 20 results fill
- No error or message explaining they've hit the free tier limit
- The response has `limited: true` but the UI has no way to distinguish "no results in this page" from "no results because of limit"

**Expected behavior**
Return the first 20 results regardless of page, and cap totalPages at 1 when city-limited, or return an error with a clear upgrade prompt.

Xenon010101:
Closes #2521

## Summary
Removes the skip capping and complex take calculation for free-tier city filtering. The 20-item cap is already properly enforced via the pagination.total field. The old logic caused pages beyond page 2 to return zero results.

### DIFF
--- a/server/src/module/company/company.service.ts
+++ b/server/src/module/company/company.service.ts
@@ -100,14 +100,14 @@ export class CompanyService {
     // Free/unauthenticated users: cap at 20 when filtering by city
     const isCityLimited = !!params.city && tier === "FREE";
     const effectiveLimit = isCityLimited ? Math.min(limit, 20) : limit;
-    const effectiveSkip = isCityLimited ? Math.min(skip, 20) : skip;
+    const effectiveSkip = skip;
 
     const [companies, total] = await Promise.all([
       prisma.company.findMany({
         where,
         orderBy: { avgRating: "desc" },
-        skip: isCityLimited && effectiveSkip >= 20 ? 20 : effectiveSkip,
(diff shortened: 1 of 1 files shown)


Xenon010101:
@web-flow please reopen as this issue is assigned to me.

PR #2577, opened 2026-06-26
fix: use user name instead of email in enrollment PDF welcome email

### LINKED ISSUE #2517: Bug: enrollment PDF welcome email uses req.user.email as user name instead of req.user.name
**Description**
In `server/src/module/roadmap/roadmap.controller.ts:180`, the enrollment welcome email PDF generation passes `req.user!.email` as the `user.name` field:

```typescript
const pdfBuffer = await generateRoadmapPdf({
  user: { name: req.user!.email },
  ...
});
```

The `generateRoadmapPdf` function expects `user.name` to be the user's display name, but the user's email address is passed instead.

**Impact**
- The welcome PDF roadmap shows the user's email address in place of their name
- Example: "This roadmap belongs to johndoe@gmail.com" instead of "This roadmap belongs to John Doe"

**Fix**
Use `req.user!.name` instead of `req.user!.email`:
```typescript
user: { name: req.user!.name },
```

Xenon010101:
Fixes user name display in enrollment PDF welcome email.

Closes #2517

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the welcome-email PDF to use the recipient’s name (from their profile) instead of their email address.
  * Minor end-of-file formatting adjustment with no visible behavior change.

### DIFF
--- a/server/src/module/roadmap/roadmap.controller.ts
+++ b/server/src/module/roadmap/roadmap.controller.ts
@@ -176,8 +176,12 @@ export async function enroll(req: Request, res: Response, next: NextFunction) {
         enrollmentId: enrollment.id,
       });
       if (full) {
+        const u = await prisma.user.findUnique({
+          where: { id: req.user!.id },
+          select: { name: true },
+        });
         const pdfBuffer = await generateRoadmapPdf({
-          user: { name: req.user!.name },
+          user: { name: u?.name ?? req.user!.email },
           roadmap: {
             title: full.roadmap.title,
             shortDescription: full.roadmap.shortDescription,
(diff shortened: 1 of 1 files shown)


Sachinchaurasiya360:
The intent here is right (the PDF should show the user's name, not their email), but as written this won't compile.

`req.user` is typed as only `{ id, email, role }` (see `server/src/types/express.d.ts`), and the auth middleware only ever sets those three fields from the JWT (`auth.middleware.ts`). So `req.user!.name` is a TypeScript error (TS2339, property does not exist) and would be `undefined` at runtime.

You need to source the name from the DB. Since this handler already runs queries, the simplest fix is to fetch it:

```ts
const u = await prisma.user.findUnique({
  where: { id: req.user!.id },
  select: { name: true },
});
// ...
user: { name: u?.name ?? req.user!.email },
```

The `?? req.user!.email` keeps a sensible fallback if name is ever null. Could you update this? Then it's good to merge.

Xenon010101:
Applied the suggested fix — now fetches the user's name from the DB via \prisma.user.findUnique\ instead of accessing \
eq.user!.name\, with a fallback to \
eq.user!.email\ if name is null.

PR #2594, opened 2026-06-28
refactor: address PR review feedback for application UI components

Zeelkundariya:
## Description
Addresses PR review comments on UI components by implementing the following changes:

- **Button Component Refactor:** Updated the `Export CSV` and `Import Candidates` actions in `ApplicationsList.tsx` to use the shared reusable `Button` component instead of raw `<button>` and styled `<Link>` tags, adhering to coding guidelines.
- **Color Token Migration:** Updated styling in `ApplicationStatusTimeline.tsx` to use the design system's approved color tokens. Replaced legacy `indigo` and `gray` tokens with `lime` for active/completed states and `stone` for neutral/inactive chrome. Updated hardcoded background hexes to `stone-950`.

## Changes Made
- Modified `client/src/module/recruiter/applications/ApplicationsList.tsx`
- Modified `client/src/module/student/applications/ApplicationStatusTimeline.tsx`

## Testing Instructions
1. Log in as a Recruiter and verify the applications list header renders the `Export CSV` and `Import Candidates` buttons correctly with appropriate hover states and transitions.
2. Log in as a Student and verify the application progress timeline displays correctly with the new `lime` / `stone` color scheme across all statuses (Applied, In Progress, Shortlisted, Hired, Rejected, Withdrawn).

## Summary by CodeRabbit

* **New Features**
  * Recruiter application lists now support exporting filtered results to CSV with a download button and loading state.
  * Student application progress now shows a clearer step-by-step timeline with support for rejected and withdrawn states.

* **UI Improvements**
  * Updated header actions to group import and export options more cleanly.


Sachinchaurasiya360 (changes requested):
Two issues here, one of which makes most of this PR a no-op:

1. **`archived/client/module/recruiter/applications/ApplicationsList.tsx` is a dead file.** The `archived/` directory was fully removed from main by commit `a0847fa2` ("remove the OSS Ambassador program") on 2026-06-28, and this branch was cut earlier the same day, before that removal. That file has no route, no import, and isn't bundled anywhere in `client/src`. Merging this today would resurrect ~69 of the 166 added lines (the Export CSV button, the Button-component migration) with zero effect — the recruiter feature they belong to no longer exists. Can you rebase onto current `main` and drop that file from the diff?

2. **`ApplicationStatusTimeline.tsx` shows misleading progress for rejected/withdrawn applications.** For `REJECTED`/`WITHDRAWN`, every step before the last renders as completed regardless of actual history:
   ```tsx
   if (isFailed) {
     if (idx === STEPS.length - 1) {...}
     else { icon = <Check .../>; ... }
   }
   ```
   Since the component only receives `status` (not the highest stage actually reached), a student rejected immediately after applying is shown a timeline implying they passed through "In Progress" and "Shortlisted." That's inaccurate data shown to students about their own application.

Minor (will also fail `npm run typecheck`, TS6133 under `noUnusedLocals`): drop the unused `import React from "react"` in the new file — this project uses the automatic JSX runtime. Also worth switching to the existing `getStatusColor`/`getStatusBorderColor` helpers in `client/src/lib/application-colors.ts` instead of re-deriving status colors inline, so they stay consistent with every other status badge in the app.

PR #2606, opened 2026-06-30
fix: prevent countdown timer drift after browser inactivity

sureshsuriya:
## Description

This PR fixes countdown timer drift in the Assessment, Skill Verification, and Aptitude test pages when the browser tab is backgrounded, minimized, or the device resumes from sleep.

Previously, the countdown timers relied on decrementing a local counter every second using `setInterval`. Since browsers throttle JavaScript timers in inactive tabs, the displayed remaining time could become inaccurate after the user returned to the page.

This implementation updates the countdown logic to calculate the remaining time using a fixed end timestamp and the current time (`Date.now()`), ensuring the timer remains synchronized with the actual test duration even after browser throttling or device sleep.

## Type of Change

* [x] Bug fix (non-breaking change that fixes an issue)
* [ ] New feature (non-breaking change that adds functionality)
* [ ] Breaking change (fix or feature that would cause existing functionality to change)
* [ ] Documentation update

## Related Issue

Fixes #2605

## Changes Made

* Updated countdown timer logic in:

  * `AssessmentTestView.tsx`
  * `SkillTestPage.tsx`
  * `AptitudeTopicPage.tsx`
  * `AptitudeCompaniesPage.tsx`
* Replaced interval-based decrement logic with an elapsed-time calculation using a fixed end timestamp.
* Preserved the existing timer behavior while preventing countdown drift caused by browser timer throttling.

## Testing

* [x] Verified the issue locally.
* [x] Tested timer behavior after backgrounding the browser tab.
* [x] Ran `npm run lint`.
* [x] Ran `npx tsc --noEmit`.
* [x] Ran `npm run build`.

## Checklist

* [x] My code follows the project's style guidelines.
* [x] I have performed a self-review of my changes.
* [x] I have tested the changes locally.
* [x] The changes are limited to the scope of this issue.

## Screenshots / Video

No UI changes. This PR updates the internal countdown timer logic to prevent timer drift after browser inactivity.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved countdown timers across student test pages to stay accurate even if the browser lags or the tab is inactive.
  * Timers now calculate remaining time from a fixed end point, reducing drift and preventing extra seconds from being shown.
  * Expiration behavior is now more reliable, consistently ending tests and question timers when time runs out.

### DIFF
--- a/client/src/module/student/applications/AssessmentTestView.tsx
+++ b/client/src/module/student/applications/AssessmentTestView.tsx
@@ -43,19 +43,27 @@ export function AssessmentTestView({
 
   // Countdown timer
   useEffect(() => {
-  if (secondsLeft === null || secondsLeft <= 0) return;
-  const interval = setInterval(() => {
-    setSecondsLeft((prev) => {
-      if (prev === null) return null;
-      if (prev <= 1) {
-        clearInterval(interval);
-        return 0;
-      }
-      return prev - 1;
-    });
-  }, 1000);
-  return () => clearInterval(interval);
-  // eslint-disable-next-line react-hooks/exhaustive-deps
+    if (secondsLeft === null || secondsLeft <= 0) return;
+    
+    const effectStartTime = Date.now();
+    let endTime: number | null = null;
+    
+    const interval = setInterval(() => {
+      setSecondsLeft((prev) => {
+        if (prev === null) return null;
+        if (endTime === null) {
+          endTime = effectStartTime + prev * 1000;
+        }
+        const remaining = Math.max(0, Math.ceil((endTime - Date.now()) / 1000));
+        if (remaining <= 0) {
+          clearInterval(interval);
+          return 0;
+        }
+        return remaining;
+      });
+    }, 1000);
+    return () => clearInterval(interval);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
 }, []); // runs once on mount; functional update inside setInterval handles secondsLeft
 
   // Auto-submit when timer hits 0
--- a/client/src/module/student/aptitude/AptitudeCompaniesPage.tsx
+++ b/client/src/module/student/aptitude/AptitudeCompaniesPage.tsx
@@ -94,18 +94,21 @@ export default function AptitudeCompaniesPage() {
     staleTime: 5 * 60 * 1000,
   });
 
+  const endTimeRef = useRef<number | null>(null);
+
   useEffect(() => {
-    if (!selectedCompany || !timerRunning || !companyData?.questions.length) return;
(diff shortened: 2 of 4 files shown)


Sachinchaurasiya360:
Thanks for tackling the timer drift, the `Date.now()`-anchored approach is the right fix. The `AssessmentTestView` and `SkillTestPage` conversions look correct and are robust to tab throttling.

One blocking issue in **`AptitudeTopicPage.tsx`** though: the `endTimeRef` regresses the per-page timer reset.

The `[page]` effect intentionally does `setTimeLeft(600)` on every page change so each page gets a fresh 10 minutes. In the old code the interval used a functional decrement (`t => t - 1`), so after the reset it just counted down from 600 regardless of whether the effect re-ran.

With `endTimeRef`, the anchor is only cleared in the guard branch (`!timerRunning || !topic?.questions.length`). When navigating to an **already-cached** page (staleTime is 5 min, so revisiting is instant), `topic` is defined immediately, `topic?.questions.length` stays `10`, and `timerRunning` stays `true`, so the effect's deps `[timerRunning, topic?.questions.length]` don't change and the effect doesn't re-run. `endTimeRef.current` keeps the *previous* page's end timestamp, so `setTimeLeft(600)` flashes for one render and then the next tick recomputes `remaining` from the stale ref, jumping to the leftover value from the prior page (or `0` if that page's time had expired) instead of restarting at `10:00`.

Suggested fix: drop the ref and use a local `const` inside the effect, exactly like `useCountdown` in `SkillTestPage.tsx` already does. The anchor only needs to live for the effect's lifetime, and re-anchoring on each effect run is the correct behavior (it also removes the `eslint-disable exhaustive-deps` and the stale-`timeLeft`-closure risk):

```tsx
useEffect(() => {
  if (!timerRunning || !topic?.questions.length) return;

  const endTime = Date.now() + timeLeft * 1000;

  const id = setInterval(() => {
    setTimeLeft(() => {
      const remaining = Math.max(0, Math.ceil((endTime - Date.now()) / 1000));
      if (remaining <= 0) {
        setTimerRunning(false);
        clearInterval(id);
        return 0;
      }
      return remaining;
    });
  }, 1000);
  return () => clearInterval(id);
  // eslint-disable-next-line react-hooks/exhaustive-deps
}, [timerRunning, topic?.questions.length]);
```

Note: the same `endTimeRef` pattern in `AptitudeCompaniesPage.tsx` is fine as-is. Its `[page]` effect only resets `currentQ`, not `timeLeft`, so that timer is intentionally continuous across pages and the persistent anchor is correct there. Please don't copy this fix to that file.

PR #2627, opened 2026-07-03
fix: route ATS try-free CTA to guest scorer without login

Honey-pg:
## Description

Add public `/ats-score/try` page with guest upload and scoring APIs so unauthenticated users can try the ATS checker without being sent to login.

**Problem:** The "Try free — no signup" CTA on the ATS landing page redirected unauthenticated users to `/login`, which contradicted the button copy and marketing ("No signup needed").

**Solution:**
- Route unauthenticated users from the ATS landing CTA to `/ats-score/try` instead of `/login`
- Add a public try page that reuses `AtsScorePage` in `guestMode`
- Add guest backend endpoints:
  - `POST /api/upload/guest-presigned-url` — PDF upload to `guest-resumes/{ipHash}/`
  - `POST /api/ats/guest/score` — ATS scoring without auth (2/day IP rate limit)
- Guest scores are returned in-memory only (not persisted to DB)
- Authenticated students still go to `/student/ats/score` as before

## Related Issue

Fixes #2603 

## Type of Change

- [x] Bug Fix  
- [ ] Feature  
- [ ] Enhancement  
- [ ] Documentation  

## Testing

**Automated**
- [x] `npm test -- src/module/ats/__tests__/ats.service.test.ts` — guest URL prefix validation + no DB persist
- [x] Client typecheck passes (`npx tsc --noEmit`)

**Manual**
1. Log out and open `/ats-score`
2. Click **Try free — no signup**
3. Confirm redirect to `/ats-score/try` (not `/login`)
4. Upload a PDF resume and run analysis
5. Confirm score, breakdown, keywords, and suggestions render
6. Confirm signup CTA appears after results
7. Log in as a student, revisit `/ats-score`, click the same CTA
8. Confirm redirect to `/student/ats/score`

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added a public ATS “try” page at `/ats-score/try` that runs ATS scoring in guest mode.
  * Introduced guest resume upload + guest scoring endpoints, including rate limiting and daily guest usage tracking.
* **Bug Fixes**
  * Updated guest “try now” navigation to route to `/ats-score/try`.
  * Improved guest resume validation and error/status handling for invalid, missing, or foreign resumes.
* **UI/UX**
  * Guest mode now uses guest-specific upload/scoring, applies guest-aware limits/CTAs, hides ATS tools/history, makes suggestions read-only, and adds a post-analysis guest sign-up banner.
  * History is available only in non-guest mode, with search filtering.
* **Tests**
  * Added automated tests covering guest scoring success and failure scenarios.


Sachinchaurasiya360 (changes requested):
Nice work on the guest flow overall — the IDOR/SSRF boundary is solid (`getS3KeyFromUrl` anchors to the real bucket host and the prefix check in `scoreResumeGuest` is correctly tested), file type/size is enforced server-side (schema + S3 POST policy `content-length-range`), and path traversal is closed off by `cleanFileName`. Removing the redundant `validateBody` calls in `upload.routes.ts` is fine since the controller's `isValidS3FileUrl` check is stricter and still in place.

Two things to fix before merging, both around the new `/api/ats/guest/score` endpoint being the first **unauthenticated** route that triggers a paid AI-provider call:

1. **The 2/day rate limit won't survive Vercel's serverless model.** `guestScoreRateLimit` in `ats.routes.ts` uses `express-rate-limit`'s default in-memory store — a `Map` scoped to a single lambda instance. Per this repo's own constraint ("no durable disk and no long-lived processes"), that counter resets on every cold start, and Vercel spins up fresh instances constantly under normal traffic, not just during deploys. An attacker (or just a curious user) spacing out requests gets effectively unlimited free scoring, which is unlimited billed AI calls. I know `contactLimiter`/`authIpLimiter`/etc. have this same limitation, but those don't cost money per request — this one does, and it's the first endpoint in the app with that combination (no auth + metered AI call). Please back this with something that survives across invocations — even a lightweight Postgres/Prisma-backed counter (a single upsert-and-check keyed by ipHash + day, no transaction needed for a best-effort limiter) would close this, or an external store if one's already available.

2. **`guest-ip.utils.ts`'s `getClientIp` bypasses the app's trust-proxy setup and hand-parses `X-Forwarded-For`.** `index.ts` already does `app.set("trust proxy", 1)`, which is what makes Express's own `req.ip` (and `express-rate-limit`'s default key resolution) trustworthy. This function instead reads the raw header and takes the first comma-separated value, which a client can set to anything. It doesn't break the numeric rate limit today (that's keyed on `req.ip` independently), but it makes `ipHash`/the S3 prefix a forgeable identity, and it's easy for future code to start trusting `ipHash` for more than it should. Please just derive it from `req.ip` instead of re-parsing the header.

Happy to take another look once these are addressed.

Sachinchaurasiya360 (changes requested):
Thanks for iterating on this — the guest-scoring security fixes are solid now:

- The daily rate limit is backed by a real `guestUsage` Postgres table with a compound unique constraint on `(ipHash, date)`, and the increment is a single atomic `upsert` (`INSERT ... ON CONFLICT DO UPDATE`) — no TOCTOU race, and it survives cold starts (the earlier in-memory `express-rate-limit` store didn't).
- `getClientIp` now uses `req.ip` (respecting this app's `trust proxy` setting) instead of parsing a spoofable raw `X-Forwarded-For` header.
- Migration for the new table is included.

One thing to fix before merge: this PR also adds a "Score over time" history panel to the authenticated `AtsScorePage` that calls `GET /api/ats/history` — that route doesn't exist anywhere in `server/src/module/ats/ats.routes.ts`, and `AtsService.scoreResume` is explicitly stateless (scores aren't persisted). Every authenticated visit to `/student/ats/score` will 404 on this call. Since building score history isn't in scope for "fix the guest try-free CTA," can you remove the history panel from this PR? Happy to merge right after.

Two minor, non-blocking nits for a follow-up:
- The rate-limit counter in `guest-usage-limit.middleware.ts` increments before request validation/PDF extraction, so a guest whose upload fails validation still burns one of their 2 daily attempts.
- `ats.validation.ts`'s guest schema checks `normalized.includes(prefix)` rather than `.startsWith()` — not currently exploitable since `ats.service.ts` independently enforces `startsWith` on the same URL before the S3 fetch, but worth tightening to match.

Sachinchaurasiya360 (changes requested):
Good progress on the previously-requested fixes:
- The daily guest limit is now backed by the durable `guestUsage` Postgres table with an atomic upsert — survives cold starts, no TOCTOU.
- `getClientIp` now derives from `req.ip` (trust-proxy-respecting) instead of hand-parsing `X-Forwarded-For`.
- `ats.validation.ts` now uses `.startsWith()` consistently.
- The out-of-scope "Score over time" history panel has been removed from the page.

But the latest commit (`66a0166f` "remove score history panel and resolve rate-limit & validation nits") introduced two new bugs that break the build — this is why Typecheck Server and Typecheck Client are both failing on this PR right now, not pre-existing main flakiness:

1. **`server/src/module/upload/upload.routes.ts` — duplicate import.** A new import block was added at the top of the file (`import { validateBody, presignRequestSchema, guestPresignRequestSchema } from "./upload.validation.js";`), but the original import statement further down the file (`import { validateBody, presignRequestSchema, uploadProfilePicSchema, ... }`) was never removed. `validateBody` and `presignRequestSchema` are now each imported twice from the same module, which TypeScript rejects as a duplicate identifier. Merge the two into a single import statement.

2. **`client/src/module/student/ats/AtsScorePage.tsx` — dangling reference to a removed query key.** The `onSuccess` handler still calls `queryClient.invalidateQueries({ queryKey: queryKeys.ats.history() })` (guarded by `if (!guestMode)`), but `queryKeys.ats` on `main` only has `all` and `usage` — `history` doesn't exist. This looks like a leftover from the history-panel removal; since the panel is gone, this whole `invalidateQueries` call (and the `if (!guestMode)` block wrapping it) should just be deleted.

Once these two are fixed and the build is green, this looks ready to merge — the security-sensitive parts (guest rate limiting, IP handling, S3 prefix scoping) are solid.

PR #2630, opened 2026-07-03
fix: send admin email notification on contact form submission

Akshita-2307:
Contact form submissions were silently stored in the database with no notification to the site admin. This adds a sendEmail call to the admin alert address after a successful submission, using the existing sendEmail utility which gracefully degrades in dev when RESEND_API_KEY is not set.\n\nCloses #2625

### DIFF
--- a/server/src/module/contact/contact.routes.ts
+++ b/server/src/module/contact/contact.routes.ts
@@ -3,6 +3,9 @@ import type { Request, Response } from "express";
 import { prisma } from "../../database/db.js";
 import { contactSchema } from "./contact.validation.js";
 import { contactLimiter } from "../../middleware/rate-limit.middleware.js";
+import { sendEmail } from "../../utils/email.utils.js";
+
+const ADMIN_ALERT_EMAIL = process.env["ADMIN_ALERT_EMAIL"] ?? "mrsachinchaurasiya@gmail.com";
 
 export const contactRouter = Router();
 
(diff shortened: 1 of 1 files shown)


Sachinchaurasiya360 (changes requested):
Good fix for the underlying problem (admin never got notified), but two things should be addressed before merging:

1. **Unhandled rejection risk**: `sendEmail(...)` at line ~19 is fire-and-forget with no `.catch()`. `sendEmail` throws on failure (see `email.utils.ts`), so a Resend error (bad key, rate limit, invalid `to`) becomes an unhandled promise rejection instead of a logged failure. The codebase already has the right pattern for this exact situation in `error.middleware.ts`:
   ```ts
   sendEmail({...}).catch((e) => console.error("[ContactForm] Admin alert email failed:", e));
   ```
   Please mirror that here so a failed admin notification is actually observable instead of silently disappearing (and on Vercel, the function may freeze before the rejection is even processed).

2. **Unescaped user input in email HTML**: `parsed.data.name`, `parsed.data.subject`, and `parsed.data.message` are interpolated directly into the HTML body with no escaping, and `contactSchema` only validates length/format, not content. A submitter can embed arbitrary HTML/links (e.g. a fake "verify your account" link) that renders live in the admin's inbox. Please HTML-escape these three fields before interpolating them into the `html` string (or build the body with a templating helper that escapes by default).

Happy to re-review once these are in.

Sachinchaurasiya360 (changes requested):
Good fix for the missing admin notification, two issues before this is mergeable:

1. **`sendEmail(...)` is fire-and-forget with no `.catch()`.** `sendEmail` throws on failure (`server/src/utils/email.utils.ts`). This codebase already has the correct pattern for the identical situation in `server/src/middleware/error.middleware.ts`:
   ```ts
   sendEmail({...}).catch((e) => console.error("[ErrorLog] Admin alert email failed:", e));
   ```
   Without it, a Resend failure (bad key, rate limit) becomes an unhandled promise rejection with no context on which submission failed — silently defeating the point of this PR (admin never finds out a notification failed).

2. **Unsanitized user input interpolated into email HTML.** `name`, `subject`, and `message` are spliced raw into the HTML body sent to the admin's inbox (`contactSchema` only validates length/email format, no escaping). A submitter could set `message` to `<a href="...">Verify your account</a>` and it renders as a live, styled link in your inbox — a phishing/HTML-injection vector aimed at the admin.

Can you add the `.catch()` and escape the three interpolated fields (or switch to a template helper that escapes by default)? Happy to re-review once updated.

PR #2631, opened 2026-07-03
perf(roadmap): optimize streak computation to O(1) using state-machine logic

### LINKED ISSUE #2309: perf(db): optimize roadmap streak recomputation after topic completion
## Goal
`updateEnrollmentStreak` fetches every completed topic for an enrollment and recalculates streaks in memory on each completion. This grows with learner history.

## Scope
- server/src/module/roadmap/roadmap.service.ts
- server/src/database/prisma/schema/roadmap.prisma

## Acceptance Criteria
- Recompute streak incrementally or with a bounded date-window query where possible.
- Add or verify an index for completed progress lookups by enrollment/status/completedAt.
- Preserve currentStreak, bestStreak, weeklyStreak behavior.
- Add tests for same-day completion, consecutive days, missed days, and reset cases.

Rajal-ui:
## Description
Optimized the `updateEnrollmentStreak` function in the roadmap module to completely eliminate expensive database lookups. The streak computation has been refactored from an in-memory recalculation of the entire topic completion history down to a pure $O(1)$ state-machine pattern.

Since this function is invoked directly following a successful topic completion tracking write, the user's `lastStreakDate` column serves as the single source of truth to check for same-day completions, consecutive days, or streak breaks.

## Related Issue
Fixes #2309 

## Type of Change
- [x] Enhancement / Performance Optimization
- [x] Refactoring

## Testing Done
- **Unit Tests:** Updated and expanded the test suite in `roadmap.service.test.ts` to 9 comprehensive cases (all 44 roadmap module tests are passing).
- Tested cases include:
  - Same-day early return (returns immediately without hitting the DB update).
  - Consecutive day increment (`currentStreak` increments cleanly).
  - Missed day reset (`currentStreak` drops back to 1 on a gap).
  - `bestStreak` tracking (properly updates or preserves the peak record across reset cycles).
  - Enrollment-not-found guard.
  - Weekly streak milestones.

## Checklist
- [x] Tested manually and via automated suites
- [x] No unintended database migrations or schema drift committed

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved streak tracking so same-day completions no longer change saved progress.
  * Made streak updates more reliable across consecutive, missed-day, and weekly timing scenarios.
  * Preserved the highest streak record when a streak resets and starts again.
  * Prevented unnecessary updates when enrollment data is unavailable.

### DIFF
--- a/server/src/__tests__/roadmap.service.test.ts
+++ b/server/src/__tests__/roadmap.service.test.ts
@@ -32,6 +32,7 @@ vi.mock("../database/db.js", () => ({
     roadmapTopic: { findFirst: vi.fn() },
     roadmapTopicProgress: {
       upsert: vi.fn(),
+      findFirst: vi.fn(),
       findMany: vi.fn(),
       update: vi.fn(),
       count: vi.fn(),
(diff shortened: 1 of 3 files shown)


Sachinchaurasiya360 (approved):
Reviewed the diff on `updateEnrollmentStreak`. The functional change is correct: the same-day early return (skips the now-unneeded `roadmapEnrollment.update` write) and consolidating `bestStreak = Math.max(bestStreak, currentStreak)` to run once after the if/else are both behaviorally equivalent to the prior per-branch logic — traced all three branches (`!lastStreakDate`, consecutive day, reset) against the updated tests, all consistent. Day-key helpers (`toUtcDayKey`/`addUtcDays`) are unchanged UTC-based logic, no timezone regression. The lone caller doesn't use the return value, so the new early `return` is safe.

One nit, non-blocking: the new comment on `roadmap.prisma`'s `@@index([enrollmentId, status, completedAt])` — `// Used by updateEnrollmentStreak bounded window query` — is inaccurate. `updateEnrollmentStreak` only reads/writes `roadmapEnrollment` by primary key; it doesn't query `roadmapTopicProgress` at all (before or after this PR). That index is actually used by the weekly-digest/analytics queries elsewhere in `roadmap.service.ts`. Worth a follow-up comment fix so a future reader doesn't mis-attribute (and potentially remove) that index.

Merging — the core change is a legitimate small optimization and the test coverage is solid.

PR #2633, opened 2026-07-03
fix: harden LaTeX compilation and supporting file handling

SnehalTanpure7:
# Pull Request

## Description
Describe your changes.

## Related Issue
Fixes #2604

## Type of Change
- [ ] Bug Fix  
- [ ] Feature  
- [ ] Enhancement  
- [ ] Documentation  

## Testing
Explain how you tested it.

## Screenshots / Video

_No UI changes in this PR_ (delete this line if you are making UI changes)

## Checklist
- [ ] Code follows project guidelines
- [ ] No new compile/type errors
- [ ] Tested manually (include steps above)
- [ ] No `.env`, credentials, or `node_modules` committed

## Summary by CodeRabbit

* **Bug Fixes**
  * LaTeX compilation now rejects unsafe source content before running, helping prevent failed or risky submissions.
  * Supporting files are validated more strictly, including filename rules, allowed file types, duplicate names, and size limits.
  * Large uploads are now blocked earlier, with limits applied to the main LaTeX source, each file, and the total attachment bundle.
  * Validation is enforced consistently across both local and online compilation paths.

### DIFF
--- a/server/src/module/latex/latex.service.ts
+++ b/server/src/module/latex/latex.service.ts
@@ -7,6 +7,21 @@ import crypto from "crypto";
 const COMPILE_TIMEOUT = 30_000; // 30 seconds
 const ONLINE_API = "https://latex.ytotech.com/builds/sync";
 
+const MAX_SOURCE_SIZE = 500 * 1024; // 500 KB
+const MAX_FILE_SIZE = 2 * 1024 * 1024; // 2 MB per supporting file
+const MAX_TOTAL_SIZE = 10 * 1024 * 1024; // 10 MB total
+
+const SAFE_FILENAME_REGEX = /^[a-zA-Z0-9][a-zA-Z0-9._-]{0,99}$/;
+const ALLOWED_EXTENSIONS = new Set([".sty", ".cls", ".bib", ".bst", ".png", ".jpg", ".jpeg", ".eps"]);
+
+const DANGEROUS_PATTERNS: RegExp[] = [
+  /\\write18/,
+  /\\immediate\s*\\write18/,
+  /\\input\s*\{\s*\|/,
+  /\\openin/,
+  /\\catcode.*\\active/,
+  /\\csname\s*shellescape/i,
+];
 export interface SupportingFile {
   filename: string;
   content: string;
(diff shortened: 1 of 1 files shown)


Sachinchaurasiya360 (changes requested):
Reasonable direction (pre-flight size caps, filename allowlist, path-traversal check, shell-escape blocklist), but a few issues before merge:

1. **Regression: `.tex` supporting files are now rejected.** `ALLOWED_EXTENSIONS` omits `.tex`, but `latex.validation.ts` explicitly whitelists `.tex` as a valid supporting-file extension for multi-file documents (e.g. `\input{chapter.tex}`). Any request that passes the existing zod schema with a `.tex` supporting file now throws "Unsupported file type" in the service — breaking a previously-supported, still-documented use case.

2. **New rejection messages aren't handled anywhere and surface as false server errors.** None of the new thrown messages are mapped in `latex.controller.ts` or `error.middleware.ts`; they fall through to the generic 500 branch, which also fires an admin alert email for every one. So a legitimate validation rejection (including the `.tex` regression above, or a false-positive blocklist hit) shows students an opaque "Internal Server Error" and spams the admin inbox as a false server-error alert.

3. **The shell-escape blocklist is bypassable and overbroad.** TeX ignores whitespace after control words, so `\write 18` (with a space) trivially evades the `\write18` regex; meanwhile `\catcode.*\active` and `\openin` will false-positive on legitimate constructs (babel-style active characters, file-reading packages). Local compilation is already hard-protected via `-no-shell-escape`, so this blocklist mainly matters for the online API path — exactly where it's easiest to bypass.

4. The new size caps (`MAX_SOURCE_SIZE`/`MAX_FILE_SIZE`/`MAX_TOTAL_SIZE`) are all looser than the existing zod limits in `latex.validation.ts` (200KB source, 200KB/file × 5 files), and `LatexService.compile` is only ever reached after that schema already validates — so these checks are unreachable dead code duplicating validation that already exists.

Note there's a competing PR (#2642) open for the same issue with a more thorough approach (per-job temp directory isolation to fix a real concurrent-job file collision, which this PR doesn't address) — worth comparing before investing more time here.

Sachinchaurasiya360:
Closing this in favor of #2642, which addresses the same issue and additionally fixes a real concurrency bug (concurrent compile jobs writing supporting files into the shared `os.tmpdir()` by raw filename, so they could collide/overwrite each other) that this PR doesn't touch. Both PRs currently need changes before merge (see review comments on each), but consolidating effort on #2642 makes more sense than carrying two competing fixes for the same issue forward. Thanks for the contribution — the size-cap and filename-validation ideas here were good instincts.

PR #2642, opened 2026-07-03
fix: harden LaTeX compilation and supporting file handling

vansh2604-star:
# Pull Request

## Description

This PR hardens the LaTeX resume compilation process by improving validation for LaTeX source and supporting files.

### Changes made
- Added validation to block unsafe LaTeX commands.
- Prevented path traversal attacks in supporting file names.
- Restricted supporting files to safe filenames only.
- Added file size limits for uploaded supporting files.
- Rejected duplicate supporting file names.
- Ensured supporting files are written only inside the compilation directory.
- Improved validation error messages for invalid uploads.

These changes improve the security of server-side PDF generation while maintaining compatibility with valid resume templates.

## Related Issue

Fixes #2604

## Type of Change

- [x] Bug Fix
- [ ] Feature
- [ ] Enhancement
- [ ] Documentation

## Testing

Tested manually by validating the following scenarios:

- ✅ Valid LaTeX resumes compile successfully.
- ✅ Supporting files with path traversal names (e.g. `../evil.sty`) are rejected.
- ✅ Duplicate supporting file names are rejected.
- ✅ Oversized supporting files are rejected.
- ✅ Unsafe LaTeX commands are blocked.
- ✅ Valid supporting files continue to work correctly.

## Screenshots / Video

No UI changes in this PR.

## Checklist

- [x] Screenshot or video attached for every UI change (Not applicable - no UI changes)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved LaTeX upload and compilation handling with stricter validation for file names, file size, and document content.
  * Added better protection against invalid or unsafe inputs, helping prevent failed or unexpected compilations.
  * Made compilation cleanup and error reporting more reliable, including clearer timeout behavior.

### DIFF
--- a/server/src/module/latex/latex.service.ts
+++ b/server/src/module/latex/latex.service.ts
@@ -1,11 +1,22 @@
 import { spawn } from "child_process";
-import { writeFile, readFile, unlink } from "fs/promises";
+import {
+  writeFile,
+  readFile,
+  mkdtemp,
+  rm,
+} from "fs/promises";
 import path from "path";
 import os from "os";
-import crypto from "crypto";
 
-const COMPILE_TIMEOUT = 30_000; // 30 seconds
+const COMPILE_TIMEOUT = 30_000;
 const ONLINE_API = "https://latex.ytotech.com/builds/sync";
+
+const MAX_SUPPORTING_FILES = 5;
+const MAX_TOTAL_SUPPORT_SIZE = 500_000;
+const MAX_PDF_SIZE = 10 * 1024 * 1024;
+
+const FILENAME_REGEX =
+  /^[A-Za-z0-9][A-Za-z0-9_.-]{0,99}\.(cls|sty|bst|bib|tex)$/i;
 
 export interface SupportingFile {
   filename: string;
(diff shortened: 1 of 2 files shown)


Sachinchaurasiya360 (changes requested):
The per-job `mkdtemp` isolation is a genuine, valuable fix (the prior code wrote supporting files straight into the shared `os.tmpdir()` by raw filename, so concurrent compile jobs could collide/overwrite each other). A few issues before merge:

1. **The command blocklist breaks the `.tex` supporting-file type it's meant to protect.** `UNSAFE_COMMANDS` unconditionally bans `\input\b`/`\include\b` regardless of argument, while `UNSAFE_PATHS` more carefully only blocks them when the argument contains `..`/`/`/`\`. Net effect: `\input`/`\include` are banned outright, so a legitimate multi-file resume (`main.tex` + `\input{sections}`) — a case this PR's own `filenameSchema` still allows — now 400s as "unsafe commands."

2. **Online-compile error detection regresses.** `compileOnline()`'s error messages no longer contain the substring `"compilation failed"` (now capitalized, or a bare log excerpt, or `"Server returned X"`). `latex.controller.ts`'s case-sensitive `.includes("compilation failed")` check stops matching, so a bad document compiled via the online fallback (likely the actual path in Vercel prod, where local `pdflatex` probably isn't installed) surfaces as an unhandled 500 instead of the intended 422 "fix your LaTeX" response.

3. `MAX_PDF_SIZE` is only enforced in `compileLocal` — `compileOnline` returns the fetched buffer with no size cap, so the resource-exhaustion mitigation doesn't cover the path that likely matters most in prod.

4. Minor: the new size/count/filename constants in `latex.service.ts` and `latex.validation.ts` duplicate each other (with an inconsistent `/i` flag on one copy), and roughly half the diff is reformatting unrelated to the fix — makes this harder to review than it needs to be.

Note there's a competing PR (#2633) open for the same issue — worth checking that one doesn't already cover part of this so effort isn't duplicated.

Sachinchaurasiya360:
Solid security hardening. The per-compile `mkdtemp` working dir, path-traversal checks (`path.relative` + `..`/absolute guard), filename allowlist, total-size caps, and the PDF size limit are all good improvements, as is `-halt-on-error` + `cwd` instead of `-output-directory`.

One concern before merging: the `UNSAFE_COMMANDS` regex rejects `\input`, `\include`, `\csname`, `\loop`, `\repeat`, `\catcode`, etc. in the **source** document (not just supporting files). Some legitimate resume/LaTeX templates use these, so this risks rejecting valid compiles. Could we either:

1. run a compile smoke-test confirming our own generated/templated resumes still pass, or
2. narrow the source blocklist to the genuinely dangerous primitives (`\write18`, shell-escape, `\input`/`\include`/`\openin`/`\openout` with paths)?

Happy to merge once we've confirmed real resumes aren't caught by the filter.

Sachinchaurasiya360 (changes requested):
## Requesting changes: the command blocklist will reject legitimate LaTeX, and one error path is unreachable

The core hardening here is genuinely good and I want it to land. Per-job `mkdtemp` instead of writing attacker-named files into the shared `os.tmpdir()` closes a real cross-job collision, `rm(workDir, { recursive: true, force: true })` is more reliable than enumerating extensions, the path-traversal check is correct, and `-halt-on-error` is a good addition. Two things block it.

### 1. `UNSAFE_COMMANDS` will break real `.cls` and `.sty` files (blocking)

```js
const UNSAFE_COMMANDS =
  /\\(?:write18|immediate|shellescape|openout|openin|write\b|read\b|readline\b|input\b|include\b|includeonly\b|catcode\b|csname\b|newread\b|newwrite\b|loop\b|repeat\b|fileinput\b|verbatiminput\b|CatchFileEdef)/i;
```

and it is applied to supporting file content:

```js
content: z.string().min(1).max(MAX_SUPPORT_FILE_SIZE).refine(isSafeLatex, {
  message: "Supporting file contains unsafe LaTeX content",
}),
```

`\catcode`, `\csname`, `\loop`, `\repeat`, `\immediate`, `\write` and `\newwrite` are ordinary building blocks of document classes and bibliography styles. Essentially any real `.cls` or `.bst` a user uploads will be rejected. Since the point of the supporting-files feature is to let people bring their own class file, this disables the feature.

The same refine runs on `source`, where `\input` and `\include` are how multi-file resumes are structured, and `\csname` appears in plenty of resume templates.

The genuinely dangerous ones are the shell-escape and file-write primitives: `\write18`, `\shellescape`, `\openout`, `\openin`, `\input`/`\include` pointing outside the work directory. Those are already handled better by mechanisms you have:

- `-no-shell-escape` on the `spawn` call blocks `\write18`
- `openout_any: "p"` restricts output file creation
- `UNSAFE_PATHS` already catches `\input`/`\include` with `..` or path separators
- the per-job `mkdtemp` plus traversal check bounds what can be written

Please narrow `UNSAFE_COMMANDS` to the shell-escape family (`write18`, `shellescape`, and arguably `openout`/`openin` used with explicit paths) and drop the typesetting primitives.

Relatedly, please keep the comment that was deleted from the `env` block:

```js
// openin_any is a no-op in TeX Live 2026+, so read isolation must be
// enforced at the container/sandbox/OS level in production.
```

That is important operational context for whoever reviews this next, and the PR removes it while leaving `openin_any` unset.

### 2. The `JSON.parse` fallback is unreachable

```js
try {
  const parsed = JSON.parse(text);
  ...
  throw new Error(parsed.error ?? "Compilation failed");
} catch (e) {
  if (e instanceof Error) throw e;
  throw new Error(`Server returned ${response.status}`);
}
```

When the API returns non-JSON (an HTML 502 page, for example), `JSON.parse` throws a `SyntaxError`. `SyntaxError instanceof Error` is `true`, so `throw e` rethrows it and the caller sees `Unexpected token < in JSON at position 0` instead of `Server returned 502`. The `Server returned ${status}` line can never run.

The previous code guarded on the message to distinguish its own throws from parse failures. A cleaner version parses outside the try:

```js
let parsed: any = null;
try { parsed = JSON.parse(text); } catch { /* not JSON */ }

if (!parsed) throw new Error(`LaTeX compilation failed: server returned ${response.status}`);
// ... then handle parsed.log_files / parsed.error
```

### Smaller points

- Error messages lost the `LaTeX compilation failed:` prefix in the online path, and lost the `error style` exclusion plus the `.replace(/^!\s*/, "").trim()` cleanup. Those made messages readable for users. Worth keeping.
- `RESERVED_FILENAMES` lists `output.pdf` and `compile.log`, but neither can match the extension regex (`.pdf` and `.log` are not allowed), so those two entries are dead. Blocking `resume.tex` also seems arbitrary. `main.tex` is the one that matters, since the service writes the source there.
- The diff re-wraps nearly every call site to narrower lines, which makes a security-relevant change much harder to review. Splitting formatting from behaviour would help a lot next time.

PR #2649, opened 2026-07-04
fix: prevent email enumeration by returning identical response shape for all register outcomes

Akshita-2307:
This PR addresses #2644.

## Summary by CodeRabbit

* **Bug Fixes**
  * Registration now returns a simpler success response without including user details.
  * Existing validation and error messages for duplicate email and server errors remain unchanged.

### DIFF
--- a/server/src/module/auth/auth.controller.ts
+++ b/server/src/module/auth/auth.controller.ts
@@ -34,8 +34,8 @@ export class AuthController {
       // Body is already validated & typed by route-level validateBody(registerSchema)
       const input = req.body as z.infer<typeof registerSchema>;
 
-      const data = await this.authService.register(input);
-      return res.status(201).json({ message: "Registration successful. Please verify your email to continue.", user: data.user });
+      await this.authService.register(input);
+      return res.status(201).json({ message: "Registration successful. Please verify your email to continue." });
     } catch (error) {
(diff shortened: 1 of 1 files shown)


Sachinchaurasiya360 (changes requested):
Good instinct (email enumeration is a real issue here), but this change as-is breaks registration for real users.

**Breaks the client (critical).** `client/src/module/auth/RegisterPage.tsx` does:
```ts
if (!data.user.isVerified) { navigate(verify-email) } else { login(data.user) }
```
After this PR, a successful registration returns `{ message }` only — `data.user` is `undefined`, so `data.user.isVerified` throws. The client's catch block shows a generic "Registration failed" toast to a user whose account was actually just created, and they're never routed to `/verify-email`. This regresses the primary happy path for every new signup.

**Timing side-channel still open.** Response bodies/status now match (201, `{message}`), but the success path still does `bcrypt.hash` (10 rounds) + `user.create` + an OTP update + slug generation before responding, while the duplicate-email path returns immediately. Latency still leaks which outcome occurred, so this doesn't fully close the enumeration vector it's meant to fix.

To fix both: keep returning enough info for the client to route correctly (e.g. always include a `user` shape, or have the client key off `message` instead of `user.isVerified`), and equalize timing between the two paths (e.g. a dummy hash/no-op of comparable cost on the duplicate-email path).

PR #2673, opened 2026-07-07
fix(job-index): add console.error logging to semanticSearch catch block

Xenon010101:
Add console.error logging to semanticSearch catch block so operators can see database failures.

Fixes #2667

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved search error handling so failures are now logged before returning no results, making issues easier to diagnose.
  * Search behavior still returns an empty result when an error occurs, preserving the existing user experience.

### DIFF
--- a/server/src/module/job-index/job-index.service.ts
+++ b/server/src/module/job-index/job-index.service.ts
@@ -339,8 +339,8 @@ export class JobIndexService {
         limit,
       );
       return jobs;
-    } catch {
-      // pgvector / embedding column not set up yet, fall back to empty
+    } catch (err) {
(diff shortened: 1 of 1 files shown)


Sachinchaurasiya360 (approved):
LGTM. Straightforward observability fix: the catch previously swallowed all failures, including real DB errors, making semantic search silently degrade to keyword-only. One note: if pgvector/embeddings are intentionally not provisioned in an environment, this will log an error on every search call there (the old comment implied that was an expected state). Acceptable noise for the visibility gained, and prod has pgvector set up.

PR #2674, opened 2026-07-07
fix(opensource): add console.error logging to empty email catch blocks

Xenon010101:
Add console.error logging to two empty email catch blocks that silently swallowed send failures.

Fixes #2668

### DIFF
--- a/server/src/module/opensource/opensource.service.ts
+++ b/server/src/module/opensource/opensource.service.ts
@@ -393,8 +393,8 @@ export class OpensourceService {
           request.owner,
         ),
       });
-    } catch {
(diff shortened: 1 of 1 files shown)


Sachinchaurasiya360 (approved):
LGTM, merging. Same rationale as #2673: silent email failures in requestRepo/approveRepo are now visible to operators while staying non-blocking (the catch still swallows, request/approval flow is unaffected).

One nit (not blocking): the two edited `} catch (err) {` lines are indented 6 spaces instead of 4, so the braces are misaligned with their `try`. Will get cleaned up by the next formatter pass on this file.

PR #2680, opened 2026-07-08
fix: prevent email enumeration by returning identical response for all register outcomes

Akshita-2307:
Fixes #2644

Returns an identical response shape (`{ message }` without `user`) for both successful registration and duplicate email, so attackers cannot enumerate registered emails by checking for the `user` field.

## Summary by CodeRabbit

* **Bug Fixes**
  * Registration responses now return a consistent message-only payload, avoiding exposure of the newly created user details.
  * The “email already registered” response remains unchanged and now matches the success response format.

### DIFF
--- a/server/src/module/auth/auth.controller.ts
+++ b/server/src/module/auth/auth.controller.ts
@@ -35,7 +35,7 @@ export class AuthController {
       const input = req.body as z.infer<typeof registerSchema>;
 
       const data = await this.authService.register(input);
-      return res.status(201).json({ message: "Registration successful. Please verify your email to continue.", user: data.user });
+      return res.status(201).json({ message: "Registration successful. Please verify your email to continue." });
     } catch (error) {
       if (error instanceof Error && error.message === "Email already registered") {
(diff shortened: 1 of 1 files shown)


Sachinchaurasiya360 (changes requested):
This is the same change as your open PR #2649, and both blockers from the review there still apply verbatim here:

1. **Breaks registration for every new user.** `client/src/module/auth/RegisterPage.tsx` (lines 268-272) does `data.user.isVerified` on the register response. After this change `data.user` is `undefined`, so the happy path throws, the user sees a generic "Registration failed" toast, and they are never routed to `/verify-email`, even though their account was created.
2. **Timing side-channel remains open.** The success path still runs `bcrypt.hash` + `user.create` + OTP work before responding while the duplicate-email path returns immediately, so response latency still leaks which outcome occurred.

Please consolidate into a single PR (close whichever of #2649/#2680 you're not continuing) that also updates the client routing (e.g. key off the response message or always return a minimal `user` shape) and equalizes timing between the two paths. See the full review on #2649 for details.

PR #2681, opened 2026-07-08
fix: remove hardcoded admin email fallback in error middleware and expert-session service

Akshita-2307:
Fixes #2658

Removes the personal email fallback (`mrsachinchaurasiya@gmail.com`) from both `error.middleware.ts` and `expert-session.service.ts`. Both now use `process.env["ADMIN_ALERT_EMAIL"] ?? ""` so the send will fail explicitly if the env var is not set, instead of silently leaking the maintainer's email.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated admin alert email handling to use a blank default when no email address is configured, avoiding unintended delivery to a fallback address.
  * Improved consistency in booking confirmation notifications by relying on the configured email setting only.

### DIFF
--- a/server/src/middleware/error.middleware.ts
+++ b/server/src/middleware/error.middleware.ts
@@ -4,7 +4,7 @@ import { prisma } from "../database/db.js";
 import { sendEmail } from "../utils/email.utils.js";
 import { FileUploadError } from "../lib/errors.js";
 
-const ADMIN_ALERT_EMAIL = process.env["ADMIN_ALERT_EMAIL"] ?? "mrsachinchaurasiya@gmail.com";
+const ADMIN_ALERT_EMAIL = process.env["ADMIN_ALERT_EMAIL"] ?? "";
 
 const SENSITIVE_KEYS = new Set([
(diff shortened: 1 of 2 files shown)


Sachinchaurasiya360 (approved):
Approving. Removing the maintainer's personal address from the public repo is the right call, and both call sites are fire-and-forget with a `.catch` (error.middleware.ts and expert-session.service.ts), so an empty recipient can't crash anything, it just logs.

Operational note: `ADMIN_ALERT_EMAIL` was not set in the Vercel production env, so as-is this would have silently disabled 500-error alerts and booking notifications on the next deploy. I've added `ADMIN_ALERT_EMAIL` to the Vercel Production and Preview environments (same address as the old fallback), so behavior is unchanged after deploy.

Follow-up (out of scope here): `server/src/module/email-inbound/email-inbound.controller.ts` line 6 still hardcodes the same personal address as the `INBOUND_FORWARD_TO` fallback. That env var IS set in prod so there's no functional impact, but the address still leaks in the repo. Worth the same treatment in a follow-up PR.

PR #2682, opened 2026-07-08
fix: make VERIFICATION_SECRET throw on missing env var instead of falling back to JWT_SECRET

Akshita-2307:
Fixes #2659

Changes `VERIFICATION_SECRET` to throw immediately if the env var is not set, matching the `JWT_SECRET` pattern. This prevents silent token-type isolation breakage where verification tokens could be confused with auth tokens.

### DIFF
--- a/server/src/utils/jwt.utils.ts
+++ b/server/src/utils/jwt.utils.ts
@@ -2,7 +2,7 @@ import jwt from "jsonwebtoken";
 import type { UserRole } from "@prisma/client";
 
 const JWT_SECRET: string = process.env["JWT_SECRET"] ?? (() => { throw new Error("JWT_SECRET environment variable is required"); })();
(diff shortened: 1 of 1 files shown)


Sachinchaurasiya360 (approved):
Approving. Failing fast on a missing secret is the right pattern and matches how JWT_SECRET is handled two lines above.

Operational context (handled before merging): `VERIFICATION_SECRET` was not set in the Vercel production env, so merging this as-is would have crashed the server at boot on the next deploy (module-level throw in jwt.utils.ts, which is imported app-wide). I've set `VERIFICATION_SECRET` in the Vercel Production and Preview environments to the same value the fallback was using, verified empirically against the live verify endpoint, so every already-issued skill-verification certificate link (365-day tokens) remains valid. Rotating it to a distinct secret can now be done deliberately later, at the cost of invalidating outstanding certificate links at that time.

PR #2683, opened 2026-07-08
perf: add database indexes on dsaProblemReport foreign keys to prevent full table scans

Akshita-2307:
Fixes #2660

Adds `@@index([userId])` and `@@index([problemId])` on `dsaProblemReport` to prevent full table scans when querying reports by user or problem. The `contactSubmission` model already has `@@index([createdAt])`.


Sachinchaurasiya360 (approved):
Approving the schema change: both `dsaProblemReport.userId` and `.problemId` are FK columns queried by the report endpoints, and Postgres does not auto-index FK columns, so these are legitimate.

One gap, handled at merge time: the PR adds the `@@index` lines to `base.prisma` but includes no migration file, and prod migrations run via `prisma migrate deploy` (which only applies files under `prisma/migrations/`). A schema-only change would never reach the prod database. I'm generating the corresponding migration on main right after merging and applying it to prod.

PR #2684, opened 2026-07-08
fix: add 30s request timeout to resend.batch.send() to prevent hanging workers

Akshita-2307:
Fixes #2661

Wraps the `resend.batch.send(payload)` call with an AbortController-based 30-second timeout using a Promise.race pattern. If the request hangs longer than 30s, it throws a timeout error that triggers the existing retry logic. This prevents cron/serverless workers from blocking indefinitely on an unresponsive Resend API.

### DIFF
--- a/server/src/utils/email.utils.ts
+++ b/server/src/utils/email.utils.ts
@@ -148,7 +148,12 @@ export async function sendEmailBatch(
 
   for (let attempt = 0; attempt <= maxRetries; attempt++) {
     try {
-      const result = await resend.batch.send(payload);
+      const result = await Promise.race([
+        resend.batch.send(payload),
+        new Promise<never>((_, reject) =>
+          setTimeout(() => reject(new Error("Resend batch.send timed out after 30s")), 30_000)
+        ),
+      ]);
       if (result.error) {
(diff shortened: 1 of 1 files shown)


Sachinchaurasiya360 (approved):
Approving, this achieves the goal: a hung `resend.batch.send()` can no longer block a worker past 30s, which matters on Vercel where the cron functions have a hard execution cap.

Three notes for the record (none blocking):

1. The PR description says AbortController, but the implementation is a plain `Promise.race` with a rejecting timer. That's fine behaviorally, though the underlying HTTP request is not actually aborted, it just stops being awaited.
2. The description claims the timeout error "triggers the existing retry logic", but it doesn't: the catch block only retries when the message matches `/429|rate.?limit/i`, and "Resend batch.send timed out after 30s" doesn't match. So a timeout fails the batch immediately. That's arguably the right behavior (fail fast, report in `errors`), just not what the description says.
3. The 30s timer is never cleared, so after a successful send the timer stays pending for up to 30s. Harmless on Vercel (the runtime doesn't wait for the event loop to drain after the response), but a `finally { clearTimeout(...) }` would be tidier. Fine as a follow-up.

PR #2685, opened 2026-07-08
fix: sync OssContributionHeatmap selectedYear with availableYears after data loads

Akshita-2307:
Fixes #2662

Adds a `useEffect` to update `selectedYear` when `availableYears` changes (i.e. after the async query completes). Previously `selectedYear` was initialized once on mount with the fallback `availableYears[0]` (current year), and never updated when real data arrived — causing the heatmap to show empty for years outside the data range.

### DIFF
--- a/client/src/components/OssContributionHeatmap.tsx
+++ b/client/src/components/OssContributionHeatmap.tsx
@@ -1,4 +1,4 @@
-import React, { useMemo, useState } from "react";
+import React, { useEffect, useMemo, useState } from "react";
 import { useQuery } from "@tanstack/react-query";
 import { ActivityCalendar } from "react-activity-calendar";
 import api from "../lib/axios";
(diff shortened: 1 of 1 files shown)


Sachinchaurasiya360 (changes requested):
Thanks for taking this on, but this doesn't fix the scenario described in #2662.

**Why the new effect never fires on initial load:** `availableYears` unconditionally adds the current year (line 55, `years.add(new Date().getFullYear())`), and `selectedYear` initializes to the current year. So when data arrives, `availableYears.includes(prev)` is always true and the effect keeps the current year selected. The reported symptom (a user whose contributions are all in past years sees an empty current-year grid and must click a year button manually) is unchanged by this PR.

**What the effect does help with:** switching `studentId` after manually selecting a year that the new student doesn't have. That's a real but much narrower case than the issue describes.

**Suggested fix:** key the default off the years that actually contain data, not the button list:

```tsx
const dataYears = useMemo(() => {
  if (!data || data.length === 0) return [];
  const years = new Set<number>();
  data.forEach(d => years.add(new Date(d.date).getFullYear()));
  return Array.from(years).sort((a, b) => b - a);
}, [data]);

useEffect(() => {
  setSelectedYear(prev => dataYears.includes(prev) ? prev : (dataYears[0] ?? new Date().getFullYear()));
}, [dataYears]);
```

This keeps the current selection when it has data, and otherwise jumps to the most recent year that does. `availableYears` can stay as-is for rendering the year buttons (current year remains selectable, just not the default when empty).

PR #2687, opened 2026-07-09
fix: wire Zod validation schemas into upload routes

Xenon010101:
## Description
Four Zod validation schemas defined in `upload.validation.ts` were never wired into the routes. This PR imports them and adds `validateBody(schema)` middleware to the 4 upload routes, consistent with every other module in the codebase.

**Schemas wired:**
- `uploadProfilePicSchema` → `POST /profile-pic`
- `uploadCoverImageSchema` → `POST /cover-image`
- `uploadResumeSchema` → `POST /profile-resume`
- `deleteResumeSchema` → `DELETE /profile-resume`

Fixes #2678

## Summary by CodeRabbit

* **Bug Fixes**
  * Added request validation for profile upload actions, including profile picture, cover image, profile resume, and resume deletion.
  * Users will now receive clearer feedback when required upload details are missing or invalid before the request is processed.

### DIFF
--- a/server/src/module/upload/upload.routes.ts
+++ b/server/src/module/upload/upload.routes.ts
@@ -6,6 +6,10 @@ import {
   validateBody,
   presignRequestSchema,
   guestPresignRequestSchema,
+  uploadProfilePicSchema,
+  uploadCoverImageSchema,
+  uploadResumeSchema,
+  deleteResumeSchema,
 } from "./upload.validation.js";
 
 const uploadController = new UploadController();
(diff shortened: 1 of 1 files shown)


Sachinchaurasiya360 (changes requested):
Good catch that these schemas were never wired up, and for `/profile-pic`, `/cover-image`, and the DELETE this works as-is (client sends `fileUrl` / `url`, controllers read the same).

**Blocking issue on `POST /profile-resume`:** `validateBody` replaces the body with the parsed result (`req.body = result.data`), and Zod object schemas strip unknown keys. The client (`client/src/utils/upload.ts`) posts `{ fileUrl, originalName, size, mimeType }`, but `uploadResumeSchema` only declares `fileUrl`, so the other three fields are silently stripped. Two consequences:

1. The controller's server-side size cap (`typeof size === "number" && size > MAX_FILE_SIZE`) goes dead, since `size` is now always `undefined`. This PR would silently remove an existing validation while adding validation.
2. The response's `file: { url, originalName, size, mimeType }` metadata comes back with three undefined fields.

Fix is small, extend the schema:

```ts
export const uploadResumeSchema = z.object({
  fileUrl: z.string().url("Valid file URL required").refine(validateS3Url, "File must be from authorized S3 bucket"),
  originalName: z.string().optional(),
  size: z.number().optional(),
  mimeType: z.string().optional(),
});
```

Happy to approve once that's in.

PR #2688, opened 2026-07-09
fix: remove dead guestResult computation in getPublicProfile

Xenon010101:
## Description
Removes dead `guestResult` computation in `getPublicProfile()`. The `guestResult` variable was computed on every profile request (lines 496-497) but never used — non-owner/non-authorized visitors are rejected at line 482, authorized viewers return early at line 503, and the ternary at line 509 always takes the `fullResult` branch since only the owner reaches it.

**Changes:**
- Removed the `guestResult` destructure/computation (5 lines)
- Simplified return to `{ profile: fullResult, cacheHit: false }`

Fixes #2677

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated public profile responses so authenticated viewers receive the full profile data after access checks.
  * Kept admin caching behavior unchanged.

### DIFF
--- a/server/src/module/auth/auth.service.ts
+++ b/server/src/module/auth/auth.service.ts
@@ -492,10 +492,6 @@ export class AuthService {
     const fullResult = {
       ...rest,
     };
-    // Guest-safe view: strip PII regardless of who triggered the request.
-    const { email: _email, contactNo: _contactNo, ...guestRest } = fullResult as any;
-    const guestResult = guestRest;
-
     // Authorized tier (admin): cache and return the full view under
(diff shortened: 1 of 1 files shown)


Sachinchaurasiya360 (approved):
Approving. Verified the reachability claim against the current code: non-owner non-admin visitors throw `Profile is private` at line 482, admins return at line 503, so the final return is only ever reached by the owner and the `isOwner ? fullResult : guestResult` ternary always took the `fullResult` branch. `guestResult` was computed on every uncached request and never used or cached, so this is dead code with no behavior change and no PII exposure risk.

PR #2689, opened 2026-07-09
fix: make job-agent email daily limit check atomic to prevent race condition

Xenon010101:
## Description
Replaces the non-atomic daily limit check + email send + log insert pattern with an atomic `INSERT ... SELECT ... WHERE (SELECT COUNT(*)...) < limit RETURNING id` query. The log row is inserted first — if the daily cap is reached, the INSERT selects 0 rows and the request is rejected before any email is sent. If the email send fails, the log row is cleaned up using the returned ID.

**Changes:**
- Removed separate `sentToday` count query (was race-prone)
- Removed non-atomic `prisma.jobAgentEmailLog.create` (was after email send)
- Added atomic conditional `INSERT ... RETURNING id` — limit check and log insert in one statement
- Email send happens only after the insert confirms the limit is not exceeded
- On email failure, the log row is deleted using the returned ID

Fixes #2676

### DIFF
--- a/server/src/module/job-agent/job-agent.service.ts
+++ b/server/src/module/job-agent/job-agent.service.ts
@@ -441,7 +441,7 @@ export class JobAgentService {
     const now = new Date();
     const context = truncateContext(input.context);
 
-    const [user, lastEmail, sentToday] = await Promise.all([
+    const [user, lastEmail] = await Promise.all([
       prisma.user.findUnique({
         where: { id: userId },
         select: { name: true, email: true },
@@ -450,9 +450,6 @@ export class JobAgentService {
         where: { userId },
         orderBy: { createdAt: "desc" },
         select: { createdAt: true },
-      }),
-      prisma.jobAgentEmailLog.count({
-        where: { userId, createdAt: { gte: startOfToday() } },
       }),
     ]);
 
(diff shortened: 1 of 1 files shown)


Sachinchaurasiya360 (changes requested):
The atomicity approach is right (conditional `INSERT ... SELECT ... RETURNING` is exactly the pattern for this), but the raw SQL targets a table that doesn't exist, so every send request would fail at runtime.

**Blocking: wrong table/column names.** The `jobAgentEmailLog` model in `base.prisma` has no `@@map`/`@map` attributes, so the actual Postgres table is `"jobAgentEmailLog"` with columns `"userId"`, `"jobIds"`, `"context"`, `"sentCount"`, `"createdAt"`. The query uses `"job_agent_email_log"` / `"user_id"` / etc., which will throw `relation "job_agent_email_log" does not exist` on the first request. This needs a run against a real database before re-requesting review.

Two smaller things while you're in there:

1. The `string_to_array(csv)::int[]` workaround isn't needed: Prisma's `$queryRaw` serializes a JS number array to a Postgres array parameter, so you can pass `${orderedJobs.map(j => j.id)}` directly.
2. In the failure cleanup paths, wrap the `prisma.jobAgentEmailLog.delete` in its own try/catch (log on failure). If the delete itself throws, it currently replaces the 503 `JobAgentEmailError` with an unhandled error and permanently burns one slot of the user's daily limit.

PR #2715, opened 2026-07-10
fix: return 409 for duplicate email registration (#2644)

Akshita-2307:
### Description

Fixes email enumeration vulnerability in the register endpoint.

### Changes
- Changed status code from 201 to 409 for duplicate email registration
- Returns error message instead of pretending success

Before: returned 201 with success message (inconsistent - no user field) -> leaks email existence
After: returns 409 Conflict with proper error message

Closes #2644

GSSoC

## Summary by CodeRabbit

- **Bug Fixes**
  - Registration attempts using an already registered email now return a **409 Conflict** response instead of appearing successful.

### DIFF
--- a/server/src/module/auth/auth.controller.ts
+++ b/server/src/module/auth/auth.controller.ts
@@ -38,7 +38,7 @@ export class AuthController {
       return res.status(201).json({ message: "Registration successful. Please verify your email to continue.", user: data.user });
     } catch (error) {
       if (error instanceof Error && error.message === "Email already registered") {
-        return res.status(201).json({ message: "Registration successful. Please verify your email to continue." });
+        return res.status(409).json({ message: "Email already registered" });
       }
       console.error(error);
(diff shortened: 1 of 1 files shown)


Sachinchaurasiya360:
This duplicates #2744, which makes the same change (duplicate-email registration `201` -> `409`). We'll pick one to avoid a conflict.

Same note as there: the current `201`/generic-message response looks like an intentional **account-enumeration** defense, so switching to `409` "Email already registered" is a security/UX tradeoff that needs a decision. Holding both pending that call.

Sachinchaurasiya360 (commented):
Thanks for this! It duplicates #2744, which changes the same line (fake 201 to 409 on duplicate email) and was just merged, so I'm closing this as a duplicate.

One note on the framing: returning 409 on a duplicate actually reveals that the email exists, whereas the previous fake-201 hid it. So this is a UX-vs-enumeration tradeoff, and we went with the clearer UX. Appreciate the contribution!

PR #2742, opened 2026-07-16
fix: make job-agent email daily limit check atomic to prevent race condition

Xenon010101:
# Pull Request

## Description
Replaces the non-atomic daily limit check + email send + log insert pattern with an atomic INSERT ... SELECT ... WHERE (SELECT COUNT(*)...) < limit RETURNING id query. The log row is inserted first — if the daily cap is reached, the INSERT selects 0 rows and the request is rejected before any email is sent. If the email send fails, the log row is cleaned up using the returned ID.

## Related Issue
Fixes #2676

## Type of Change
- [x] Bug Fix

## Changes
- Removed separate sentToday count query (was race-prone)
- Added atomic conditional INSERT with limit check in one statement
- Email send happens only after the insert confirms limit not exceeded
- On email failure, the log row is deleted using the returned ID
- Wrapped cleanup delete in try/catch to prevent error masking

## How to test
1. Send multiple concurrent email requests — only JOB_EMAIL_DAILY_LIMIT (5) should succeed
2. Verify that failed email sends clean up the log row (daily limit slot is not burned)
3. Verify cooldown still works (60s between sends)

## Checklist
- [x] Tested manually
- [x] No .env, credentials, or node_modules committed

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved enforcement of the daily job-email sending limit.
  * Failed email attempts no longer consume a daily sending allowance.
  * Prevented duplicate or over-limit sends under concurrent requests.
  * Email delivery failures continue to return a clear service error.


Sachinchaurasiya360:
Thanks for this. The atomic `INSERT ... SELECT ... WHERE (count) < limit RETURNING id` approach is the right one and matches our DB guidelines (single round-trip, no separate `count()` before the write, cleanup of the log row on send failure). I verified the `jobAgentEmailLog` columns (`userId`, `jobIds Int[]`, `context`, `sentCount`, `createdAt`) match the raw insert.

One blocker: this branch is currently **conflicting with `main`** (`job-agent.service.ts` changed underneath it). Could you rebase onto latest `main` and resolve the conflict? Once it shows mergeable we can bring it in.

Sachinchaurasiya360 (commented):
Heads up: the diff for this PR is currently empty (no changes against `main`). The last commit ("Merge upstream/main — resolve conflicts by taking upstream's version") appears to have reset the branch to match upstream, and the atomic daily-limit fix for #2676 may already be on main via another PR.

There's nothing to merge here as-is. Could you confirm whether #2676 is already resolved on main, and close this if so, otherwise re-push your changes? Thanks!

PR #2746, opened 2026-07-16
fix(tracker): prevent silent overwrite of tracked applications by inc…

ssuyashhhh:
## Description
Fixed a critical bug where tracking multiple job applications with the same generic URL (e.g., a company's main careers portal) would silently overwrite the previous application. The duplicate check now properly requires the `role` to match in addition to the `applicationUrl`, preventing data corruption and silent data loss for multi-role applications.

## Related Issue
Fixes #2745 

## Type of Change
- [x] Bug Fix  
- [ ] Feature  
- [ ] Enhancement  
- [ ] Documentation  

## Testing
- Ran TypeScript checks (`npx tsc --noEmit`) to verify that the modified Prisma queries conform to the schema and cause no regressions.
- Validated the logic that `applicationUrl` AND `role` are correctly evaluated before performing the upsert in `application-tracker.service.ts`.

## Screenshots / Video

_No UI changes in this PR_

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Autofill now respects field-specific settings, allowing selected information to be disabled.
  - Application dates now accept timezone-offset datetime formats.

- **Bug Fixes**
  - Improved duplicate detection for tracked applications using normalized roles and URLs.
  - Legacy applications are filtered more accurately in lists.
  - Invalid application identifiers are rejected consistently.
  - Application activity events are recorded more reliably.

- **Improvements**
  - AI-generated application answers are extracted more robustly from responses that include additional text.


ssuyashhhh:
@Sachinchaurasiya360 do reopen it

ssuyashhhh:
@Sachinchaurasiya360 do review

Sachinchaurasiya360:
Reviewed in depth and the server-side changes look correct:

- `parseId` correctly strips the `tracked-` prefix and rejects `legacy-` IDs, which matches what the list endpoint emits (`tracked-${id}` / `legacy-${id}`), and it stays backward-compatible with plain numeric IDs.
- The `applicationUrl` + `role` (case-insensitive) dedup key fixes the silent-overwrite bug.
- The `FOR UPDATE` transaction in `addEvent` is race-safe; table/column names verified against the schema (no `@@map`).
- `datetime({ offset: true })` is a valid fix for ISO strings with a timezone offset.

One thing to sequence before merge: this also touches the browser extension (`extension/src/content/*`), which ships as a prebuilt zip. The `autofillSettings` behavior won't take effect for users until the extension bundle is regenerated, so we should pair the zip rebuild with this merge.

ssuyashhhh:
@Sachinchaurasiya360 i had made changes as u said do review and do reconsider the difficulty label i believe this pr is atleast intermediate

Sachinchaurasiya360 (changes requested):
Thanks for the effort here. The blocker is scope: this PR bundles several unrelated changes into one:

- extension autofill settings (`fill-engine.ts`, `content/index.ts`)
- JSON brace-matching parse in `application-profile.service.ts`
- `parseId` prefix handling + the legacy-fetch conditional in `application-tracker`
- a rebuilt `internhack-extension.zip` (binary)

Two problems: (1) the description says the fix requires `role` to match alongside `applicationUrl`, but I don't see that dedup change in the diff (the tracker service change is about conditional legacy fetching, not the dedup). (2) A binary zip plus four unrelated concerns can't be reviewed safely as one unit.

Could you split this into one focused PR per concern, and in the tracker PR point to the exact role+applicationUrl dedup change? Regenerate the extension zip only in the PR that actually changes the extension source. Happy to review each quickly once split.

Sachinchaurasiya360 (changes requested):
## Requesting changes: four unrelated changes, and the dedupe fix has gaps

Please split this. The title covers only one of the four changes in the diff (extension autofill gating, Gemini JSON extraction, tracker id parsing, tracker dedupe), which makes it hard to review or revert any of them independently.

On the dedupe fix itself:

### 1. Dedupe is disabled whenever `role` is empty

```js
const role = normalizeText(data.role);
const existing = (applicationUrl && role) ? await prisma.trackedJobApplication.findFirst({ ... }) : null;
```

`normalizeText` returns `""` for null or blank, so a record with no role skips the lookup entirely and always creates a duplicate. Before this PR, `applicationUrl` alone was enough to find an existing row. Adding `role` to the match key is reasonable (the same URL can host different roles), but falling back to no dedupe at all when role is missing is a regression. Please fall back to matching on `applicationUrl` alone in that case.

### 2. `update` still overwrites silently

The PR moves `role` and `applicationUrl` out of the generic loop and normalizes them:

```js
if (data.role !== undefined) payload.role = normalizeText(data.role);
if (data.applicationUrl !== undefined) payload.applicationUrl = normalizeText(data.applicationUrl) || null;
```

but `update` never re-checks the dedupe constraint. So an edit that changes a row's `applicationUrl` and `role` to match an existing row still produces the duplicate the PR is meant to prevent. Also note the second line turns an empty-string update into `null`, which will clear a stored URL when a client sends `""`. Worth confirming that is intended.

### 3. `shouldFetchLegacy` duplicates a value it should read

```js
const shouldFetchLegacy =
  !filters.savedOnly &&
  (!filters.status || filters.status === "APPLIED") &&
  (!filters.sourceType || filters.sourceType === "INTERNHACK_ADMIN");
```

The filter twenty lines below uses the helper instead of a literal:

```js
.filter(() => !filters.status || filters.status === statusForLegacyExternal())
```

`statusForLegacyExternal()` returns `"APPLIED"` today so the two agree, but they now have to be kept in sync by hand, and if the helper ever changes the guard silently drops legacy rows. Please call `statusForLegacyExternal()` here too.

Skipping the query when no legacy row can match is a good optimization otherwise.

### 4. The Gemini JSON extraction is quadratic

```js
for (let i = startIndex + 1; i <= text.length; i++) {
  if (text[i - 1] === '}') {
    try {
      const candidate = text.slice(startIndex, i);
      JSON.parse(candidate);
```

This attempts a full `JSON.parse` on a growing substring at every `}` in the response. For a long reply with many closing braces that is a lot of repeated parsing work. Tracking brace depth in a single pass to find the matching close, then parsing once, is both faster and easier to read. The existing fence-stripping (`replace(/```json|```/g, "")`) plus a single parse also handled the common case; if it was failing on some specific response shape, a comment with that example would help.

### 5. `parseId` prefix contract is undocumented

```js
if (!raw || raw.startsWith("legacy-")) return null;
const clean = raw.replace(/^tracked-/, "");
```

Rejecting `legacy-` ids is correct, since those rows are not updatable. But the `tracked-` and `legacy-` prefixes are now a real part of the route contract with nothing documenting or testing them. A short comment naming where the client produces these, plus a test, would make this safe to change later.

The extension autofill gating looks fine, though `autofillSettings?: Record<string, unknown>` with an `as` cast at the call site loses the type safety you could get from a proper settings interface.

PR #2753, opened 2026-07-17
fix: remove as any casts in notes controller

Akshita-2307:
## Description

The notes.controller.ts used \s any\ casts for the \contentType\ parameter, bypassing TypeScript's type checking.

**Changes**
- Replaced \s any\ casts with proper \NoteContentType\ enum casts
- Added \NoteContentType\ import from the notes validation module
- Extracted \contentType\ and \search\ from \
eq.query\ with explicit type narrowing instead of passing \ParsedQs\ directly to \getNotes\

Closes #2735

GSSoC

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of note content types when retrieving, saving, and deleting notes.
  * Enhanced note search requests to process content type and search filters more reliably.

### DIFF
--- a/server/src/module/notes/notes.controller.ts
+++ b/server/src/module/notes/notes.controller.ts
@@ -1,4 +1,5 @@
 import type { Request, Response, NextFunction } from "express";
+import { NoteContentType } from "@prisma/client";
 import { NotesService } from "./notes.service.js";
 
 export class NotesController {
@@ -8,7 +9,8 @@ export class NotesController {
     try {
       const userId = req.user?.id;
       if (!userId) return res.status(401).json({ error: "Unauthorized" });
-      const notes = await this.notesService.getNotes(userId, req.query as any);
+      const { contentType, search } = req.query as { contentType?: NoteContentType; search?: string };
+      const notes = await this.notesService.getNotes(userId, { contentType, search });
       res.json({ notes });
     } catch (err) {
       next(err);
@@ -22,8 +24,8 @@ export class NotesController {
       const { contentType, contentId } = req.params;
       const note = await this.notesService.getNote(
         userId,
-        contentType as any,
-        contentId as string
+        contentType as NoteContentType,
+        contentId
       );
       res.json({ note });
     } catch (err) {
@@ -39,8 +41,8 @@ export class NotesController {
       const { note } = req.body;
       const saved = await this.notesService.saveNote(
         userId,
-        contentType as any,
-        contentId as string,
+        contentType as NoteContentType,
+        contentId,
         note
       );
       res.json({ note: saved });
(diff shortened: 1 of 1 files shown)


Sachinchaurasiya360:
Good direction removing the `as any` casts. The `getNote` / `saveNote` / `deleteNote` changes are fine (`contentId` is already a `string` from params, and `contentType as NoteContentType` is correct).

The one that will break the build is `getNotes(userId, req.query)`: `getNotes` expects `{ contentType?: NoteContentType; search?: string }`, but `req.query` is `ParsedQs` (its index signature is `string | string[] | ...`), which is not assignable to that shape. That's exactly why the `as any` was there. Could you either parse/cast `req.query` into the expected filter shape (e.g. read `contentType`/`search` explicitly), or widen the service parameter? Everything else looks good.

Sachinchaurasiya360 (changes requested):
Thanks for cleaning up the `as any` casts. The `contentType as NoteContentType` replacements are good.

One blocker: changing `getNotes(userId, req.query as any)` to `getNotes(userId, req.query)` will fail the server typecheck. `getNotes` expects `filters?: { contentType?: NoteContentType; search?: string }`, but `req.query` is Express's `ParsedQs`, where `contentType` is `string | string[] | ParsedQs` and is not assignable to `NoteContentType`.

Please either narrow it with an explicit cast (e.g. `req.query as { contentType?: NoteContentType; search?: string }`) or validate/parse the query into that shape. Could you run `npm run typecheck` in `server/` and confirm it passes? Once that line compiles, happy to merge.

Sachinchaurasiya360 (changes requested):
## Requesting changes: this breaks the server typecheck

Replacing `as any` with the real `NoteContentType` is a good direction, but removing `as string` from `contentId` does not typecheck. `Typecheck Server` fails with three errors, and this is not the known main-branch breakage:

```
src/module/notes/notes.controller.ts(28,9): error TS2345: Argument of type 'string | string[]' is not assignable to parameter of type 'string'.
src/module/notes/notes.controller.ts(45,9): error TS2345: ...
src/module/notes/notes.controller.ts(62,9): error TS2345: ...
```

`req.params` is typed `string | string[]` in this project's Express setup, so the cast was load-bearing.

### A second point worth addressing

```js
const { contentType, search } = req.query as { contentType?: NoteContentType; search?: string };
```

and `contentType as NoteContentType` are still assertions on unvalidated input. They are narrower lies than `as any`, but they are still lies: a request with `?contentType=bogus` satisfies the compiler and then reaches Prisma, which throws a validation error and surfaces as a 500 rather than a 400.

Since the goal here is type safety rather than a behaviour change, the fix that actually earns it is validating at the boundary. This module has no validation file yet, but the pattern is established elsewhere in the codebase, for example `opensource.validation.ts`:

```ts
import { z } from "zod";
import { NoteContentType } from "@prisma/client";

export const noteParamsSchema = z.object({
  contentType: z.nativeEnum(NoteContentType),
  contentId: z.string().min(1),
});
```

Then `safeParse` in the controller and return 400 on failure. That removes the assertions instead of renaming them, and gives a correct status code for bad input.

Minimum to unblock: restore the `contentId` handling so it compiles. The validation cleanup is the more valuable half if you are willing to take it on.

PR #2761, opened 2026-07-17
fix(admin): invalidate JWT sessions immediately when banning users

ManoharKonala:
Fixes #2764 

This PR fixes a critical security bug where banned users could continue using the platform because their existing JWTs were not invalidated upon deactivation.

### What was happening? 
The updateUserStatus method in admin-platform.service.ts updated the isActive flag in the database but failed to increment the user's tokenVersion or invalidate the version cache. Because the auth middleware only checks tokenVersion and not the isActive flag on every request, banned users remained authenticated until their token expired naturally.

### What changed?

Updated updateUserStatus to increment tokenVersion: { increment: 1 } when changing a user's status.
Added a call to invalidateVersionCache(userId) immediately after the database update to ensure existing sessions are instantly revoked.

### DIFF
--- a/server/src/module/admin/admin-platform.service.ts
+++ b/server/src/module/admin/admin-platform.service.ts
@@ -192,11 +192,13 @@ export class AdminPlatformService {
     if (userId === adminId) throw new Error("Cannot modify your own status");
 
     try {
-      return await prisma.user.update({
+      const user = await prisma.user.update({
         where: { id: userId },
-        data: { isActive },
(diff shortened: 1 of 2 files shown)


Sachinchaurasiya360:
The diff here only adds two blank lines, it doesn't add the `deleteExternalApplication` controller method the title describes, so there's nothing functional to merge as-is. Looks like the intended change wasn't committed/pushed.

Could you push the actual controller method (and wire it to its route)? Happy to review once the implementation is in.

PR #2767, opened 2026-07-19
fix: reset subscriptionPlan to FREE on cancellation

Akshita-2307:
## Description

cancelSubscription does not reset subscriptionPlan to FREE -- stale plan data persists indefinitely.

### Problem
After cancellation, the user retains subscriptionPlan: MONTHLY or YEARLY indefinitely. Compare with expireSubscription which correctly resets both status and plan.

### Changes
- Added subscriptionPlan: FREE to the cancelSubscription update payload, matching the pattern used in expireSubscription

Closes #2758

GSSoC

## Summary by CodeRabbit

* **Bug Fixes**
  * Cancelled subscriptions now correctly switch to the Free plan.
  * Subscription status and plan remain consistent after cancellation.

### DIFF
--- a/server/src/module/payment/payment.service.ts
+++ b/server/src/module/payment/payment.service.ts
@@ -332,7 +332,7 @@ export class PaymentService {
 
     await prisma.user.update({
       where: { id: payment.userId },
-      data: { subscriptionStatus: "CANCELLED" },
+      data: { subscriptionStatus: "CANCELLED", subscriptionPlan: "FREE" },
     });
     await invalidateUserTierCache(payment.userId);
   }


Sachinchaurasiya360 (commented):
Thanks for the fix. One thing to confirm before merging, since this touches billing semantics:

Paid access is gated by `subscriptionStatus === "ACTIVE"` (see `config/usage-limits.ts`), so once `cancelSubscription` sets status to `CANCELLED`, access is already revoked regardless of `subscriptionPlan`. The codebase deliberately keeps `subscriptionPlan` on cancel and only resets it to `FREE` in `expireSubscription` (and the daily expiry cron), so the plan label is retained through the CANCELLED window (useful for showing "your MONTHLY plan is cancelled, active until X" and for reactivation).

So this change is harmless to access but overrides that intentional cancel-vs-expire separation. Is the immediate reset to FREE the intended product behavior, or should the plan stay until expiry? Happy to merge once you confirm the intent.

PR #2773, opened 2026-07-20
fix: batch fixes for session toast, cert validation, streak validation

saurabhhhcodes:
fix: batch fixes for session toast, cert validation, streak validation

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented repeated authentication-expiration alerts and redirects during the same session.
  * Ensured authentication expiration state resets after returning to the login page.
  * Streaks now update only after the required daily activity is verified.
  * Certificates can only be issued after the associated guide’s completion requirements are met.

### DIFF
--- a/client/src/App.tsx
+++ b/client/src/App.tsx
@@ -1,6 +1,7 @@
 import { lazy, Suspense, useEffect, type ComponentType, type LazyExoticComponent } from "react";
 import { Navigate, Route, Routes, useParams, useNavigate } from "react-router";
 import { useAuthStore } from "./lib/auth.store";
+import { resetAuthExpiredFlag } from "./lib/axios";
 import type { ProgramType } from "./module/student/opensource/OrgBrowserPage";
 import toast, { Toaster } from "./components/ui/toast";
 import { ProtectedRoute } from "./components/ProtectedRoute";
@@ -332,6 +333,7 @@ function AuthExpiredRedirect() {
     const handler = () => {
       toast.error("Session expired. Please log in again.");
       navigate("/login", { replace: true });
+      resetAuthExpiredFlag();
     };
     window.addEventListener("auth:expired", handler);
     return () => window.removeEventListener("auth:expired", handler);
--- a/client/src/lib/axios.ts
+++ b/client/src/lib/axios.ts
@@ -4,10 +4,16 @@ import { useAuthStore } from "./auth.store";
 export const API_BASE = (import.meta.env.VITE_API_URL as string | undefined) ?? "http://localhost:3000/api";
 export const SERVER_URL = API_BASE.replace(/\/api\/?$/, "");
 
+let authExpiredFired = false;
+
+export function resetAuthExpiredFlag() {
+  authExpiredFired = false;
+}
+
 const api = axios.create({
   baseURL: API_BASE,
-  withCredentials: true, // sends httpOnly cookie automatically
-  timeout: 30_000,       // 30-second timeout to avoid hanging requests
+  withCredentials: true,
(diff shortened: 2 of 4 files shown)


Sachinchaurasiya360 (commented):
This currently has merge conflicts against `main`, so it can't be merged as-is. Could you rebase on the latest main and resolve them?

Two notes while you're in there:
1. This bundles three separate fixes (session-toast dedup, cert validation, streak validation) across client and server. Splitting into focused PRs would make each much faster to review and merge.
2. In `axios.ts` the change removed the explanatory comments (`// sends httpOnly cookie automatically`, `// Dispatch a custom event so the router can handle navigation...`). Please keep those, they explain non-obvious behavior.

The auth-expired dedup flag (`authExpiredFired` + `resetAuthExpiredFlag`) approach itself looks good.

Sachinchaurasiya360 (changes requested):
## Requesting changes: one of the three files no longer exists on main, and the PR has conflicts

This bundles three unrelated fixes. Please split them, because two are salvageable and one has to be dropped.

### 1. The streak change targets deleted code (blocking)

`server/src/module/opensource/opensource-streak.service.ts` is not on main. It was removed, along with its table:

```
server/src/database/prisma/migrations/20260703000000_drop_opensource_streak/migration.sql
```

So the whole `actionVerified` block, and its queries against `opensourceStreak`, cannot be merged. This is why the PR shows as conflicting. Please drop that file from the PR entirely.

For what it is worth, the verification idea was sound: trusting a client-supplied `action` without checking the activity happened is a real gap. If streaks come back, two notes for next time. The thrown `Error` would surface as a 500 through the global handler, when "you have not done this activity today" is a 400. And any `action` value outside the three branches leaves `actionVerified` false and always throws, so a fourth action type would break silently.

### 2. The `auth:expired` dedupe does not dedupe

```js
if (!authExpiredFired) {
  authExpiredFired = true;
  window.dispatchEvent(new CustomEvent("auth:expired"));
}
```

with the handler in `App.tsx`:

```js
navigate("/login", { replace: true });
resetAuthExpiredFlag();
```

`window.dispatchEvent` is synchronous, so the listener runs to completion inside the `dispatchEvent` call. The flag is set, the event fires, the handler resets the flag, and `dispatchEvent` returns with `authExpiredFired` back to `false`. The next 401 in the same burst fires another toast, which is the case this is meant to fix.

Either do not reset it at all (the flag is per page load, and a successful login reloads state anyway), or reset on successful login rather than in the expiry handler.

Also please restore the comments that were removed as drive-by edits:

```js
-  withCredentials: true, // sends httpOnly cookie automatically
-  timeout: 30_000,       // 30-second timeout to avoid hanging requests
-      // Dispatch a custom event so the router can handle navigation
-      // without a full-page reload that wipes in-memory state.
```

The last one in particular explains a non-obvious design decision.

### 3. Certificate gating depends on a slug guess

```js
const guideSlug = guideName.toLowerCase().replace(/[^a-z0-9]+/g, "-").replace(/^-|-$/g, "");
const progress = await prisma.guideProgress.findUnique({
  where: { userId_guideSlug: { userId, guideSlug } },
```

Gating certificates on real completion is worth doing, but deriving `guideSlug` from `guideName` assumes the stored slugs match this exact transform. If any row was written with a different slug, `findUnique` misses and every certificate request for that guide fails. Please look up the slug from wherever `guideProgress` rows are created, or pass it in, instead of reconstructing it.

The `completedStepIds.length >= 7` fallback also needs a comment or a named constant. As written it is a magic number applied regardless of which guide is being requested.

PR #2785, opened 2026-07-21
fix: persist retry state in sessionStorage to prevent infinite reload loop (#2774)

Akshita-2307:
## Description

\
etryState\ was a module-level \Map\, which gets recreated empty on every page reload. This means the check \if (!retryState.get(factory))\ is always truthy, causing \window.location.reload()\ to be called infinitely every time a chunk fails to load.

**Fix:** Replace the in-memory Map with \sessionStorage\, which persists across reloads. The retry key is derived from the factory function's source string (which is stable across reloads since it matches the import path).

Closes #2774

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recovery when loading application routes or components fails.
  * Retry attempts now persist across page reloads during the same browser session, helping prevent repeated reload loops while still allowing a one-time recovery attempt.

### DIFF
--- a/client/src/App.tsx
+++ b/client/src/App.tsx
@@ -11,14 +11,15 @@ import ScrollProgressBar from "./components/common/ScrollProgressBar";
 import ScrollToTop from "./components/common/ScrollToTop";
 const ContributorsPage = lazyWithRetry(() => import("./module/contributors/ContributorsPage"));
 
-const retryState = new Map<() => Promise<{ default: ComponentType<unknown> }>, boolean>();
+const RETRY_KEY_PREFIX = "lazy-retry:";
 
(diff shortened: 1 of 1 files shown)


Sachinchaurasiya360 (approved):
﻿This fixes the actual bug rather than the symptom. The old retry flag lived in a module-level `Map`, and the retry path calls `window.location.reload()`, which wipes that module state. So the flag was never observed as set, and a persistently failing chunk reloaded forever. `sessionStorage` survives the reload, which is exactly the property needed.

Using `factory.toString()` as the key is fine here since each factory closes over a distinct import path, so the strings stay distinct after minification.

One tradeoff worth noting for later: the flag is never cleared, so if a chunk fails once and then succeeds, a genuinely new failure later in the same tab (a deploy mid-session) will not get its one retry. That is the right side to err on versus an infinite reload loop.

PR #2786, opened 2026-07-21
fix: wrap resume upload in Prisma transaction to prevent TOCTOU race (#2775)

Akshita-2307:
## Description

\uploadProfileResume\ performed a read-then-write (\indUnique\ → compute → \update\) outside a transaction. Two concurrent uploads could both read the same old resume list, and the second \update\ would silently overwrite the first, losing a resume.

**Fix:** Wrap the read and write inside \prisma.\\ so the entire read-modify-write sequence is atomic. If two requests race, one will retry automatically and see the correct state.

Closes #2775

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resume uploads to prevent conflicting updates when multiple uploads occur close together.
  * Ensured resume records remain consistent if an upload update fails.
  * Continued enforcing the maximum resume limit by removing the oldest stored resume when necessary.
  * Resume files are now removed only after the updated resume information is successfully saved.

### DIFF
--- a/server/src/module/upload/upload.controller.ts
+++ b/server/src/module/upload/upload.controller.ts
@@ -251,21 +251,27 @@ export class UploadController {
       }
 
       const userId = req.user.id;
-      const current = await prisma.user.findUnique({ where: { id: userId }, select: { resumes: true } });
 
-      let updatedResumes = current?.resumes ?? [];
-      if (updatedResumes.length >= MAX_RESUMES) {
-        const oldest = updatedResumes[0]!;
-        deleteFile(oldest);
-        updatedResumes = [...updatedResumes.slice(1), fileUrl];
-      } else {
-        updatedResumes = [...updatedResumes, fileUrl];
-      }
+      const user = await prisma.$transaction(async (tx) => {
+        const current = await tx.user.findUnique({
+          where: { id: userId },
+          select: { resumes: true },
+        });
 
-      const user = await prisma.user.update({
-        where: { id: userId },
-        data: { resumes: updatedResumes },
-        select: { id: true, name: true, email: true, role: true, contactNo: true, profilePic: true, resumes: true, company: true, designation: true, createdAt: true },
+        let updatedResumes = current?.resumes ?? [];
+        if (updatedResumes.length >= MAX_RESUMES) {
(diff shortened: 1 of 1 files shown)


Sachinchaurasiya360 (changes requested):
## Requesting changes: this does not close the TOCTOU race, and it moves S3 deletion inside the transaction

### 1. The race is still open

An interactive Prisma transaction on Postgres runs at the default isolation level, Read Committed. Wrapping a read and a write in a transaction at Read Committed does not serialize them. Two concurrent uploads can both read the same `resumes` array, both compute their own `updatedResumes`, and the second `update` overwrites the first. That is the same lost update as before the change, so the stated goal is not achieved.

Options that do close it:

**a. Serializable plus retry**, matching the pattern already used in `usage-limit.middleware.ts`:

```js
await prisma.$transaction(async (tx) => { ... }, {
  isolationLevel: Prisma.TransactionIsolationLevel.Serializable,
});
```

and handle `P2034` by retrying.

**b. Row lock**, cheaper here:

```js
await tx.$queryRaw`SELECT id FROM "user" WHERE id = ${userId} FOR UPDATE`;
```

as the first statement in the transaction, so concurrent uploads for the same user queue instead of racing.

Option b is probably the better fit, since contention is per-user and this is not a hot path.

### 2. `deleteFile` should not run inside the transaction

```js
const user = await prisma.$transaction(async (tx) => {
  ...
  if (updatedResumes.length >= MAX_RESUMES) {
    const oldest = updatedResumes[0]!;
    deleteFile(oldest);          // external S3 call, not transactional
    ...
```

S3 deletion is not covered by the database transaction and cannot be rolled back. If the transaction fails or is retried after this line (which becomes much more likely once you add Serializable plus retry from point 1), the object is already gone while the row still references it. That is a broken resume link for the user, and a retry would try to delete it a second time.

Collect the key inside the transaction, return it, and delete after commit:

```js
const { user, evictedKey } = await prisma.$transaction(async (tx) => {
  // ... compute updatedResumes, capture evictedKey, do the update
  return { user: updated, evictedKey };
});

if (evictedKey) deleteFile(evictedKey);
```

That leaves the transaction covering only database work, which is also what the repo's DB guidance asks for.

PR #2787, opened 2026-07-21
fix: add LRU eviction to cacheSet instead of silently dropping entries (#2776)

Akshita-2307:
## Description

When the in-memory cache reached \MAX_ENTRIES\ (10,000) and neither \sweepExpired()\ nor the new-key check freed space, \cacheSet\ silently returned without storing the entry. This meant valid cache writes were lost with no indication.

**Fix:**
- \cacheGet\ now re-inserts accessed entries to maintain LRU order
- \cacheSet\ evicts the least-recently-used entry (oldest by insertion/access order) instead of dropping the new entry when at capacity
- Eviction runs in a loop to ensure space for the new entry

Closes #2776

### DIFF
--- a/server/src/utils/cache.ts
+++ b/server/src/utils/cache.ts
@@ -23,13 +23,23 @@ export async function cacheGet<T>(key: string): Promise<T | null> {
     store.delete(key);
     return null;
   }
+  // Move to end to mark as recently used
+  store.delete(key);
+  store.set(key, entry);
   return JSON.parse(entry.val) as T;
 }
 
+function evictLRU(): void {
+  const oldest = store.keys().next().value;
+  if (oldest !== undefined) store.delete(oldest);
+}
+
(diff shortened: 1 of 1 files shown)


Sachinchaurasiya360 (approved):
﻿Real bug fix. The old code silently dropped the write once the map was full:

```js
if (store.size >= MAX_ENTRIES) return;
```

so past 10k entries every `cacheSet` became a no-op and the cache froze with whatever happened to be in it. Now it evicts.

The LRU mechanics are right: re-inserting on read in `cacheGet` moves the key to the end of the `Map` insertion order, and `evictLRU` takes `store.keys().next().value`, which is the least recently used. The `while` loop terminates because `MAX_ENTRIES` is 10000, so `evictLRU` always finds a key to delete. Sweeping expired entries before evicting live ones is the right order.

PR #2789, opened 2026-07-21
fix: always reset saving state in ImageCropModal handleConfirm (#2778)

Akshita-2307:
## Description

\handleConfirm\ sets \setSaving(true)\ before the async crop operation, but only resets it on error (in the \catch\ block). On success, \onCrop(blob)\ is called but \setSaving(false)\ is never invoked, leaving the button disabled permanently.

**Fix:** Move \setSaving(false)\ into a \inally\ block so it runs on both success and error paths.

Closes #2778

GSSoC


Sachinchaurasiya360 (approved):
﻿Correct. `setSaving(false)` was only in the `catch`, so on the success path the button stayed in its saving state forever. Moving it to `finally` covers both. If the parent unmounts the modal in response to `onCrop`, the trailing `setSaving` is a harmless no-op in React 18.

PR #2807, opened 2026-07-23
test:add coverage for useDebounce hook

SujalMahapatra:
# Pull Request

## Description
Describe your changes.
This PR adds dedicated unit test coverage for the reusable `useDebounce` hook.

The test suite validates the hook's core debounce behavior, ensuring delayed state updates work correctly while preventing regressions in features that rely on debounced user input.

## Changes Made

- Added `client/src/hooks/useDebounce.test.ts`
- Added tests for:
  - Initial value is returned immediately
  - Debounced value updates only after the specified delay
  - Previous timeout is cancelled when the value changes rapidly
  - Multiple consecutive value changes correctly debounce to the latest value
  - Pending timers are cleaned up on component unmount

## Related Issue
Fixes #2806 

## Type of Change
- [ ] Bug Fix  
- [ ] Feature  
- [x] Enhancement  
- [ ] Documentation  

## Testing
Explain how you tested it.

Executed:

```bash
npm test -- src/hooks/useDebounce.test.ts
```

All added tests pass successfully.

## Screenshots / Video

_No UI changes in this PR_ (delete this line if you are making UI changes)

## Checklist

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive automated coverage for debounced value updates.
  * Verified timing behavior, rapid input changes, consecutive updates, and cleanup when components are removed.

### DIFF
--- a/client/src/hooks/useDebounce.test.ts
+++ b/client/src/hooks/useDebounce.test.ts
@@ -1,0 +1,123 @@
+/**
+ * @vitest-environment jsdom
+ */
+
+import { renderHook, act } from "@testing-library/react";
+import { describe, it, expect, beforeEach, afterEach, vi } from "vitest";
+import { useDebounce } from "./useDebounce";
+
+describe("useDebounce", () => {
+    beforeEach(() => {
+        vi.useFakeTimers();
+    });
+
+    afterEach(() => {
+        vi.useRealTimers();
+    });
+
+    it("returns the initial value immediately", () => {
+        const { result } = renderHook(() =>
+            useDebounce("initial", 500)
+        );
+
+        expect(result.current).toBe("initial");
+    });
+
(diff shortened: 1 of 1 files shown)


Sachinchaurasiya360 (approved):
﻿Good tests, and they match the hook's actual semantics. The rapid-change case is the one that matters and it is asserted correctly: rerendering at t=300 clears the pending timer and starts a fresh 500ms window, so the value is still `"A"` at t=500 and only becomes `"C"` at t=800. That would catch a missing `clearTimeout` in the effect cleanup.

Both CI checks are green on this one.

Minor, not blocking: the unmount test spies on `clearTimeout`, but cleanup also runs on every effect re-run, so the assertion would pass without any unmount handling. Asserting the value never updates after `unmount` would pin the intended behaviour more tightly.

PR #2866, opened 2026-07-25
fix: handle future dates >30 days in formatDate

saurabhhhcodes:
When a date more than 30 days in the future was passed to formatDate, the negative diff fell through to the past-tense 'just now' check (since -3456000 < 60). Added a fallback return for dates >30 days in the future.

## Summary by CodeRabbit

- **Bug Fixes**
  - Future dates in relative date displays now show a localized short date when they fall beyond the supported “in N days” range.
  - Check-in errors now consistently return the appropriate responses when an employee is not found or has already checked in that day.


Sachinchaurasiya360 (changes requested):
## Requesting changes: committed `.DS_Store` files, unrelated edits, and merge conflicts

### 1. Remove the `.DS_Store` files (blocking)

```
.DS_Store              (new binary file)
server/.DS_Store       (new binary file)
```

These are macOS metadata and should not be in the repo. Please remove them from the branch and add `.DS_Store` to `.gitignore` if it is not already covered.

### 2. The attendance controller change is a no-op that reads worse

```js
if (error instanceof Error) {
-  if (error.message === "Employee not found") return res.status(404)...
+  if (error instanceof Error && error.message === "Employee not found") return res.status(404)...
```

The added `error instanceof Error` is inside a block already guarded by that exact check on the line above. It cannot be false. This also has nothing to do with the PR title. Please revert this file.

### 3. Rebase needed

The PR is currently conflicting with main.

### The actual fix is correct

The `date-utils.ts` change is right and worth landing on its own:

```js
if (ahead < 2592000) return `in ${Math.floor(ahead / 86400)} days`;
+ return d.toLocaleDateString("en-IN", { day: "numeric", month: "short", year: "numeric" });
```

Without it a future date more than 30 days out fell through the future branch entirely and got formatted by the past-date logic below, producing nonsense. Returning an absolute date is the right answer, and `en-IN` is consistent with the rest of the codebase.

Strip this PR down to just `client/src/lib/date-utils.ts`, rebase, and it is good to go.

PR #2867, opened 2026-07-26
test: add comprehensive coverage for usage-limit middleware

swarupio:
# Pull Request

## Description
Added comprehensive unit tests for `usage-limit.middleware.ts` using Vitest to ensure strict quota enforcement and concurrency safety. The test suite fully covers the required scenarios:
* Under-limit requests successfully passing to the next handler.
* At-limit requests being blocked with the correct HTTP 429 status code and payload.
* Premium tier accounts bypassing the standard usage limits.
* Concurrent load races resulting in Prisma `P2034` serialization errors being safely caught and handled.

## Related Issue
Fixes #2832

## Type of Change
- [ ] Bug Fix  
- [ ] Feature  
- [x] Enhancement  
- [ ] Documentation  

## Testing
Tested locally by executing the new test suite via Vitest. 
Run command: `npx vitest run server/src/middleware/__tests__/usage-limit.middleware.test.ts`
All 4 test cases passed successfully without errors.

## Test Cases
Screenshot 2026-07-26 163646

## Checklist

## Summary by CodeRabbit

* **Tests**
  * Added a Vitest suite covering usage-limit enforcement, including unauthenticated requests, quota-allowed vs quota-exceeded users, premium-tier bypass behavior, and concurrency conflict handling.
  * Verified returned status codes, user-facing error messages, and the structured usage payload; confirmed when request processing proceeds vs is blocked.
  * Added coverage for missing user records returning an appropriate authentication-style error.


Sachinchaurasiya360 (approved):
﻿Good coverage of the branch that matters. I checked the assertions against the middleware and they line up: the 429 body uses the FREE-tier message string exactly, `usageInfo` matches the shape assigned at line 74, and the at-limit case correctly asserts that `usageLog.create` is *not* called, which is the real invariant (`used >= limit` returns before the insert).

Driving the transaction by executing the callback against a fake `tx` is the right approach for this middleware. `Typecheck Server` and `Test Server` are both green.

Minor for a follow-up: `mockRequest.user` uses a string id (`'user-123'`) where the real `req.user.id` is a number. It compiles through the `as any`, but a numeric id would keep the fixture honest.

PR #2872, opened 2026-07-26
feat(ats): add polished PDF + CSV export with dropdown

Suhas334:
## Description

Adds a polished **Export** dropdown to the ATS Score results panel, giving users two formats to download their report for tracking improvements over time or sharing with mentors.

### PDF Report (client-side via jsPDF)
The existing basic PDF export has been replaced with a fully structured, multi-section document:
- **Header banner** (lime tint) with report title, generation date, and score tier label
- **Meta block** — resume filename and target role (if provided)
- **Overall score** rendered large with a `/100` annotation in the correct tier colour
- **Category Scores grid** (3-up layout) — uses friendly labels (`Formatting`, `Keywords`, `Experience`, `Skills`, `Education`, `Impact`) with colour-coded values; was previously raw keys only
- **Keyword Analysis** — all three groups (Present ✓, Partial ~, Missing ✗) with matching colours; was previously limited to missing keywords only
- **Improvement Suggestions** — numbered, line-wrapped, never truncated thanks to `checkBreak()` guards before every section
- **Page footer** — `InternHack ATS Report · Page N of M` on every page + generation date

### CSV Data (zero dependencies, Blob download)
New export format — useful for pasting into spreadsheets or sending to mentors:
- Columns: `Section`, `Field`, `Value`
- Rows cover: meta, all 6 category scores, all keywords (present/partial/missing labelled), numbered suggestions
- RFC-4180 compliant quoting

### Export Dropdown UX
- Replaced the plain "Download Report" button with an **Export** button + `ChevronDown` indicator
- Dropdown closes on outside-click (`mousedown` listener via `useRef`)
- `ChevronDown` rotates 180° when open
- `aria-haspopup` / `aria-expanded` on the trigger button
- Unique IDs on each menu item (`ats-export-btn`, `ats-export-pdf`, `ats-export-csv`)
- Disabled while analysis is in progress
- `print:hidden` so the button is excluded from browser print

## Related Issue
Fixes #

## Type of Change
- [ ] Bug Fix
- [x] Feature
- [x] Enhancement
- [ ] Documentation

## Testing
- Ran `npx tsc --noEmit` → **0 errors**
- Analyzed a resume → clicked **Export → PDF Report** → verified all sections present (header, score, all 6 categories with labels, all keyword groups, suggestions, footer)
- Clicked **Export → CSV Data** → verified file opens correctly in Excel/Sheets with correct columns and all rows
- Verified dropdown closes on outside-click and on option selection
- Verified button is disabled while analysis is in-flight

## Checklist
- [x] Tested manually (see steps above)

## Summary by CodeRabbit

* **New Features**
  * Added an “Export” dropdown to ATS score reports.
  * Export ATS reports as professionally formatted PDF files, including scores, keyword analysis, suggestions, and page numbering.
  * Export ATS report data as CSV, including metadata, category scores, keywords, and recommendations.
  * The export menu now closes automatically when clicking outside or pressing Escape.


Sachinchaurasiya360 (changes requested):
## Requesting changes: close, but the PDF palette is off-system and there is leftover dead code

The feature is well built. CSV injection is handled properly, which is easy to miss:

```js
const sanitizeCsvValue = (val: string) => {
  if (/^[=+\-@\t\r]/.test(val)) return `'${val}`;
  return val;
};
```

The BOM prefix for Excel, the object-URL cleanup, the page-break helper, and the footer pagination loop are all done right. Also worth noting: removing the unused `useMemo` import here fixes the one error currently breaking `Typecheck Client` on main, so `Typecheck Client` passes on this PR.

Three things to fix.

### 1. The PDF uses slate, not stone

The project palette is lime plus stone only. These RGB values are Tailwind's slate ramp:

```js
doc.setTextColor(100, 116, 139);   // slate-500
doc.setTextColor(148, 163, 184);   // slate-400
doc.setTextColor(15, 23, 42);      // slate-900
doc.setFillColor(248, 250, 252);   // slate-50
doc.setTextColor(51, 65, 85);      // slate-700
doc.setTextColor(30, 41, 59);      // slate-800
```

They are raw RGB rather than class names, but it is still new code introducing slate, and the exported report is a user-facing artifact that should match the brand. The stone equivalents:

- stone-500 `120, 113, 108`
- stone-400 `168, 162, 158`
- stone-900 `28, 25, 23`
- stone-50 `250, 250, 249`
- stone-700 `68, 64, 60`
- stone-800 `41, 37, 36`

The lime accents (`236, 252, 203` and `132, 204, 22`) are already correct.

### 2. Dead variable

```js
const row = Math.floor(idx / 3);
if (col === 0 && idx > 0) y += 14;
...
void row;
```

`row` is computed, never used, then suppressed with `void row`. Please delete both lines. If the intent was to derive `y` from `row` instead of the running `col === 0` increment, that would actually be clearer than the current approach.

### 3. The tier color ternaries are hard to read

```js
doc.setTextColor(isGood ? 21 : isWeak ? 220 : 161, isGood ? 128 : isWeak ? 38 : 161, isGood ? 61 : isWeak ? 38 : 22);
```

Three nested ternaries interleaved across the R, G and B arguments. A lookup keeps it reviewable:

```js
const TIER_RGB: Record<string, [number, number, number]> = {
  good: [21, 128, 61],
  weak: [220, 38, 38],
  mid:  [161, 161, 22],
};
const [r, g, b] = TIER_RGB[isGood ? "good" : isWeak ? "weak" : "mid"]!;
doc.setTextColor(r, g, b);
```

### Optional

The dropdown handles click-outside and sets `aria-haspopup` / `aria-expanded` / `role="menu"`, which is more than most such menus get. It does not close on Escape or support arrow-key navigation between the two items. Not blocking for a two-item menu, but Escape is three lines and worth adding.

PR #2874, opened 2026-07-26
fix: added type="button" to all buttons repo wide

Abfa41:
## Description

Added explicit `type="button"` declarations to non-submit buttons across the client to prevent unintended form submissions and page reloads when buttons are used inside forms.

The shared `Button` component now defaults to `type="button"` while still allowing callers to explicitly override the type when a button is intended to submit a form.

Also updated raw `<button>` usages across the client to explicitly declare `type="button"` where they are not intended to submit forms.

## Related Issue

Fixes #2841

## Type of Change

- [x] Bug Fix
- [ ] Feature
- [ ] Enhancement
- [ ] Documentation

## Testing

- Verified the shared `Button` component defaults to `type="button"`.
- Added explicit `type="button"` attributes to non-submit raw `<button>` elements across the client.
- Verified that existing button behavior and styling remain unchanged.
- Confirmed that buttons intended for form submission can still explicitly use `type="submit"` when required.
- Ran the relevant project checks successfully.

## Screenshots / Video

_No UI changes in this PR_

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented non-submit controls across navigation, forms, dialogs, filters, learning activities, and admin tools from triggering unintended form submissions.
  * Preserved existing actions such as approvals, resets, removals, retries, modal controls, and navigation while improving button behavior in form contexts.


Sachinchaurasiya360 (approved):
﻿Good sweep, and the risk with this kind of change is real, so I verified it: none of the 25 touched files contains a `<form>` element, a `type="submit"` button, or an `onSubmit` handler. The only `FormSection` hits in `ResumeBuilderPage` are a custom layout component, not a form. So no button was relying on implicit submit behaviour and nothing breaks.

I also ran eslint and `tsc -b` on this branch. It produces exactly the same 12 problems / 3 errors and the same single `useMemo` type error as origin/main, so this adds no new lint or type failures. The two remaining CI failures are the known main-branch breakage.

One nit for a follow-up, not worth another round trip: `button.tsx` line 370 changed `type = "button"` to `type = 'button'`, which is an unrelated quote-style change in the one file that needed nothing.

PR #2876, opened 2026-07-26
fix: refactor job match writes to use memory map diffing and bulk batching (#2819)

swarupio:
# Pull Request

## Description
This PR resolves a significant N+1 write bottleneck in the `JobMatchingService.generateMatches` method. 

Previously, the logic executed a database `upsert` in an individual transaction for every single matched job inside a loop, severely impacting database pool connection limits. The logic has been completely refactored to:
*   **Batch Reads:** Extract all necessary `jobId`s from the vector search results and execute a single `findMany` to retrieve job data.
*   **In-Memory Diffing:** Fetch existing match records in chunks and use a `Set` to compute the difference between candidates and existing records in memory.
*   **Bulk Writes:** Separate the resulting candidates into arrays for Creates and Updates.
*   **Optimized Execution:** Use `createMany` with `skipDuplicates` for massive bulk inserts, and chunked `$transaction` blocks for updates, heavily reducing load and connection pool exhaustion on the Postgres adapter.

## Related Issue
Fixes #2819

## Type of Change
- [ ] Bug Fix  
- [ ] Feature  
- [x] Enhancement  
- [ ] Documentation  

## Testing
*   Verified that the logic accurately separates creates from updates by matching `userId` and `jobIndexId`.
*   Validated chunking logic (`CHUNK_SIZE = 1000`) prevents hitting Postgres query parameter limits.
*   Ran local ESLint checks (`npm run lint`) to ensure no new type or syntax errors were introduced.

## Checklist

## Summary by CodeRabbit

* **Performance Improvements**
  * Improved job matching efficiency by batching candidate selection and persistence.
  * Reduced database workload for large match sets with chunked reads and writes.
  * More robust processing so failures in a batch are less likely to disrupt overall matching.

* **Bug Fixes**
  * Avoids unnecessary processing when there are no users with embeddings or no eligible jobs.
  * Ensures dismissed or below-threshold jobs are excluded from generated matches.
  * Corrected match counting during bulk create/update.
  * Vector-search issues are now logged for easier troubleshooting.

### DIFF
--- a/server/src/module/job-index/job-matching.service.ts
+++ b/server/src/module/job-index/job-matching.service.ts
@@ -105,11 +105,14 @@ export class JobMatchingService {
       },
     });
 
-    let matchCount = 0;
+    if (users.length === 0) return 0;
 
+    const allJobIds = new Set<number>();
+    const userMatches = new Map<number, Array<{ id: number; similarity: number }>>();
+
+    // 1. Gather all vector searches (must remain per-user due to the `<=>` distance operator)
     for (const pref of users) {
       try {
-        // Vector search: top 20 most similar new jobs
         const results = await prisma.$queryRawUnsafe<Array<{ id: number; similarity: number }>>(
           `SELECT ji.id, 1 - (ji.embedding <=> up.embedding) AS similarity
            FROM "jobIndex" ji, "userJobPreference" up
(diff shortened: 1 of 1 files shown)


Sachinchaurasiya360 (changes requested):
## Requesting changes: the write path trades one pool problem for a worse one

The read side of this refactor is a clear improvement. Collapsing the per-user `findMany` into one batched fetch with a `jobMap` is exactly right, and letting write errors propagate instead of being swallowed by the `catch` that was only meant for missing pgvector columns is a real bug fix. Three things about the write side need rework before this lands.

### 1. `CHUNK_SIZE = 1000` transactions are far larger than what they replaced

```js
for (let i = 0; i < toUpdate.length; i += CHUNK_SIZE) {
  const chunk = toUpdate.slice(i, i + CHUNK_SIZE);
  await prisma.$transaction(chunk.map((u) => prisma.jobMatch.update({ ... })));
}
```

Before this PR each transaction held at most 20 statements, because it was scoped to one user's vector-search results. Now a single transaction can hold 1000 sequential `UPDATE`s on one pooled connection. Against Prisma Postgres over the network that is a long-lived connection and a transaction-timeout risk, on a cron that already exceeds the Hobby 60s cap. Please drop the chunk size for the write path to something closer to the old scope (50 to 100).

### 2. A 1000-clause `OR` is a query planner problem

```js
const existing = await prisma.jobMatch.findMany({
  where: { OR: chunk },   // chunk is up to 1000 { userId, jobIndexId } pairs
  select: { userId: true, jobIndexId: true },
});
```

Postgres handles large `OR` lists of composite equality poorly. Since you already know both id sets, this is one cheap query instead:

```js
const existing = await prisma.jobMatch.findMany({
  where: {
    userId: { in: [...new Set(matchCandidates.map((c) => c.userId))] },
    jobIndexId: { in: [...new Set(matchCandidates.map((c) => c.jobIndexId))] },
  },
  select: { userId: true, jobIndexId: true },
});
```

That over-fetches some pairs, but the `existingMatches` set membership test already filters correctly, and it drops the chunk loop entirely.

### 3. Replacing `upsert` with read-then-`update` introduces a TOCTOU that aborts the whole run

`upsert` tolerated a row disappearing between decision and write. `prisma.jobMatch.update` throws `P2025` if the row is gone, and because it is inside `$transaction`, that rejects the whole chunk and now propagates (correctly, per point 1 above) all the way out of `generateMatches`. So one deleted row fails every remaining match in the run.

This is reachable: `internhack-ai.cron.ts` calls `prisma.jobMatch.deleteMany({ where: { dismissed: true, createdAt: { lt: ... } } })` in the same pipeline, and `job-feed.service.ts` lets users mutate these rows concurrently.

Simplest fix that keeps the batching win, using `updateMany` which is a no-op on zero matches rather than a throw:

```js
await prisma.$transaction(
  chunk.map((u) =>
    prisma.jobMatch.updateMany({
      where: { userId: u.userId, jobIndexId: u.jobIndexId },
      data: { score: u.score, skillMatch: u.skillMatch, locationMatch: u.locationMatch, salaryMatch: u.salaryMatch, vectorScore: u.vectorScore },
    })
  )
);
```

### Minor

`matchCount += toCreate.length` counts rows that `skipDuplicates` skipped, so the returned count can overstate what was written. Using the `count` from the `createMany` result would be accurate.

Sachinchaurasiya360 (approved):
Reviewed the latest commit. The earlier write-path concerns have been addressed with bounded reads, smaller update transactions, updateMany for deleted-row races, and database-backed counts. Local server \
pm run typecheck\ and \
pm run lint\ pass in a clean PR worktree.

PR #2878, opened 2026-07-26
fix: bound presigned-URL generation concurrency to prevent resource spikes (#2818)

swarupio:
# Pull Request

## Description
This PR resolves an unbounded concurrency vulnerability in `s3.utils.ts`. 

Previously, `signUrls` utilized an unbounded `Promise.all`, which risked spiking outbound network calls, hitting rate limits, and contending for event loop resources if passed a massive array of URLs. 

The function has been refactored to chunk the input array, safely processing the presigned URL generation in bounded batches (max 10 concurrent operations).

## Related Issue
Fixes #2818

## Type of Change
- [ ] Bug Fix  
- [ ] Feature  
- [x] Enhancement  
- [ ] Documentation  

## Testing
*   Verified that chunking logic correctly concatenates and resolves the full original array.
*   Ran `npm run lint` locally to ensure no syntax or typing violations.

## Checklist

### DIFF
--- a/server/src/utils/s3.utils.ts
+++ b/server/src/utils/s3.utils.ts
@@ -109,7 +109,18 @@ export async function signUrl(url: string, expiresIn = 3600): Promise<string> {
 
 /** Sign every URL in a string array (e.g. user.resumes). */
 export async function signUrls(urls: string[], expiresIn = 3600): Promise<string[]> {
-  return Promise.all(urls.map((u) => signUrl(u, expiresIn)));
+  // Cap concurrency to prevent spiking outbound calls and resource contention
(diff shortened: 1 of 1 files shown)


Sachinchaurasiya360 (approved):
﻿Good fix, and it matches the repo guidance against unbounded `Promise.all` on a pool-backed path. `signUrls` is called with `user.resumes` and similar arrays, so an unbounded fan-out was a real spike risk. Chunking at 10 is simple and dependency-free. There is some head-of-line blocking at each chunk boundary versus a rolling window, but at this size it does not matter.

PR #2880, opened 2026-07-27
fix(a11y): add accessible names to icon-only buttons

Abfa41:
## Description

Added accessible names to icon-only buttons across the client application.

This ensures buttons that rely solely on icons, such as copy, navigation, edit, close, toggle, and similar controls, have meaningful accessible names for screen reader users and other assistive technologies.

The changes include adding appropriate `aria-label` attributes to icon-only button usages across relevant client components and pages.

## Related Issue

Fixes #2845

## Type of Change

- [ ] Bug Fix
- [ ] Feature
- [x] Enhancement
- [ ] Documentation

## Testing

- Audited icon-only button usages in the modified client components.
- Added descriptive `aria-label` values based on each button's action.
- Verified that the affected controls now expose meaningful accessible names to assistive technologies.
- Ran the project's lint/validation checks.

## Screenshots / Video

_No UI changes in this PR_

## Checklist

- [ ] **Screenshot or video attached for every UI change** (PR will be rejected without this)

## Summary by CodeRabbit

* **Accessibility Improvements**
  * Added `aria-label` attributes to sharing, downloading, copying, and other action buttons to improve screen-reader clarity.
  * Enhanced accessibility for crop modal actions, code runner controls, and sign-out/logout buttons in navigation and sidebars.
  * Updated administrative action buttons (approve, edit, delete, save/cancel) with contextual labels derived from item names/titles.


Sachinchaurasiya360 (changes requested):
## Requesting changes: two of these patterns work against screen reader users

Most of this PR is good. The genuinely icon-only buttons (share icons in `CompletionCelebration`, copy buttons in `NodeEditor` / `ResumeGeneratorPage` / `RoadmapCertificatePage`, the icon action buttons in the admin pages) had no accessible name at all, and the contextual labels like `aria-label={`Edit ${topic.name}`}` are exactly right. Two things need fixing.

### 1. `aria-label` on buttons that already have visible text

`AdminAptitudePage.tsx:424`:

```jsx
<button
  aria-label="Cancel question edit"
  onClick={() => setEditingQ(null)}
  className="px-4 py-2 bg-gray-800 text-gray-300 rounded-lg hover:bg-gray-700 text-sm"
>
  Cancel
</button>
```

The button's visible text is "Cancel". Adding `aria-label` **replaces** the accessible name rather than supplementing it, so the announced name is now "Cancel question edit" while the visible label reads "Cancel". That breaks WCAG 2.5.3 Label in Name, and it breaks speech-input users who say "click Cancel" and no longer match the control.

Same pattern on "Save question", "Cancel category edit", "Save category", "Cancel topic edit", "Save topic". These are text buttons, not icon-only ones, so they should be left alone. If the extra context is worth having, `aria-describedby` adds it without overriding the name.

Please go through the additions and keep `aria-label` only where the button renders no text.

### 2. `aria-controls` points at an id that does not exist

`LessonCodeRunner.tsx:62`:

```jsx
aria-label="Toggle output"
aria-expanded={outputOpen}
aria-controls="output"
```

There is no `id="output"` anywhere in the file, so this is a dangling reference that assistive tech cannot resolve. Either add `id="output"` to the element the button expands, or drop `aria-controls` and keep `aria-expanded`, which is useful on its own.

`aria-expanded={outputOpen}` is a nice addition either way.

### Minor

`AdminSkillTestsPage.tsx` has an `aria-label` added to a status toggle that uses `rounded-full`. Not introduced by this PR, but that is a pill badge, which the project design rules exclude. Worth flagging for whoever touches that file next rather than fixing here.

PR #2881, opened 2026-07-27
perf: replace N+1 bookmark upserts with batched writes

Abfa41:
## Description

Replaced the per-bookmark `upsert` loop in `bulkMigrateBookmarks` with batched bookmark synchronization.

The previous implementation performed one database upsert per repository ID, which could result in N+1 writes and hold multiple database connections when migrating a large set of bookmarks.

This change:

* Fetches valid repository IDs in a single query.
* Fetches the user's existing bookmarks in a single query.
* Computes the bookmark delta in memory.
* Creates new bookmarks using `createMany({ skipDuplicates: true })`.
* Removes stale bookmarks using a single `deleteMany`.
* Wraps the batched writes in a single transaction.

This reduces the number of database operations and avoids the previous per-item upsert pattern.

## Related Issue

Fixes #2823

## Type of Change

* [ ] Bug Fix
* [ ] Feature
* [x] Enhancement
* [ ] Documentation

## Testing

* Verified the bookmark migration flow after replacing the per-item upsert logic with batched reads and writes.
* Confirmed that valid repository IDs are handled correctly.
* Confirmed that new bookmarks are created and stale bookmarks are removed through the computed delta.
* Confirmed that the final bookmarked repository IDs are returned correctly.

## Screenshots / Video

*No UI changes in this PR*

## Checklist

* [x] Code follows project guidelines
* [x] No new compile/type errors
* [x] Tested manually (include steps above)
* [x] No `.env`, credentials, or `node_modules` committed
* [x] Docs updated (if needed)
* [x] **Screenshot or video attached for every UI change** (PR will be rejected if missing)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved bookmark migration reliability by preventing duplicate entries.
  * Ensured bookmarks no longer included in the selected repository set are removed.
  * Preserved the latest bookmarked repository list after migration.

### DIFF
--- a/server/src/module/opensource/opensource.service.ts
+++ b/server/src/module/opensource/opensource.service.ts
@@ -1,6 +1,5 @@
 import { prisma } from "../../database/db.js";
-import { Prisma } from "@prisma/client";
-import type {opensourceRepo, RepoDomain, RepoDifficulty } from "@prisma/client";
+import type { opensourceRepo, RepoDomain, RepoDifficulty } from "@prisma/client";
 import { fetchGithubGoodFirstIssues, fetchGithubStats, fetchRepoHealthData } from "../../lib/github.js";
 import { sendEmail } from "../../utils/email.utils.js";
 import { cacheGet, cacheSet, cacheDel } from "../../utils/cache.js";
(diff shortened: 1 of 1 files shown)


Sachinchaurasiya360 (changes requested):
## Requesting changes: this turns an additive import into a destructive sync

The batching goal is right, and `createMany({ skipDuplicates: true })` is the correct replacement for N `upsert` calls. But the PR also adds a `deleteMany` that changes what this endpoint means.

`bulkMigrateBookmarks` is the handler for `POST /bookmarks/migrate`, documented in `opensource.routes.ts` as "bulk-import localStorage bookmarks to the server". It is an **additive merge**. This PR makes it a **full replace**:

```js
const idsToDelete = existingBookmarks
  .map((bookmark) => bookmark.repoId)
  .filter((repoId) => !targetIds.has(repoId));
```

Failure scenario: a user bookmarks 20 repos on their laptop. They open the site in a fresh browser where localStorage holds 2 stale bookmarks. The client posts those 2 ids, and this code deletes the other 18 server-side bookmarks. Silent data loss on a path whose only job was to import.

Related, the early return was dropped:

```js
// before
if (validIds.length === 0) return this.getBookmarkedRepoIds(userId);
```

Now, if every incoming id is stale and resolves to nothing valid, `validIds` is empty, `idsToDelete` becomes every existing bookmark, and the user's whole list is wiped. `bulkMigrateBookmarksSchema` enforces `.min(1)` on the request, so an empty array cannot arrive from the client, but ids pointing at deleted repos get you to the same place.

### What to change

1. Drop the `deleteMany` block and `idsToDelete` entirely. Keep the function additive.
2. Restore the `validIds.length === 0` early return.
3. With deletes gone, `Serializable` and the 3-attempt retry loop are no longer needed. `createMany({ skipDuplicates: true })` is already idempotent under concurrency, so a plain call is enough and avoids holding a serializable transaction on this path.

That leaves the actual win from this PR: one `createMany` instead of N `upsert` round-trips.

### Nit

`import type {opensourceRepo, RepoDomain, RepoDifficulty }` lost the space after the brace.

If a replace-style sync is genuinely wanted, that belongs in a separate endpoint with a name that says so, not in `migrate`.

PR #2882, opened 2026-07-27
fix: add NaN validation for parseInt in aptitude controller

Akshita-2307:
Closes #2861

This PR adds NaN validation after parseInt(req.params.id) in the aptitude controller's submitAnswer method. Previously, non-numeric :id parameters would pass NaN directly to Prisma, causing a 500 error.

Changes:
- server/src/module/aptitude/aptitude.controller.ts — Added if (isNaN(questionId)) check returning 400 Bad Request

GSSOC

## Summary by CodeRabbit

- **Bug Fixes**
  - Added validation for invalid question IDs when submitting answers.
  - Requests with malformed question IDs now return a clear 400 error instead of being processed.

### DIFF
--- a/server/src/module/aptitude/aptitude.controller.ts
+++ b/server/src/module/aptitude/aptitude.controller.ts
@@ -31,6 +31,7 @@ export class AptitudeController {
   async submitAnswer(req: Request, res: Response, next: NextFunction) {
     try {
       const questionId = parseInt(req.params.id as string);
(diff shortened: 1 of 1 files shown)


Sachinchaurasiya360 (approved):
﻿Correct guard, and it matches the style already in this handler (the `if (!answer)` check on the next line uses the same early-return shape). Without it a non-numeric `:id` gave `NaN`, which reaches Prisma and surfaces as a 500 rather than a 400.

PR #2883, opened 2026-07-27
fix: reset subscriptionPlan to FREE in holdSubscription

Akshita-2307:
Closes #2863

This PR fixes the holdSubscription method in payment.service.ts which was setting subscriptionStatus: 'EXPIRED' without resetting subscriptionPlan to 'FREE'. This caused stale plan data to persist, unlike the expireSubscription method which correctly resets both fields.

Changes:
- server/src/module/payment/payment.service.ts — Added subscriptionPlan: 'FREE' to the holdSubscription update call

GSSOC

## Summary by CodeRabbit

* **Bug Fixes**
  * Expired subscriptions now correctly switch to the free plan, preventing outdated subscription plan information from being retained.

### DIFF
--- a/server/src/module/payment/payment.service.ts
+++ b/server/src/module/payment/payment.service.ts
@@ -366,7 +366,10 @@ export class PaymentService {
 
     await prisma.user.update({
       where: { id: payment.userId },
-      data: { subscriptionStatus: "EXPIRED" },
+      data: {
+        subscriptionStatus: "EXPIRED",
+        subscriptionPlan: "FREE",
+      },
     });
     await invalidateUserTierCache(payment.userId);
   }


Sachinchaurasiya360 (approved):
﻿Correct and consistent with the rest of the lifecycle. The cancel/expire path a few lines above already writes both `subscriptionStatus: "EXPIRED"` and `subscriptionPlan: "FREE"`, so `holdSubscription` was the odd one out leaving a stale paid plan on a user who no longer has access. No recovery path depends on the retained value either: both `activateSubscription` and `renewSubscription` re-derive the plan from `sub.metadata["billing"]`, so a resumed subscription gets its plan back from the webhook. Cache invalidation is already handled below.

PR #2884, opened 2026-07-27
fix: add orderBy to DSA daily problem query for deterministic selection

Akshita-2307:
Closes #2864

This PR adds orderBy: { id: 'asc' } to the DSA daily problem query which previously had no ordering. Without an explicit ORDER BY, PostgreSQL does not guarantee row order, so the 'daily problem' could change between requests depending on the query planner.

Changes:
- server/src/module/dsa/dsa.service.ts — Added orderBy: { id: 'asc' } to prisma.dsaProblem.findMany() in getDailyProblem()

GSSOC

### DIFF
--- a/server/src/module/dsa/dsa.service.ts
+++ b/server/src/module/dsa/dsa.service.ts
@@ -1353,7 +1353,9 @@ Return ONLY a JSON array with exactly ${testCases.length} items in the same orde
   async getDailyProblem(userId?: number) {
     const today = new Date().toISOString().slice(0, 10);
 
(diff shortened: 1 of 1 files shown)


Sachinchaurasiya360 (approved):
﻿Correct fix. `getDailyProblem` indexes into the result set with `problems[seed % problems.length]`, and Postgres gives no ordering guarantee for a bare `findMany()`, so the "daily" problem could differ between requests on the same day. Adding `orderBy: { id: "asc" }` makes the seed deterministic, which is the whole point of the seed. Minimal and scoped.

